Issue #29, opened 2014-01-24
Can't get the http status code on callbacks


slorber:
When a callback is fired, the xhr object is not passed, neither the status code.

```
    this.failFetch = function(xhr, status) {
        console.error("failFetch called with",xhr,status);
        this.addStatus(xhr.req, status)
        kb.add(xhr.uri, ns.link('error'), status)
        this.requested[$rdf.uri.docpart(xhr.uri.uri)] = false
        this.fireCallbacks('fail', [xhr.requestedURI, status])
        xhr.abort()
        return xhr
    }
```

The status here is a String usually, like "XHR Error: "

When the fetch failed it would be nice to have a way to get the http status code in the callbacks when calling requestURI

timbl:
Fixed in commit ca77d9ed113f9f0b5f5f8dc0bf97a76d053aaa26

PR #32, opened 2014-02-05
Add XHR failed object when calling this.fireCallbacks('fail',[...])); to...

slorber:
... help with issue #29 of rdflib

I didn't add the http status to the args because the most open way to handle this is to provide directly the xhr object.
There may not even be any http status codes for certain errors (like same origin policy errors, timeout errors etc...)

The XHR object contains the URI, the http status code and more informations that can be useful to the callback user. This is not the most "user friendly" but at least it handles all the cases

### DIFF
--- a/web.js
+++ b/web.js
@@ -464,7 +464,7 @@ $rdf.Fetcher = function(store, timeout, async) {
         this.addStatus(xhr.req, status)
         kb.add(xhr.uri, ns.link('error'), status)
         this.requested[$rdf.uri.docpart(xhr.uri.uri)] = false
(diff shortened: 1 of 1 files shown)


jeswr:
Closing with thanks: the XHR-based fetcher this patched (web.js) was replaced by the fetch-based Fetcher (src/fetcher.ts) long ago. Its failure path now rejects with a `FetchError` that carries the HTTP status (src/fetcher.ts:98; failFetch at :1296), and the response metadata is recorded in the store — which covers what this PR set out to expose to fail callbacks.

Issue #36, opened 2014-03-18
Support possibility to fetch resources with custom header


slorber:
It would be nice to be able to specify additional headers to the fetcher (or to nowOrWhenFetched)

Our usecase with @bblfish is that some RWW WAC protected resources are public for all (wac mode READ) but the owner has additional permissions on it (like READ WRITE).
As the resource is public for all, the server doesn't need to auth the client using a X509 WebID certificate and can give the content to the client. The matter is that we introduced a response header `Allow` that lists all http verbs that can be used by the client to use this resource.
So if anyone can READ, anyone can get the resource without auth/certificate, and will get `Allow: GET`. The problem is for the resource owner. The resource owner can get its own resource in 2 different modes: authenticated and not authenticated. As the resource is public, the server doesn't ask for a certificate by default. The resource owner's client needs to send a hint to the server to ask for an authentication even if the resource is public. This way the server can return the appropriate header for the owner's client: `Allow: GET,POST,PUT,PATCH,DELETE`

This should probably be discussed with @deiu and @melvincarvalho but anyway I think it would be great to support some flexible way to customize the request in rdflib.js

timbl:
This feature has been provider in the options to load() and nowOrWhenFetched for  a long time now. Closing as done.

PR #37, opened 2014-03-18
#36 try to be able to customize jquery ajax request in requestURI method

slorber:
Pull request for https://github.com/linkeddata/rdflib.js/issues/36


timbl:
I'm afraid we are removing/have removed the use of JQuery in the library. So this no longer applies

PR #44, opened 2014-06-23
add jsonld capabilities

chrysn:
this only works if the jsonld.js library[1] is included. no dedicated
parser is implemented, the work is delegated to that library, and a file
type handler takes care of converting jsonld.js' node format into rdflib
terms.

current caveats are:
- jsonld.js brings has to launch xmlhttprequests by itself
- there is content type handling code which i failed to trigger (see
  FIXME comment)
- i didn't do test cases that involve the `why` part of statements, and
  my literals tests were few too (barely enough to see datatypes are
  terms but languages aren't)
- accept priorities were set arbitrarily
- no tests are implemented
- the dependency on jsonld.js is not declared, and no tests were done
  without the library present.
- when using the widespread crossSiteProxyTemplate
  `http://data.fm/proxy?uri={uri}`, be aware that json-ld does not seem
  to work with it.

[1] https://github.com/digitalbazaar/jsonld.js

### DIFF
--- a/web.js
+++ b/web.js
@@ -357,6 +357,13 @@ $rdf.Fetcher = function(store, timeout, async) {
                     return
                 }
 
+                if (rt.slice(0, 500).match(/["']@(context|vocab)["']:/)) {
+                    sf.addStatus(xhr.req, "May be JSON-LD. We'll assume it's that "
+                            + "but its content-type wasn't JSON or JSON-LD.\n")
+                    sf.switchHandler('JSONLDHandler', xhr, cb)
+                    return
+                }
+
                 // We give up finding semantics - this is not an error, just no data
                 sf.addStatus(xhr.req, "Plain text document, no known RDF semantics.");
                 sf.doneFetch(xhr, [xhr.resource.uri]);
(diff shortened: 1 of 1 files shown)


chrysn:
i think that a jsonld module needs to to xmlhttprequests by design: the @context needs to be resolved before triples can be generated. things would be different if the context were an rdf document by itself, the yes, i'd expect the jsonld module to be integrated deeper and to use rdflib facilities. if some more control is needed (eg. to make the xhrs mockable), a documentLoader can be passed into .toRDF and will then be used to fulfill the requests.

as for the second comment (on quad store, first/rest generation), i don't understand what you are suggesting.

for the time being, i've rebased the otherwise unmodified pull request onto current master. this might solve issue #59. one further change was necessary:

i've noticed that the jsonld library is already used in rdflib these days, but in completely different parts (only available in node.js, and only for conversion to jsonld). to have those modes co-exist peacefully, i'd have to dig deeper into how to get a module available both from inside node.js and in the browser. (the `var jsonld = require('jsonld');` made the jsonld module as loaded in the browser unavailable, i renamed it to avoid the clash).

akuckartz:
Any news regarding this PR ?

josephguillaume:
Looks like this PR is outdated and can be closed given rdflib now has jsonld capabilities?
https://github.com/linkeddata/rdflib.js/issues/59 has been closed

https://linkeddata.github.io/rdflib.js/doc/globals.html#jsonldparser https://github.com/linkeddata/rdflib.js/blob/master/src/jsonldparser.js

jeswr:
Closing with thanks: JSON-LD support landed via a dedicated parser (src/jsonldparser.js on the `jsonld` library, wired into parse() and the Fetcher), and the companion issue #59 is closed. The single-file web.js this patched no longer exists, so there is nothing left to merge — appreciated as the first push toward JSON-LD in rdflib.

PR #62, opened 2015-03-12
fixed logic for using proxy

bennpearson:
We've been using the proxy functionality and found that it was still throwing cors errors in the console but recovering. We tracked it down to a logic error.


ajtucker:
Any chance we can get this merged?  It's only a tinsy-tiny thing!

deiu:
Thanks!

PR #86, opened 2015-11-16
Parse XMLLiteral object value is XML source, not DOM

jamsden:
When parsing an XMLLiteral object, the value of the object should be
normalized XML source, not the parsed XML DOM. This solution serializes
the dom childNodes back into source, but doesn’t yet do the
normalization. Normalization is required to allow equality to work on
XMLLiteral objects.

### DIFF
--- a/dist/rdflib-node.js
+++ b/dist/rdflib-node.js
@@ -1978,6 +1978,8 @@ $rdf.RDFParser = function (store) {
     return true
   }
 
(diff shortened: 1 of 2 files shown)


deiu:
+r for me. Waiting for a second opinion.

timbl:
I agree that the value of the RDF node should be the serialization. That is how it is defined.

Issue #94, opened 2016-02-03
Do we still need `rdflib-alpha.js`?


dmitrizagidulin:
Is it safe to remove the `dist/rdflib-alpha.js` Make step? It's identical to the main `rdflib.js` file:

```Make
dist/rdflib-alpha.js: $R module.js
	echo "(function(root, undef) {" > $@
	cat $R module.js >> $@
	date '+$$rdf.buildTime = "%Y-%m-%dT%H:%M:%S";'  >> $@
	echo "})(this);" >> $@

dist/rdflib.js: $R module.js
	echo "(function(root, undef) {" > $@
	cat $R module.js >> $@
	date '+$$rdf.buildTime = "%Y-%m-%dT%H:%M:%S";'  >> $@
	echo "})(this);" >> $@
```
Some net searching seems to indicate that it was possibly intended to include RDFa parsing originally? (Source: ["Maybe we should make a rdflib-alpha.js which does."](https://github.com/linkeddata/rdflib.js/issues/52)). In any case, it doesn't seem to be needed at the moment.

deiu:
:+1: to remove that step.

dmitrizagidulin:
Talked to @timbl yesterday, he gave the go-ahead to unify the three dist/ releases into one (and merge the RDFa parser into the main one, even though it's currently buggy).

dmitrizagidulin:
Fixed in https://github.com/linkeddata/rdflib.js/issues/95, closing.

Issue #93, opened 2016-02-03
Fix browserify usage during build step


dmitrizagidulin:
The current Makefile / build procedure has a couple of problems that make `rdflib.js` impossible to use in other Node apps that want to use [browserify](http://browserify.org/):

1. The Make process uses Browserify on only one file (`web.js`), and then concatenates a bunch of other javascript files together with the browserified `web.js` into one big file in `dist/rdflib.js` (and wraps the component modules in its own wrapper, `(function(root, undef) { ...modules... })(this);`.

  This is basically re-implementing Browserify, except a browserified module is embedded in that.

2. The NPM packaging (the `main:` directive in `package.json`) then points to the concatenated/browserified bundle `dist/rdflib.js` instead of using Node's `include()` syntax. 

The end result of these two things, is that if another Node.js app requires `rdflib`, it becomes impossible to use Browserify on *that* app. (I've ran into this with `solid.js`). (Since browserifying an already-browserified library crashes and burns.)

The solution to all of this is to slightly modify the `rdflib.js` structure and build steps:

1. Create a main `index.js` Node module that imports all the other dependent modules that the Make build step concatenates together. And then point `package.json` / `main:` directive to `index.js`.

2. Create a similar function for the `rdflib-rdfa.js` distribution, that imports `parseRDFa.js` in addition to the other modules. (Better yet, just include it in the main one).

3. Use `browserify index.js > dist/rdflib.js` to actually concatenate all the component modules and browserify *those*. (Which is what it was created to do).

dmitrizagidulin:
Should be fixed with PR #95. Awaiting further testing results from @timbl

dmitrizagidulin:
- Ran into issue where browserify was not exposing `$rdf` global, fixed in recent commit.
- Currently experiencing an error related to `var xhr = new global.XMLHttpRequest()`, which is most likely being caused by `util.js:121`, `var XMLHttpRequest = require("xmlhttprequest").XMLHttpRequest;`. The `xmlhttprequest` package is not browserify-compatible. Need to replace it with `xhr2`

dmitrizagidulin:
Seems to be working now, tested by myself and @timbl. Closing issue for the moment.

PR #95, opened 2016-02-04
Fix node/browserify workflow

dmitrizagidulin:
See issue #93 for background.
This is a quick fix (without a full re-architecting of rdflib.js to use node
modules and `require()`).

1. Instead of browserifying `web.js` and then concatenating, it first
  concatenates all files into `dist/rdflib-node.js`.
  This exports `$rdf`, and is suitable for use with other node modules.

2. The build step then also runs Browserify on `dist/rdflib-node.js` and turns
  it into `dist/rdflib.js`, which is what can be imported into browser
  scripts.

3. The npm `main:` is now pointed to `dist/rdflib-node.js`.

Removes the other two alternate dist packages, `dist/rdflib-alpha`
and `dist/rdflib-rdfa` (see issue #94).

Removes Coffeescript from the build steps. (Now using .js files only)
Previous .coffee files are still kept around, for reference.


dmitrizagidulin:
(Ignore the previous failed build, forgot to remove `*.js` files from `.gitignore`).

deiu:
:+1:

PR #103, opened 2016-02-26
Move .coffee files to reference/

dmitrizagidulin:
To avoid confusion (since the `.coffee` files are no longer involved in the Make process), move all the `*.coffee` files into the `reference/` directory.


timbl:
makes sense to me

PR #112, opened 2016-03-07
Fix CORS credential flag logic to avoid 'previously failed' issue

timbl:
Now when CORS-tiggered redirects are done to proxy servers, tracks the status of the redirected request.  

Avoided calling a non-cred-flag fetch in parallel with another one.


dmitrizagidulin:
+1, looks reasonable.

Issue #111, opened 2016-03-06
TypeError: $rdf.fetcher.lookUpThing is not a function


ckristo:
Hi,

I updated my rdflib.js example from version 0.2.6 to 0.5 and encounter the following error:

````
$rdf.fetcher.lookUpThing(st.subject, 'sparql:' + st.subject)
                     ^

TypeError: $rdf.fetcher.lookUpThing is not a function
    at Object.$rdf.SPARQLToQuery (/Users/christoph/Documents/Ausbildung/TU/033532 Medieninformatik/Bakk-Arbeit/Arbeit/examples/js/rdflib.js-example/node_modules/rdflib/dist/rdflib-node.js:6751:22)
````

It seems that `$rdf.fetcher` is a function, and not an object, see https://ckristo.net/owncloud/index.php/s/mcV0FZCP838dfJz

As workaround, setting `$rdf.fetcher` to null causes the faulty function call to be skipped.

BR, Chris

**Comment by ckristo** [comment]
cf. https://github.com/linkeddata/rdflib.js/issues/70

**Comment by ckristo** [comment]
Shouldn't `testMode` if set to `false` disable resource fetching? Why does `$rdf.SPARQLToQuery` then even try?

**Comment by timbl** [comment]
Yes, $rdf.fetcher is a function which returns you a new fetcher but you must  pass it a graph.   The two are linked together.   
```
   var g = $rdf.graph()
   var f = $rdf.fetcher(g)

g.add() ...
f.load() ...

```
You can think of $rdf.fetcher(g)  as a shorthand for new $rdf.Fetcher(g)   -- does this make sense?

DO NOT do:
```
 $rdf.fetcher = null; /*   BAD */ 
```
that breaks the system!

ckristo:
Hi,

I do not use `$rdf.fetcher` at all, the error is produced by rdflib.js's `SPARQLToQuery` method (see the example included above).

There is a check which tests if `$rdf.fetcher` is set in the code, and if you set it to null, the erroneous function call `.lookUpThing()` gets skipped. Sure, it is not a good idea to set it to null, but in my case this is the only way to get my example to work.

This is a bug and should stay open imo.

BR, Chris

ckristo:
`$rdf.fetcher` is used as some kind of factory method, right? Why does `SPARQLToQuery` use it as object? I suppose it expects `$rdf.fetcher` be a fetcher instance.

[...]

PR #113, opened 2016-03-08
A first suggestions for fixing bug #111.

ckristo:
First shot for fixing issue / bug #111.
- removes the formula (`kb` argument) dependency for `SPARQLToUpdate`
- `SPARQLToUpdate` takes an optional fetcher instances as argument
- Fixed typo in web.js (`cnsole` instead of `console`)

This PR was tested with the following example on a system with node v4.2.2: https://ckristo.net/owncloud/index.php/s/Mjtp2tdjIn4wqU3

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,9 @@ size:
 dist/rdflib.js: dist/rdflib-node.js
 	browserify -r ./dist/rdflib-node.js:rdflib --exclude "xmlhttprequest" --standalone "\$$rdf" > dist/rdflib.js
 
+dist/rdflib.min.js: dist/rdflib-node.js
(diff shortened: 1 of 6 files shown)


ckristo:
Alternatively, I could keep the formula as argument and only check if the formula's fetcher (`kb.fetcher`) is set .. In this case, I could use the formula's create* functions for creating RDF primitives again.

Please let me know which variant you prefer @timbl -- thanks!

BR, Chris

jeswr:
Closing with thanks: every file this patched (web.js, sparql.js, the checked-in dist bundles) was removed in the modular/TypeScript restructuring, so the diff has nothing to apply to — and the bug it fixed (#111) is itself closed.

PR #122, opened 2016-03-31
Timbl web standard -- fix problems with CORS mitigation and proxy calls

timbl:
And various other smaller fixes.


dmitrizagidulin:
:+1:, looks good.

Issue #124, opened 2016-04-06
Check for Tabulator presence before using it


dmitrizagidulin:
Standalone usage of rdflib (via solid.js, for example) is throwing an error due to missing tabulator dependency.

To replicate, load `rdflib.js` and `solid.js`, and call `solid.web.getParsedGraph(url)` (which calls rdflib.js/`fetcher.nowOrWhenFetched()`. Results in the following error (first reported by @rhiaro here: https://github.com/solid/solid.js/issues/81):

```
ReferenceError: tabulator is not defined
```

The following files (where `tabulator` is being used) need to be fixed:

1. [`web.js#Line 486`](https://github.com/linkeddata/rdflib.js/blob/911bcb467431c1d6ae40904999e0bf0547df1701/web.js#L486)
2. [`update.js#Line 45`](https://github.com/linkeddata/rdflib.js/blob/42673062a5ab7bec4a51dff5b1918bb3e69aa27e/update.js#L45) and [Line 41](https://github.com/linkeddata/rdflib.js/blob/42673062a5ab7bec4a51dff5b1918bb3e69aa27e/update.js#L41)

PR #125, opened 2016-04-09
fix refs to tabulator and rename .sparqlUpdate to .UpdateManager

timbl:
to fix https://github.com/linkeddata/rdflib.js/issues/124


dmitrizagidulin:
:+1:, merging

PR #127, opened 2016-05-09
RDFa parsing using a wrapper for RDFaProcessor

csarven:
Tested with green-turtle's tests (tests are not part of this PR) and sample RDFa documents


timbl:
Suggest bringing green-turtle in with require() and adding it to the dependencies.

timbl:
Otherwise looks good reading it t me.  excited to have this working

PR #128, opened 2016-05-11
Option to turn off predicate mapping in statementsToN3

csarven:
Currently `__Serializer.prototype.statementsToN3` is used to serialize N3, however the caller functions like `$rdf.serialize` is utilizing it for contentTypes other than N3 eg., `text/turtle`. One issue with that for non-N3 contentTypes is that, statementsToN3 maps the predicates to their syntactic sugar equivalents eg.,
```
var predMap = {
  'http://www.w3.org/2002/07/owl#sameAs': '=',
  'http://www.w3.org/2000/10/swap/log#implies': '=>',
  'http://www.w3.org/1999/02/22-rdf-syntax-ns#type': 'a'
}
```
When that happens, it creates invalid Turtle serialization.

This PR introduces an optional parameter for statementsToN3 to omit the predicate mapping. The default behaviour (without options) is still the same, however wherever Turtle uses statementsToN3, it passes `{noPredMap: true}`.

@timbl and I already discussed this.

### DIFF
--- a/dist/rdflib-node.js
+++ b/dist/rdflib-node.js
@@ -8248,9 +8248,10 @@ __Serializer.prototype._notNameChars =
                     ( __Serializer.prototype._notQNameChars + ":" ) ;
 
 
-__Serializer.prototype.statementsToN3 = function(sts) {
+__Serializer.prototype.statementsToN3 = function(sts, options) {
     var indent = 4;
     var width = 80;
+    var options = options || {};
 
     var predMap = {
         'http://www.w3.org/2002/07/owl#sameAs': '=',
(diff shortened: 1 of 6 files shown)


timbl:
@csarven Sorry I didn't see you working on this one. -- though you wanted me to fix it.  I did a simplar thing, except using the existing  letter flags structure .   Now  flags 't' 'i' and 's' suppress generation of shorthand for rdf:type, log:implies, owl:sameAs, and respectively. Propose to close this ... sorry about the duplication

csarven:
@timbl  No worries. I like your approach more. We can use that (or something alike) towards the N-Triples serialisation.

PR #132, opened 2016-07-01
Fix scope and promise issues with webCopy

timbl:
This only changes webCopy, fixing bugs with scope of variables and sequence of resolve()

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "rdflib",
   "description": "an RDF library for node.js. Suitable for client and server side.",
-  "version": "0.7.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "rdflib",
   "description": "an RDF library for node.js. Suitable for client and server side.",
-  "version": "0.7.1",
+  "version": "0.7.2",
```
deiu: You shouldn't manually bump the version. Publishing on npm takes care of this.

dmitrizagidulin:
Version bump reverted. Ready to merge (as per @timbl )

Issue #75, opened 2015-08-11
RDF/XML Literals not being parsed properly


jamsden:
Given some RDF/XML that contains:

<pre>
&lt;oslc:serviceProvider&gt;
    &lt;oslc:ServiceProvider rdf:about=&quot;https://oslclnx2.rtp.raleigh.ibm.com:9443/ccm/oslc/contexts/_pMhMgPsWEeSnQvDHoYok5w/workitems/services.xml&quot;&gt;
      &lt;dcterms:title rdf:parseType=&quot;Literal&quot;&gt;JKE Banking (Change Management)&lt;/dcterms:title&gt;
      &lt;oslc:details rdf:resource=&quot;https://oslclnx2.rtp.raleigh.ibm.com:9443/ccm/process/project-areas/_pMhMgPsWEeSnQvDHoYok5w&quot;/&gt;
      &lt;jfs_proc:supportLinkDiscoveryViaLinkIndexProvider rdf:parseType=&quot;Literal&quot;&gt;false&lt;/jfs_proc:supportLinkDiscoveryViaLinkIndexProvider&gt;
      &lt;jfs_proc:supportContributionsToLinkIndexProvider rdf:parseType=&quot;Literal&quot;&gt;true&lt;/jfs_proc:supportContributionsToLinkIndexProvider&gt;
      &lt;jfs_proc:globalConfigurationAware rdf:parseType=&quot;Literal&quot;&gt;compatible&lt;/jfs_proc:globalConfigurationAware&gt;
      &lt;jfs_proc:consumerRegistry rdf:resource=&quot;https://oslclnx2.rtp.raleigh.ibm.com:9443/ccm/process/project-areas/_pMhMgPsWEeSnQvDHoYok5w/links&quot;/&gt;
    &lt;/oslc:ServiceProvider&gt;
  &lt;/oslc:serviceProvider&gt;
</pre>

An a query such as:
someKb.the(aServiceProvider, DCTERMS('title’));

returns:

<pre>
&lt;dcterms:title rdf:parseType=&quot;Literal&quot;&gt;JKE Banking (Change Management)&lt;/dcterms:title&gt;    
</pre>

instead of the text. Am I missing something of is the dcterms:title being parsed incorrectly?

jamsden:
In the this.parseDOM() function, changing:
<pre>
                        var nv = parsetype.nodeValue;
                        if (nv === "Literal"){
[...]

PR #133, opened 2016-07-18
Updated addLiteral to properly support XMLLiteral

jamsden:
Parsing an XMLLiteral should store a normalized (sorted attributes) XML
string as the value for the literal, not the parsed XML dom.

Addresses issue #75.

### DIFF
--- a/dist/rdflib-node.js
+++ b/dist/rdflib-node.js
@@ -1971,7 +1971,18 @@ $rdf.RDFParser = function (store) {
         this.addSymbol(this.ARC, uri)
       },         /** Add a literal to this frame */'addLiteral': function (value) {
         if (this.parent.datatype) {
-          this.node = this.store.literal(value, '', this.store.sym(this.parent.datatype))
+          if (this.parent.datatype === RDFParser.ns.RDF + 'XMLLiteral' ) {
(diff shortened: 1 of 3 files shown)


dmitrizagidulin:
@jamsden @timbl - is this PR ready to merge?

jamsden:
It doesn’t do the attribution normalization (sorting the attributes) so === works on XMLLiterals that are the same except for ordering, but it does at least capture the correct contents.

Not sure === on XMLLiterals is a high priority, so this can possibly be deferred.

timbl:
Sounds like a definite improvement - suggest merge

timbl:
(I had earlier left the DOM pointer as th value, as that is what is useful to software. - if we have the DOM then maybe it would be be usefulk to leave a pointer to it, so that the caller doesn't have to waste time parsing it again?

jamsden:
Good point, but it doesn't follow the RDF/XLM spec and creates inconsistencies for consumers. For example, it breaks OSLC resources.

timbl:
This now needs to be patched in
 rdfxmlparser.js
at line: 
if (this.parent.datatype) {
Line 156 in rdfxmlparser.js

PR #135, opened 2016-07-19
Convert rdflib.js to Node-style require()/modules

dmitrizagidulin:
This is a conversion of all rdflib.js code to use Node `require()` module syntax (instead of concatenating all the files manually). In addition, it converts most modules to ES6 class syntax.
Uses `babelify` along with browserify for the browser builds.


dmitrizagidulin:
All tests pass (as well as solid.js client lib tests). Ready for review (cc @timbl).

Issue #139, opened 2016-08-05
Fix cyclic dependency in term.js and collection.js


dmitrizagidulin:
To reproduce current error:

```bash
node
> require('./index').graph().any([])
TypeError: Collection is not a constructor
    at term (/Users/dmitri/code/_Solid/rdflib.js/src/term.js:24:13)
    at IndexedFormula.statementsMatching (/Users/dmitri/code/_Solid/rdflib.js/src/indexed-formula.js:634:31)
    at IndexedFormula.anyStatementMatching (/Users/dmitri/code/_Solid/rdflib.js/src/indexed-formula.js:269:18)
    at IndexedFormula.any (/Users/dmitri/code/_Solid/rdflib.js/src/formula.js:29:19)
    at repl:1:13
    at REPLServer.defaultEval (repl.js:272:27)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.<anonymous> (repl.js:441:10)
    at emitOne (events.js:101:20)
```

dmitrizagidulin:
Fixed. Closing.

PR #140, opened 2016-08-05
Fix cyclic dependency with term.js and collection.js

dmitrizagidulin:
Fixes issue #139.
Also: Moves XSD data type declarations to xsd.js


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "rdflib",
   "description": "an RDF library for node.js. Suitable for client and server side.",
-  "version": "0.8.0",
+  "version": "0.8.1",
```
dmitrizagidulin: Ignore this version bump - it was from a previous npm publish, forgot to push to master.

timbl:
Looks good to me

Issue #137, opened 2016-08-03
Implement RDFJS Representation Task Force spec


dmitrizagidulin:
Update rdflib.js to match the [RDFJS Representation Task Force spec](https://github.com/rdfjs/representation-task-force). Road map:

##### Core (Terms, Quads, Graphs)
* [x] Implement basic [DataFactory API](https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md#datafactory) (`rdf.namedNode()`, `.literal()`, `.blankNode()`, `.triple()`, `.quad()` etc)
* [x] Implement the required attributes (`.termType` and `.value`)
* [x] Implement the required methods (`.equals()` and `.toCanonical()`)
* [ ] Implement [`DefaultGraph`](https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md#defaultgraph-extends-term)
* [ ] Refactor `termType` property to match the RDFJS spec.
* [ ] Pass the [RDFJS Data Model test suite](https://github.com/bergos/rdf-data-model-test-suite)

##### Stream interfaces
* [ ] Source (for parsers)
* [ ] Sink (for serializers)
* [ ] Store

timbl:
The Turtle/N3  serializer cannot have a stream interface, as it takes the statements and sorts them before outputting them, to get a compact and pretty version.

RubenVerborgh:
The _interface_ can still be streaming: the writer is free to buffer it all and only output at the end.

In other words: a non-streaming writer can be realized with a streaming interface; but a streaming writer can only be realized with a streaming interface. Hence, RDF/JS chooses the most generic option by defining a stream interface.

elf-pavlik:
I think in many use cases the 'pretty' requirement may not matter. Of course for people doing development, tools need to exist to pretty print some output, for CLI use one can always use pipes for that and modern (extensible) code editors often provide 'pretty format' plugins.

With progress on RDF/JS soon developers will have possibility to pick parsers, serializers and stores that they find fitting their requirements best. Some applications will use serializers that offer pretty output while others may choose better performing serializer that provides valid but not prettified output. I hope soon participants of RDF/JS will start documenting differences between various implementations of modules with similar functionality, this should assist people to choose which combination of modules may meet their requirements the best. Graphy.js has IMO nice example of comparing it's own parser to the one in N3.js https://github.com/blake-regalia/graphy.js#performance

elf-pavlik:
[...]

PR #141, opened 2016-08-09
Implement kb.match() alias and tests

dmitrizagidulin:
- Added `.addAll()` convenience method to IndexedFormula
- Added `.match()` method to IndexedFormula, to match the RDFJS TF specs.
  This is currently an alias for `statementsMatching()`.
- Added a `tests/unit/` folder for Tape unit tests

Partly addresses issue #137


dmitrizagidulin:
Discussed with @timbl on gitter. Merging.

PR #143, opened 2016-08-24
Publish a transpiled 'lib' directory rather than the current ES6 src

dan-f:
@dmitrizagidulin


dan-f:
@dmitrizagidulin mind giving this a quick review? the fact that we're currently distributing the es6 source is blocking some current work

Makefile
```diff
@@ -54,7 +58,6 @@ minify: dist/rdflib.min.js
 writable:
 	@sed -i -re 's/git:\/\/github.com\//git@github.com:/' .git/config
 
-.PHONY: test
```
dmitrizagidulin: I think we should leave the `.PHONY` statements in?

dmitrizagidulin:
Other than the makefile comment, looks good to me!

Makefile
dan-f: I did - I just collected them all at the top, because I find it a bit more readable.

dmitrizagidulin: I don't mind either way, but I thought the general Makefile convention was to place them right before their respective statements? (vs at the top)

dan-f: I haven't written Makefiles in a while, so I dunno. I like your suggestion though. I'll add a `.PHONY` on the line prior to each phony target.

dmitrizagidulin:
Looks good. (Thanks for doing this btw! It was making my life difficult as well.)

dan-f:
np! I'll merge and publish a new version once tests pass. :)

PR #154, opened 2016-11-21
Set q=0.9 for application/rdf+xml in the Accept header (fixes #153)

### LINKED ISSUE #153: Reduce preference for application/rdf+xml in favor of text/turtle.
Currently, when rdflib.js builds the Accept header values, it doesn't set a `q` value for `application/rdf+xml`, which means that by default is interpreted as `q=1`.

I would like to suggest a minor deprecation, by setting `q=0.9` for `application/rdf+xml`, in favor of using `text/turtle` with `q=1`.

I will submit a PR with a fix very soon.

deiu:
Fixes #154 by setting the Accept header `q` value for `application/rdf+xml` to `q=0.9`.

Edit: I wanted to write a test for this but as it turns out we don't have tests for the Fetcher.

### DIFF
--- a/dist/rdflib.js
+++ b/dist/rdflib.js
@@ -1125,9 +1125,237 @@
 }());
 
 }).call(this,require('_process'))
-},{"_process":10}],2:[function(require,module,exports){
+},{"_process":24}],2:[function(require,module,exports){
+'use strict'
+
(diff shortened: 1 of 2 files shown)


dmitrizagidulin:
looks good.

Issue #151, opened 2016-10-28
Tests


dan-f:
I hope to start a conversation in this issue. Though I am a newcomer to the
inner-workings of `rdflib.js`, and it's possible that I've missed some details, I think we can do a lot better with our test suite.

According to a code dive, we've got a few different tests:
- We run nodeunit on `tests/*.js` (currently just `tests/term.js` and
  `tests/uri.js`)
- We run the serialization tests located at `tests/serialize`
- We run tape on `tests/unit/*.js` (currently just `match-test.js`)

Each of these three test "suites" uses a different test tool (the serialization
tests using a home-grown approach).

Furthermore, we've got this `test` directory ("test" without the "s"), which
doesn't appear to get called at all during continuous integration on Travis, or
by the standard test commands (e.g. `npm test` and `make test`). When I open the
`index.html` (served over http) and follow the instructions, I see in the
console that `/dist/rdflib-rdfa.js` failed to load (404), and then see six "$rdf
is not defined" ReferenceErrors. Is this test "suite" deprecated?

There are also some test files within `tests/` that don't seem to be touched by `npm test`.

Would anyone with more `rdflib.js` history be willing to give some context on what tests are currently expected to work, and which are deprecated? I'd like to work with folks to convert what we have into a more cohesive test suite with a clear developer interface.

dan-f:
I'll add that it's my hope to move towards a traditional test "pyramid" architecture in which unit tests are the most common followed by integration and other higher level (functional etc) tests.

dmitrizagidulin:
Looks like this issue has been addressed by the mocha test suite, and is safe to close?

dan-f:
I'd like to leave this open. There are still a ton of broken tests that are not run on CI that need to be converted.

jeswr:
As part of the maintainer-signed-off triage (plan #824): the 2016 situation this issue described (nodeunit + tape + a homegrown runner, plus dead test directories) no longer exists — `npm test` now drives a single mocha unit suite plus serialize and type checks. The remaining test-suite work is tracked in the open issues #400 and #713 and the modernization plan. Closing this meta-issue as superseded; specific remaining gaps are best filed against those trackers.

PR #156, opened 2016-12-07
Add a coherent unit test framework

dan-f:
This PR is part of the effort I outlined in #151. This is just meant to create a consistent framework for adding unit tests. Further work will be required for integration/higher level tests.

Given that BDD-style test frameworks have become the de-facto standard, I added `mocha` and `chai`, which are both more actively maintained than `tape` and `nodeunit`. I created some example tests, and converted the old tape tests to the new BDD style as well.

I couldn't figure out the `term.js` and `uri.js` unit tests that use `nodeunit`, so I haven't tried to convert them yet, but rather stuck them in their own directory. If someone is more familiar with them, perhaps they could help convert them to `mocha`.

@timbl @dmitrizagidulin thoughts?


dan-f:
I forgot to mention - this was also spurred by wanting to make a code change to how Literals choose their numeric datatypes, so there's a change there. The idea is that in JS, all numbers are 64 bit IEEE doubles. Integer/double/decmial types are chosen based on the precision of the provided value.

dmitrizagidulin:
Looks great to me. Thanks for doing this!

dmitrizagidulin:
Wait a second, what happened to this code? :) I'm not seeing this merged PR in the `master` branch at all!

dan-f:
:( did someone `git push -f origin master`?

PR #158, opened 2016-12-22
Fix the Query import bug in indexed-formula.  Add Formula.connectedSt…

timbl:
…atements


src/indexed-formula.js
```diff
@@ -62,7 +62,7 @@ function handleRDFType (formula, subj, pred, obj, why) {
 }
 
 class IndexedFormula extends Formula {
-  constructor (features) {
```
dmitrizagidulin: Why remove the spaces? (standard formatting requires the space in the function declaration)

PR #159, opened 2016-12-22
Removed dependency to global $rdf

megemege77:
When including rdflib.js into a web application using webpack, the
resulting application doesn't run. The reason for this is that there are
references to the global $rdf object in modules which are initialized
before the global $rdf object has been properly initialized.

Removed all references to the global $rdf object and
replaced them by direct references to the underlying functionality.

### DIFF
--- a/src/fetcher.js
+++ b/src/fetcher.js
@@ -1,3 +1,4 @@
+/* global $SolidTestEnvironment */
 /**
  *
  * Project: rdflib.js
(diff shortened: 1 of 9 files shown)


dmitrizagidulin:
@megemege77 - Great catch on the usage of the `$rdf` global. Thanks!

src/fetcher.js
```diff
@@ -37,6 +37,7 @@ const parseRDFaDOM = require('./rdfaparser').parseRDFaDOM
 const RDFParser = require('./rdfxmlparser')
 const Uri = require('./uri')
 const Util = require('./util')
+const serialize = require('./serialize')
```
dmitrizagidulin: I assume you're adding `const serialize = require('./serialize')` to `fetcher.js` because you want to fix the usage of `$rdf.serialize` on line 590?
Do you mind adding that fix to this PR?
(Also, while you're in `fetcher.js`, can you also fix the global usage of `$rdf.Util` on line 867?)

megemege77:
Dear Dmitri

Thanks for the constructive feedback, I'm glad you like the idea of removing the reference to the global. I reworked it to really remove all references, as you suggested, but did it in one comit in a new PR, see linked PR #161.

PR #161, opened 2016-12-26
Removed dependency to global $rdf

megemege77:
This is a rework of https://github.com/linkeddata/rdflib.js/pull/159
- it now **really** removes all references to the global $rdf, as I had originally intended
- I fumbled when working with GitHub and broke the link to the original PR

When including rdflib.js into a web application using webpack, the
resulting application doesn't run. The reason for this is that there are
references to the global $rdf object in modules which are initialized
before the global $rdf object has been properly initialized.

Removed all references to the global $rdf object and
replaced them by direct references to the underlying functionality.


dmitrizagidulin:
@megemege77 - this is great, thanks! Merging.

PR #162, opened 2016-12-27
Fixed the IndexedFormula.substitute bug

timbl:
- some files ferossified afte have been damed by feross standard --format
- Added IndexedFormula.connectedStatements

### DIFF
--- a/src/fetcher.js
+++ b/src/fetcher.js
@@ -617,9 +617,7 @@ var Fetcher = function Fetcher (store, timeout, async) {
         }
       }
       xhr.open(method, uri, true)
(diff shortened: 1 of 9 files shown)


timbl:
Also modified UpdateManager.update() behavior when deleting things with bnodes, when it can't find a WHERE clause using Function Properties or Inverse FPs, to just send all the arcs connected to that bnode in the hope (without checking) that that will be enough to work on the actual data when it arrives.  Works patching bnnodes in VCards.

dmitrizagidulin:
:+1:, merging

PR #164, opened 2017-01-03
Fix a circular dependency

dan-f:
Turns out rdflib is riddled with circular dependencies. Node deterministically handles circular dependencies, but it a) results in modules being incompletely defined (not exporting all of their exports) and b) given the complexity of our dependency graph, it's unreasonably difficult to reason through that mess. Depending on the order that modules are imported, you may see a module fail to export what it should.

This caused a bug in #152, which I'm trying to resolve here in the short term.

There are a lot of these circular dependencies. This PR fixes one in which the `data-factory` module imports `indexed-formula`, which imports `util`, importing `uri`, importing `data-factory`.

One major source of these circular dependencies is the anti-pattern of creating giant index modules which re-export smaller modules. For example using `data-factory`'s `defaultGraph` offers no benefit over using `DefaultGraph` directly.


dan-f:
@dmitrizagidulin plz review :)

@elf-pavlik based on testing, once this merges, your changes (once rebased on top of master) should work.

dan-f:
As an added benefit, I've finally killed off `nodeunit`; all unit tests get to be mocha moving forward :)

dmitrizagidulin:
Nicely done! 👍 on this.

PR #167, opened 2017-02-02
Strip spaces and any params from media type when generating an RDF type

timbl:
Strip spaces and any params from media type when generating an RDF type
Add content type default of turtle to rdf.parse

### DIFF
--- a/src/fetcher.js
+++ b/src/fetcher.js
@@ -584,9 +584,8 @@ var Fetcher = function Fetcher (store, timeout, async) {
   // Returns promise of XHR
   //
   //  Writes back to the web what we have in the store for this uri
-  this.putBack = function (uri, options) {
+  this.putBack = function (uri, options = {}) {
     uri = uri.uri || uri // Accept object or string
-    options = options || {}
     var doc = new NamedNode(uri).doc() // strip off #
     options.data = serialize(doc, this.store, doc.uri, options.contentType || 'text/turtle')
     return this.webOperation('PUT', uri, options)
@@ -594,8 +593,8 @@ var Fetcher = function Fetcher (store, timeout, async) {
 
   // Returns promise of XHR
   //
-  this.webOperation = function (method, uri, options) {
-    uri = uri.uri || uri; options = options || {}
+  this.webOperation = function (method, uri, options = {}) {
+    uri = uri.uri || uri
     uri = this.proxyIfNecessary(uri)
     var fetcher = this
     return new Promise(function (resolve, reject) {
--- a/src/fetcher.js
+++ b/src/fetcher.js
@@ -586,6 +586,7 @@ var Fetcher = function Fetcher (store, timeout, async) {
   //  Writes back to the web what we have in the store for this uri
   this.putBack = function (uri, options) {
     uri = uri.uri || uri // Accept object or string
+    options = options || {}
     var doc = new NamedNode(uri).doc() // strip off #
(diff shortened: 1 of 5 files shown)


src/named-node.js
```diff
@@ -15,7 +15,13 @@ class NamedNode extends Node {
     super()
     this.termType = NamedNode.termType
     if (iri.indexOf(':') < 0) {
-      console.log('Warning: NamedNode IRI "' + iri + '" must be absolute. Relative URIs will fail in future versions')
+      console.log('Error: NamedNode IRI "' + iri + '" must be absolute. Relative URIs will fail in future versions')
+    }
+    if (iri.indexOf(' ') >= 0) {
```
dmitrizagidulin: maybe `iri.includes(' ')` might be slightly clearer/more idiomatic?

dmitrizagidulin:
👍 , looks fine to me.

src/named-node.js
```diff
@@ -15,7 +15,13 @@ class NamedNode extends Node {
...
+      var message = 'Error: NamedNode IRI "' + iri + '" must not contain unencoded spaces.'
+      console.log(message)
+      // throw new Error(message)
+      iri = encodeURI(iri)
```
dmitrizagidulin: Er, wait a second, should this be the right behavior? To automatically `encodeURI` the full node uri?

What's the use case you're encountering, when spaces are being passed in to a named node?

src/fetcher.js
```diff
@@ -586,6 +586,7 @@ var Fetcher = function Fetcher (store, timeout, async) {
   //  Writes back to the web what we have in the store for this uri
   this.putBack = function (uri, options) {
     uri = uri.uri || uri // Accept object or string
+    options = options || {}
```
dan-f: Can alternatively use [ES2015 default argument syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters)

src/named-node.js
```diff
@@ -15,7 +15,13 @@ class NamedNode extends Node {
     super()
     this.termType = NamedNode.termType
     if (iri.indexOf(':') < 0) {
-      console.log('Warning: NamedNode IRI "' + iri + '" must be absolute. Relative URIs will fail in future versions')
+      console.log('Error: NamedNode IRI "' + iri + '" must be absolute. Relative URIs will fail in future versions')
+    }
+    if (iri.indexOf(' ') >= 0) {
+      var message = 'Error: NamedNode IRI "' + iri + '" must not contain unencoded spaces.'
```
dan-f: It could be worth it to use a URL validation library rather these ad-hoc checks.

src/named-node.js
```diff
@@ -15,7 +15,13 @@ class NamedNode extends Node {
...
+      console.log(message)
```
dan-f: Why `console.log()` the message? It seems like constructing a named node from an invalid URI should simply be an error.

src/named-node.js
dmitrizagidulin: Agreed. Shall we open an issue for that?

src/fetcher.js
dmitrizagidulin: Ok, changed it to use default argument syntax.

dmitrizagidulin:
Added tests.

src/named-node.js
dan-f: I mean, might as well do it now? I don't particularly mind either way.

dmitrizagidulin: I opened issue #168 as a reminder.

Issue #168, opened 2017-02-02
Use an IRI validation library in NamedNode constructor


dmitrizagidulin:
Currently (as of PR #167), we're only checking for spaces in named nodes' IRI values, and checking for absolute vs relative. 
As @dan-f mentioned, we should use a dedicated IRI validation library & do this properly.

Neur0mante:
https://github.com/medialize/URI.js is the most comprehensive library for URI parsing and manipulation I've found. Documentation at http://medialize.github.io/URI.js/docs.html
https://github.com/garycourt/uri-js has a bit less functionalites BUT supports conversion to and from IRIs, and still has normalization and parsing which is basically all we need.
For IRI to/from URI manipulations https://github.com/awwright/node-iri is probably the most complete, but it's only for node (shouldnt be too hard to port tho)
URI validation is still kind of a arbitrary cause a lot of URI are "valid" as far as the RFC is concerned but will probably not be interepreted as expected in case the specification is used "loosely".
In the case of the aforementioned file:/// protocol if you try to process `"file://C:/rdfKB/test.xml"` (which is wrong for a couple of reasons, mainly cause the protocol requires the triple slash if there's no host), URI.js returns:
```
  { protocol: 'file',
     username: null,
     password: null,
     hostname: 'C',
     urn: null,
     port: null,
     path: '/rdfKB/test.xml',
     query: null,
     fragment: null,
     duplicateQueryParameters: false,
     escapeQuerySpace: true },
  _deferred_build: false }
```
normalized as: `file://c/rdfKB/test.xml`
Since the syntax says that the first thing after the 2 slashes is the host, that's how C is treated.
Not allowing more than one colon is not a solution cause `urn:isbn:0201896834` is a valid uri. In general URN use colon as separators. Also, `"file:///C:/rdfKB/test.xml"` is probably the preferred way to express a local file on windows as that will correctly be parsed by browsers and node.fs. "file:///C/rdfKB/test.xml" will generally not work, while C|/ is sometimes allowed but discouraged (chrome is fine with it but I haven't tested the various libraries). The exception is the xmlhttprequest node library under windows (that wants 2 slashes) which is frankly just an oversight on their part (at the moment using 3 slashes under windows will fail because of this). It would be probably best to test in the fetcher if the URI is using the file:/// protocol and we are under node, in which case you could use node.fs that is more predictable than the stubbed xhr.
[...]

PR #170, opened 2017-02-24
Build with npm

Neur0mante:
Makefile scripts converted to npm scripts.
The scripts from the main and the tests/serialize makefile have been converted and added as npm scripts.
Added four small libraries for the required functionalities
    "grep-cli": "^0.2.1" for grep
    "node-mkdirp": "0.0.1" for mkdir
    "rimraf": "^2.6.1" for rm -rf
    "jsdiff-cli": "^0.1.2" for diff

This should allow the build and testing process to be platform-independent.
All the scripts are working as intended and so are the tests as far as I can tell.


dmitrizagidulin:
Hi @Neur0mante, what was the original motivation for this? What is it fixing etc?

dan-f:
I agree that the dependency on `make`, `grep`, etc. is a problem. That said, I also think that some of the test scripts in that `Makefile` are quite opaque. Rather than add npm scripts titled `t1` `t2` etc, I'd like to see us accomplish this task as part of #151.

Neur0mante:
The idea is to make the building and testing process platform independent. I'm developing on both windows and linux machines and the shell dependencies are a problem.  I can work on refining the tests, but as I said this wasn't aimed to improve them, just to make them work on windows. Maybe I'll work on that on a separate pull? 
By the way, the travis is failing cause the grep library doesnt work on linux, I'm gonna fix it soon.
I'm almost done translating the whole library to typescript, that's why I needed the tests, still a wip tho.

dmitrizagidulin:
Wow, Typescript? That's pretty intense! Let us know how it goes, btw!

Neur0mante:
It should now work on both linux and windows machines.
So to recap, this leaves all the makefiles unchanged, but allows to also use npm script commands to build and test the library, making it possible to work on nearly any platform. I could duplicate the npm commands to have a linux version if you prefer, since the native grep and diff are most likely faster.
Basically it all comes down to
```
 "scripts": {
    "test": "npm run test:unit && npm run test:serialize",
    "test:unit": "mocha --growl --require babel-register tests/unit/**-test.js",
    "test:unit:dev": "mocha --watch --growl --require babel-register tests/unit/**-test.js",
    "test:serialize": "npm run lib  && cd tests/serialize && npm run all && cd ../.. && npm run cleantest",
    "dist": "mkdirp dist",
    "dist/rdflib.js": "browserify -r ./src/index.js:rdflib --exclude 'xmlhttprequest' --standalone '$rdf' -t [ babelify --presets [ es2015 ] ] -o dist/rdflib.js",
    "dist/rdflib.min.js": "browserify -r ./src/index.js:rdflib --exclude 'xmlhttprequest' --standalone '$$rdf' -d -t [ babelify --presets [ es2015 ] ] -p [ minifyify --no-map --uglify [ --compress [ --dead_code --conditionals --unused --if_return ] --mangle --screw-ie8 ] ] -o dist/rdflib.min.js",
    "all": "npm run dist && npm run dist/rdflib.js && npm run dist/rdflib.min.js",
    "lib": "babel src -d lib",
    "detach": "git checkout origin/master &&  git reset--hard HEAD",
    "clean": "rimraf dist/* lib/*",
    "cleantest": "rimraf tests/serialize/,*",
    "status": "@git branch -v && @git status -s",
    "minify": "npm run dist/rdflib.min.js",
    "size": "node-line-counter .",
    "gh-pages": "npm run detach && git branch - D gh - pages ||: && git checkout - b gh - pages && npm run all && git add - f dist /*.js src/*.js *.js && git commit -m 'gh-pages: update to latest' && git push -f origin gh-pages && git checkout master",
    "prepublish": "npm run lib"
  }
```
as opposed to 
```
# rdflib.js Makefile

PATH:=./node_modules/.bin:${PATH}

all: dist dist/rdflib.js dist/rdflib.min.js

size:
	wc $R

dist/rdflib.js:
	browserify -r ./src/index.js:rdflib --exclude "xmlhttprequest" --standalone "\$$rdf" -t [ babelify --presets [ es2015 ] ] -o dist/rdflib.js

dist/rdflib.min.js:
	browserify -r ./src/index.js:rdflib --exclude "xmlhttprequest" --standalone "\$$rdf" -d -t [ babelify --presets [ es2015 ] ] -p [ minifyify --no-map --uglify [ --compress [ --dead_code --conditionals --unused --if_return ] --mangle --screw-ie8 ] ] -o dist/rdflib.min.js

dist:
	mkdir -p dist

.PHONY: lib
lib:
	babel src -d lib

.PHONY: detach
detach:
	git checkout origin/master
	git reset --hard HEAD

#   WARNING  .. don't do this if you have uncommitted local changes
#
.PHONY: gh-pages
gh-pages: detach
	git branch -D gh-pages ||:
	git checkout -b gh-pages
	make -B
	git add -f dist/*.js src/*.js *.js
	git commit -m 'gh-pages: update to latest'
	git push -f origin gh-pages
	git checkout master

clean:
	rm -f dist/*
	rm -f lib/*

cleantest:
	rm tests/serialize/,*

status:
	@pwd
	@git branch -v
	@git status -s

minify: dist/rdflib.min.js

writable:
	@sed -i -re 's/git:\/\/github.com\//git@github.com:/' .git/config

.PHONY: test
test:	lib
	make -C tests/serialize
	make cleantest

```
and a couple of basic libraries plus the package.json in tests.

dmitrizagidulin:
Looks good overall! Thank you for the PR.

Two minor things: 1) can you remove the stray `console.log` in `tests/serialize/data.js #L102`, and 2) What is the `greptest.ttl` file? Looks like it got committed by accident?

Neur0mante:
Yeah sorry about that

dmitrizagidulin:
np! Thanks!

Neur0mante:
I've added a little fix on the deprecationWarning: "root" is deprecated, use "global"

Neur0mante:
So the problem with doc.uri.slice(0, 8) !== 'file:///' is that on windows process.cwd() doesn't place the starting / on the folder name, which actually turns out to be ok since then you slice "file//" away and fs.writeFile automatically works on both systems. I know the protocol require file:/// but that code actually only lives on those tests, and there's no entry point for a full URI that might be parsed wrong. I was trying very hard to not add any if(onWindows) statements since fs and xmlhttprequest handle the "slash promiscuity" pretty well but I can formalize it more if you want. I've been working on refactoring that whole testing batch with mocha/chai and promises anyway.

About the package.json, that's usually done when you wanna include scripts from 3rd party libraries in your module. The "children" scripts inherit the parent environment with all the libraries and globals, it's pretty much as if you were running them in the main folder. I was just trying to not pollute the main json with too much stuff, but if you prefer I can move everything.

tests/serialize/data.js
```diff
@@ -99,7 +99,7 @@ var doNext = function (remaining) {
             return
           }
           doc = $rdf.sym($rdf.uri.join(right, base))
-          if (doc.uri.slice(0, 8) !== 'file:///') {
+          if (doc.uri.slice(0, 7) !== 'file://') {
```
Neur0mante: So the problem with doc.uri.slice(0, 8) !== 'file:///' is that on windows process.cwd() doesn't place the starting / on the folder name, which actually turns out to be ok since then you slice "file//" away and fs.writeFile automatically works on both systems. I know the protocol require file:/// but that code actually only lives on those tests, and there's no entry point for a full URI that might be parsed wrong. I was trying very hard to not add any if(onWindows) statements since fs and xmlhttprequest handle the "slash promiscuity" pretty well but I can formalize it more if you want. I've been working on refactoring that whole testing batch with mocha/chai and promises anyway.

tests/serialize/package.json
```diff
@@ -0,0 +1,15 @@
+{
```
Neur0mante: About the package.json, that's usually done when you wanna include scripts from 3rd party libraries in your module. The "children" scripts inherit the parent environment with all the libraries and globals, it's pretty much as if you were running them in the main folder. I was just trying to not pollute the main json with too much stuff, but if you prefer I can move everything.

dan-f: Could you post a link to some documentation and examples of this pattern? It's just that I've never encountered it before nor can I find any documentation in the official `npm` docs or even stackoverflow. I'd just like to avoid taking advantage of an unsupported feature.

tests/serialize/data.js
Neur0mante: I was trying to streamline this and I found out that xmlhttprequest on windows require "file://C:/..." with 2 slashes to retrieve files from the disk. 
https://github.com/driverdan/node-XMLHttpRequest/issues/107
Don't ask me why. Again this apparently fit automafically in the fetcher, the problem is that in this way the uri in the literal referring the code is using the 2 slashes format. This is the case in the main library too, if you feed fetcher.nowOrWhenFetched with the 3 slashes uri it will break.
```
NamedNode {
  termType: 'NamedNode',
  value: 'file:///C:/Users/enrico/test.ttl' }
Uri presented to xhr: file:///C:/Users/enrico/test.ttl
Masked error - status 0 for file:///C:/Users/enrico/test.ttl
xhr content:
{ UNSENT: 0,
  OPENED: 1,
  HEADERS_RECEIVED: 2,
  LOADING: 3,
  DONE: 4,
  readyState: 4,
  onreadystatechange: [Function],
  responseText: 'Error: ENOENT: no such file or directory, open \'C:\\C:\\Users\\enrico\\test.ttl\'\n    at Error (native)',
  responseXML: '',
  status: 0,
  statusText:
   { Error: ENOENT: no such file or directory, open 'C:\C:\Users\enrico\test.ttl'
       at Error (native)
     errno: -4058,
     code: 'ENOENT',
     syscall: 'open',
     path: 'C:\\C:\\Users\\enrico\\test.ttl' },
  withCredentials: false,
  open: [Function],
  setDisableHeaderCheck: [Function],
  setRequestHeader: [Function],
  getResponseHeader: [Function],
  getAllResponseHeaders: [Function],
  getRequestHeader: [Function],
  send: [Function],
  handleError: [Function],
  abort: [Function],
  addEventListener: [Function],
  removeEventListener: [Function],
  dispatchEvent: [Function],
  req: BlankNode { termType: 'BlankNode', id: 1, value: '1' },
  original:
   NamedNode {
     termType: 'NamedNode',
     value: 'file:///C:/Users/enrico/test.ttl' },
  options: { userCallback: [Function] },
  resource:
   NamedNode {
     termType: 'NamedNode',
     value: 'file:///C:/Users/enrico/test.ttl' },
  onerror: [Function],
  timeout: 5000,
  actualProxyURI: 'file:///C:/Users/enrico/test.ttl',
  requestedURI: 'file:///C:/Users/enrico/test.ttl',
  ontimeout: [Function],
  handleResponseDone: true,
  headers: {} }
```

tests/serialize/package.json
Neur0mante: The npm team still hasnt given any definitive response about how to make this less ugly
https://github.com/npm/npm/pull/10958

But the technique appear to be used here and there
 http://stackoverflow.com/questions/32783885/is-it-possible-to-use-npm-to-run-scripts-in-multiple-subfolders

http://stackoverflow.com/questions/31773546/the-best-way-to-run-npm-install-for-nested-folders

But moving everthing isn't much of a problem. The other advantage of this format is that the single testing scripts aren't accessible from the main folder, and it will be cleaner to migrate them to the mocha testing later.

dan-f: I think that keeping the serialize scripts out of the top level `package.json` is orthogonal to whether there should be a `tests/serialize/package.json` file. It would be possible, for example, to use a tool like gulp or grunt, or a regular old JS file using node APIs to run scripts from the `tests/serialize/` subdirectory.

The reason I'm pointing this out is that,  far as I can tell, having a `package.json` file at the root of a directory is what defines that directory as an npm package. Since the `tests/serialize` subdirectory is not in fact an `npm` package, I'm hesitant to introduce another `package.json`.

tests/serialize/data.js
dan-f: So on second pass here, it appears that your change is correct; file urls take the form `file://<host>//path`, so I'm happy with this change. (cc @timbl)

Neur0mante: >  A file URI takes the form of
file://host/path
Note that when omitting host, the slash is not omitted (while "file:///foo.txt" is valid, "file://foo.txt" is not, although some interpreters manage to handle the latter).

So the node xhr behaviour is kind of incorrect.
It works, it's just ambiguous on what is loaded on the kb, as the nodes rapresenting the files are different on linux and windows which is not ideal. 

Windows will have
```
NamedNode {
  termType: 'Literal',
  value: 'file://C:\\destination\\folder/file.ttl' }
```
(process.cwd() is also using the reverse backslashes for windows which are then escaped by uri.sym I think. )
while in linux it will be 
```
NamedNode {
  termType: 'Literal',
  value: 'file:///D:/destination/folder/file.ttl' }
```
The slashes can be inverted at will since both node.fs and xmlhttprequest treat them as equal, but the initial slashes are still a problem.
This is a more general issue then the simple tests tho. Maybe it should be sanitized in the fetcher?

tests/serialize/package.json
Neur0mante: Yeah I can see that. I would stay away from building tools if you can since the building is not very complex and maintaining the plugins and scripts for those is too much of a hassle imo, I would rather temporary place the scripts on the package.json. 
I've rewritten a couple of the tests in mocha for now and made data.js into a class that return promises, they should all work async. I'm writing all in typescript cause strong typing saves me some time, but I can just transpile them to es6 or directly in es5 if it's a problem.

dan-f: > I would stay away from building tools if you can since the building is not very complex and maintaining the plugins and scripts for those is too much of a hassle imo

I agree.

> I would rather temporary place the scripts on the package.json.

If you can point me to some documentation that shows this multiple-package-dot-json-in-one-npm-package pattern as a supported feature of npm, I'd definitely consider it.

> I've rewritten a couple of the tests in mocha for now and made data.js into a class that return promises, they should all work async.

Nice! This is something I've really wanted to see happen (#151) but have not had the bandwidth for. I'd like to stay in sync with you on this, since the structure of the test suite will influence how we continue to add tests in the future.

> I'm writing all in typescript cause strong typing saves me some time, but I can just transpile them to es6 or directly in es5 if it's a problem.

I'm a fan of static typing too. That said we have no plans to introduce typescript into `rdflib.js`, so if you would like to get your test refactor merged it would have to be in es2015.

Also, `rdflib.js` unfortunately already contains checked-in source which was transpiled without edits from coffeescript and python (yes, python...). I wouldn't accept transpiled code unless it passed the turing test, if you see what I mean.

Thanks!

Neur0mante: For now I've moved the scripts out, I think the inner package can be considered an exploit until they streamline it so it's probably better to not rely on it.

For the sake of argument I've completed the tests in typescript and then transpiled and checked them by hand (the ts are in tests/serialize-ts, the js in tests/serialize).
https://github.com/Neur0mante/rdflib.js/tree/serialize-tests-in-mocha
 The big difference between ts and the other transpiled sources you used is that ts is written with that sole purpose in mind, it's a superset of js and it can target different ES releases and module systems flowlessly. I've also added the stub for mocha (it's just an npm script: `"test:serialize:ts": "mocha --require ts-node/register --watch-extensions ts tests/serialize-ts/*.spec.ts"`,), so there would be no need of having the transpiled files anywhere as they are hidden inside the tests runner. Just compare the two test-helper and tell me you prefer js. Anyway it's just for show, I can cut all the ts out pretty easily, all the sources are untouched.
You'll find the new suite in tests/serialize, the test-helper is basically the data.js in a class format that return promises, the serialization.spec.js is the actual suite, the ref and source files have been moved in the sample-files folder.  
All the tests should be working asyncronously thanks to chai-as-promised.
You can run the js tests with npm run test:serialize and the ts tests with test:serialize:ts (which for some reason are faster on my machine)

package.json
```diff
@@ -1,82 +1,108 @@
 {
```
dan-f: It looks like your editor might have re-indented this file. It's a bit hard to review the changes here with all the whitespace changes. Could you revert the indentation changes?

package.json
```diff
@@ -1,82 +1,108 @@
...
+        "detach": "git checkout origin/master &&  git reset--hard HEAD",
+        "clean": "rimraf dist/* lib/*",
+        "cleantest": "rimraf tests/serialize/,*",
+        "status": "@git branch -v && @git status -s",
+        "minify": "npm run dist/rdflib.min.js",
+        "size": "node-line-counter .",
+        "gh-pages": "npm run detach && git branch - D gh - pages ||: && git checkout - b gh - pages && npm run all && git add - f dist /*.js src/*.js *.js && git commit -m 'gh-pages: update to latest' && git push -f origin gh-pages && git checkout master",
+        "prepublish": "npm run lib",
+        "serialize-tests": "npm run t1 && npm run t2 && npm run t3 && npm run t4 && npm run t5 && npm run t6 && npm run t7",
+        "t1": "cd tests/serialize && node ./data.js -in=t1.ttl -format=application/rdf+xml -out=,t1.xml && fs-grep http://www.w3.org/2001/XMLSchema#integer ,t1.xml",
```
dan-f: I think these would be a bit clearer (short of a refactor of the tests themselves) if renamed to something along the lines of `serialize-test-1`, `serialize-test-2`, etc.

package.json
```diff
@@ -1,82 +1,108 @@
...
-      "DOMParser",
-      "netscape",
-      "$rdf",
-      "Services",
-      "tabulator",
-      "WebSocket"
-    ]
-  }
-}
+}
```
dan-f: Might be my nitpicky nature, but I'd prefer if we kept our final newlines in all files.

Neur0mante:
Should be fine now. If you wanna check the tests in mocha from my previous comment I can also include them here, just the js part (https://github.com/Neur0mante/rdflib.js/tree/serialize-tests-in-mocha/tests/serialize). Or I can split it in another PR.

dan-f:
Thanks this looks good! The only remaining thought I have is that we should remove the `Makefile`s to remove any ambiguity as to where to add/modify scripts. @timbl would that change be alright with you?

Neur0mante:
There's a problem in the test suit, inside named-node-test.js the describe.only() is disabling all the other tests in the folder when called in a glob from mocha. I've changed it but you might wanna do that in master in the meanwhile.

dan-f:
:man_facepalming: thank you for catching that @Neur0mante

dan-f:
Taken care of in #174.

tests/serialize/data.js
timbl: Am minded to put a test in the NamedNode to reject URIs with more than one colon even though people do it a lot 
Maybe we could make a shim layer on Xhr to map to file:///C/.... for our use internally

dan-f: I propose we migrate this discussion to #168. URI validation is something which has come up several times and is separate from the work here migrating `Makefile` scripts to the `package.json`.

Neur0mante: agreed

Neur0mante:
In the indexed-formula, I've switched from array indexes to Map indexes, loading https://www.wikidata.org/wiki/Special:EntityData/Q2005.ttl went from 15+ sec to 11 sec. It's a pretty straightforward change but I'd prefer to solve this PR before starting another, mostly because I needed the test and I branched from this. https://github.com/Neur0mante/rdflib.js/tree/indexed-formula-with-map
By the way if you are using and external linter like eslint/jslint it would be useful to have your config so the formatting would be more consistent.

dmitrizagidulin:
@Neur0mante (re linter) - we're using feross's [Standard.js](https://github.com/feross/standard) for linting / formatting.

dan-f:
@Neur0mante - could you go ahead and rebase this on `master` and remove the `./Makefile` and `./tests/serialize/Makefile`?

dan-f:
Hey @Neur0mante - note that #172 just merged which added some serializer tests. Could you rebase this on `master`, move the new/modified scripts to the `package.json`, and remove the old `Makefile`s? We should be good to merge at that point!

Neur0mante:
Why is it failing to `nvm install 6.0.0` ?

dan-f:
Here's your failure: https://travis-ci.org/linkeddata/rdflib.js/builds/212043848#L397

Looks like you had a bad merge or rebase with the master branch (check the commit log on this PR).

I recommend using the reflog to reset to where you were prior to the merge with master. At that point you'll want to make sure your master branch is up to date with upstream/master and rebase your branch on top of `master`.

Neur0mante:
I don't think that's the problem tho
https://travis-ci.org/linkeddata/rdflib.js/builds/212043848#L156

dan-f:
I can certainly look into that `nvm` warning but it's clear to me that with 49 commits from several different authors and 22 files changed there's been a bad merge.

Neur0mante:
Ok I fixed that mess but the rebase is still behind a couple of commits, I'm fixing it in couple of hours

Neur0mante:
All new tests added. Needless to say the checks on the file paths fail under windows. Sadly I can't comment on the json but in the new tests format we should use jsdiff in place of riff as that's the npm lib. Also rimraf for rm -f and mkdirp for mkdir -p for the build scripts. I'm not sure if and where to add those info.

package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "rdflib",
   "description": "an RDF library for node.js. Suitable for client and server side.",
-  "version": "0.15.0",
+  "version": "0.14.0",
```
dan-f: This looks like a mistake in your rebase. You'll want to leave the current version at 0.15.0.

package.json
```diff
@@ -1,9 +1,9 @@
...
   "engines": {
-    "node": ">=6.0"
+    "node": "^6.0"
```
dan-f: Another rebase error here.

Neur0mante:
fixed

package.json
```diff
@@ -41,14 +41,44 @@
...
+    "serialize-test-2": "cd tests/serialize && node ./data.js -in=t2.ttl -format=application/rdf+xml  -out=,t2.xml && jsdiff ,t2.xml t2-ref.xml",
+    "serialize-test-3": "cd tests/serialize && node ./data.js -in=t3.ttl -format=application/rdf+xml  -out=,t3.xml && jsdiff ,t3.xml t3-ref.xml",
+    "serialize-test-4": "cd tests/serialize && node ./data.js -in=t3.ttl -out=,t4.ttl && jsdiff ,t4.ttl t4-ref.ttl",
+    "serialize-test-5": "cd tests/serialize && node ./data.js -in=t5.n3 -format=text/turtle -out=,t5.ttl && jsdiff ,t5.ttl t5-ref.ttl",
+    "serialize-test-6": "cd tests/serialize && node ./data.js -in=t5.n3 -format=text/n3 -out=,t6.n3 && jsdiff ,t6.n3 t6-ref.n3",
+    "serialize-test-7": "cd tests/serialize && node ./data.js -in=t7.n3 -format=application/n-triples -out=,t7.nt && jsdiff ,t7.nt t7-ref.nt",
+    "serialize-test-8": "cd tests/serialize && node ./data.js -in=t5.n3  -format=application/n-quads -dump=,t8.nq && jsdiff ,t8.nq t8-ref.nq",
+    "serialize-test-10": "cd tests/serialize && node ./data.js -in=details.ttl -format=text/turtle -out=,t10.ttl && jsdiff ,t10.ttl t10-ref.ttl",
```
dan-f: Looks like you're missing test 9

dan-f:
Even though test 9 isn't working at the moment, I think we will want to leave it in for the eventual serializer test re-write. Once that's done I can merge this!

Neur0mante:
It should be in now

package.json
```diff
@@ -41,14 +41,45 @@
...
+    "serialize-test-5": "cd tests/serialize && node ./data.js -in=t5.n3 -format=text/turtle -out=,t5.ttl && jsdiff ,t5.ttl t5-ref.ttl",
+    "serialize-test-6": "cd tests/serialize && node ./data.js -in=t5.n3 -format=text/n3 -out=,t6.n3 && jsdiff ,t6.n3 t6-ref.n3",
+    "serialize-test-7": "cd tests/serialize && node ./data.js -in=t7.n3 -format=application/n-triples -out=,t7.nt && jsdiff ,t7.nt t7-ref.nt",
+    "serialize-test-8": "cd tests/serialize && node ./data.js -in=t5.n3  -format=application/n-quads -dump=,t8.nq && jsdiff ,t8.nq t8-ref.nq",
+    "serialize-test-9": "cd tests/serialize && node ./data.js -in=t7.n3 -format=application/ld+json -out=,t9.json && jsdiff ,t9.json t9-ref.json",
+    "serialize-test-10": "cd tests/serialize && node ./data.js -in=details.ttl -format=text/turtle -out=,t10.ttl && jsdiff ,t10.ttl t10-ref.ttl",
+    "serialize-test-11": "cd tests/serialize && node ./data.js -in=structures.n3 -format=text/turtle -out=,structures.ttl && jsdiff ,structures.ttl t12-ref.ttl",
+    "serialize-test-12": "cd tests/serialize && node ./data.js -in=t5.n3  -format=application/n-quads -dump=,t8.nq && jsdiff ,t8.nq t8-ref.nq",
```
dan-f: This is the wrong script. Check out the source of test 12.

Also this test fails, but it's not properly failing the travis build. It seems like `jsdiff` doesn't return a failure exit code like `diff` does. Can you look into making `jsdiff` fail for files that don't match?

Neur0mante:
I forked the jsdiff library and added the exit code 1 in case of different files so it will not pass the travis if the tests fail. I haven't bothered publishing the new lib on npm just for that tiny change, so the package.json is now pointing to my fork on github ( `"jsdiff-cli": "Neur0mante/jsdiff-cli"`) Tell me if that acceptable, I will try to make a PR on the original one, or I can publish a new one if we need.

dan-f:
Hey @Neur0mante. I appreciate that you're putting in the time to make this work. I didn't realize that `jsdiff-cli` doesn't already support error exit codes. I've had negative experiences creating dependencies on forked repos, so I'm going to say no to declaring a dependency on `Neur0mante/jsdiff-cli`. If your fork can merge to `jsdiff-cli`, then great, but let's also remember that these tests will eventually be rewritten at a more granular level than file diffing.

Neur0mante:
Yeah that's what I didn't focus too much on them. As I mentioned I've already rewritten them in typescript and converted seemlessly in js here https://github.com/Neur0mante/rdflib.js/tree/serialize-tests-in-mocha/tests/serialize I'm using those at the moment. I still have a minor hiccup with the import of index that I'm trying to solve, but if you think that's a good starting point for a conversion I'll tidy it up in javascript and add the new tests for a PR. It's much more verbose than the old tests obviously, but you have  a lot more control, and they can run all asyncronously.

Neur0mante:
Ditched jsdiff-cli library and instead placed the single script file as diff.js under tests/serialize. This way you can simply call `node diff file1 file2` while in that folder from npm. How about this?

dan-f:
@Neur0mante like I said I don't wish to maintain a fork of `jsdiff-cli`, and placing a modified version of that library directly into this repo is effectively the same thing.

dan-f:
I am running low on bandwidth. I would like to see this PR merge. To do so, it needs to meet the following criteria:
1) Retain existing functionality of the test suite (this is mostly working, but **failing tests must fail the suite** otherwise this change will break our process)
2) Make the tests work across platforms (this is currently working!)
3) Not introduce any technical debt (no new forks or hacks)

Neur0mante:
I mean it's not really a library, it's literally 20 lines of code. I can rewrite them and remove the copyright but it's under MIT anyway.

Neur0mante:
Cleaned it up. The tests are failing in case of a file mismatch, both on linux and on windows. Notice that the tests are failing under windows right now because some of the reference files are sensitive to file paths(usual slash problem). That is fixed in my mocha version. I don't think we should worry about that for now anyway, the build process and the testing infrastructure are both crossplatform and that's what this PR was aiming for.

Neur0mante:
Maybe I wasnt clear but anyway jsdiff-cli is just a small stub for the terminal of the library "diff", which is a well maintained and stable project.

dan-f:
@Neur0mante MIT license requires that the license and copyright notice accompany the redistributed software.

Neur0mante:
```
'use strict';

(function () {
  require('colors')
  var jsdiff = require('diff')
  var fs = require('fs')

  var args = process.argv.slice(2)
  var file1 = args[0]
  var file2 = args[1]

  var data1 = fs.readFileSync(file1, 'utf-8')
  var data2 = fs.readFileSync(file2, 'utf-8')

  var diff = jsdiff.diffChars(data1, data2)

  diff.forEach(function (part) {
    var color = part.added ? 'green'
      : part.removed ? 'red' : 'grey'

    process.stderr.write(part.value[color])
  })
  console.log()
  if (diff.length > 1) {
    process.exit(1)
  }
})()
```
"The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software."
This is all I'm using, and I borrowed part of the code from jsdiff-cli. I don't think this qualifies, but I can rewrite it in a different way if that's a problem.

dan-f:
Just include the copyright & license notices.

Neur0mante:
done

dan-f:
Thanks!

PR #171, opened 2017-02-25
No spaces

timbl:
URIs cannot contain spaces (unless they are encoded as %20)
This has always been the case, but recently we have made it a run-time error to try to generate a named node with them, so various bugs have been surfaced. This PR

- fixes two places where spaces were ending in URIs, eg from link headers

- upgrade the code in serializer.js to pass standard     #technicalcredit

### DIFF
--- a/src/fetcher.js
+++ b/src/fetcher.js
@@ -516,7 +516,8 @@ var Fetcher = function Fetcher (store, timeout, async) {
     } else {
       // See https://www.iana.org/assignments/link-relations/link-relations.xml
       // Alas not yet in RDF yet for each predicate
-      predicate = kb.sym(Uri.join(rel, 'http://www.iana.org/assignments/link-relations/'))
+      /// encode space in e.g. rel="shortcut icon"
+      predicate = kb.sym(Uri.join(encodeURIComponent(rel), 'http://www.iana.org/assignments/link-relations/'))
     }
     if (reverse) {
       kb.add(obj, predicate, xhr.original, why)
--- a/src/serializer.js
+++ b/src/serializer.js
@@ -297,7 +297,7 @@ var Serializer = (function () {
           str += substr
         }
         if (typeof branch === 'string') {
-          if (branch.length === '1' && str.slice(-1) === '\n') {
+          if (branch.length === 1 && str.slice(-1) === '\n') {
             if (',.;'.indexOf(branch) >= 0) {
               str = str.slice(0, -1) + branch + '\n' //  slip punct'n on end
               lastLength += 1
--- a/src/serializer.js
+++ b/src/serializer.js
@@ -30,8 +30,6 @@ var Serializer = (function () {
     this.incoming = null // Array not calculated yet
     this.formulas = [] // remebering original formulae from hashes
     this.store = store
-
-  /* pass */
   }
 
   __Serializer.prototype.setBase = function (base) { this.base = base; return this }
@@ -95,15 +93,13 @@ var Serializer = (function () {
   // Make up an unused prefix for a random namespace
   __Serializer.prototype.makeUpPrefix = function (uri) {
     var p = uri
-    var pok
     function canUseMethod (pp) {
       if (!__Serializer.prototype.validPrefix.test(pp)) return false // bad format
       if (pp === 'ns') return false // boring
       if (pp in this.namespaces) return false // already used
       this.prefixes[uri] = pp
       this.namespaces[pp] = uri
-      pok = pp
-      return true
+      return pp
     }
     var canUse = canUseMethod.bind(this)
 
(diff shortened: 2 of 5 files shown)


src/fetcher.js
```diff
@@ -516,7 +516,8 @@ var Fetcher = function Fetcher (store, timeout, async) {
     } else {
       // See https://www.iana.org/assignments/link-relations/link-relations.xml
       // Alas not yet in RDF yet for each predicate
-      predicate = kb.sym(Uri.join(rel, 'http://www.iana.org/assignments/link-relations/'))
+      /// encode space in e.g. rel="shortcut icon"
+      predicate = kb.sym(Uri.join(encodeURIComponent(rel), 'http://www.iana.org/assignments/link-relations/'))
```
dan-f: I took a look at `http://www.iana.org/assignments/link-relations/` and didn't see any relations with spaces in them. Why is this necessary?

src/serializer.js
```diff
@@ -34,20 +34,20 @@ var Serializer = function () {
   /* pass */
   }
 
-  __Serializer.prototype.setBase = function (base) { this.base = base ; return this}
+  __Serializer.prototype.setBase = function (base) { this.base = base; return this }
```
dan-f: Was this code generated by a transpiler?

src/serializer.js
```diff
@@ -97,38 +96,39 @@ var Serializer = function () {
...
     if (canUse(p.slice(0, 2))) return pok
     if (canUse(p.slice(0, 4))) return pok
     if (canUse(p.slice(0, 1))) return pok
     if (canUse(p.slice(0, 5))) return pok
-    if (! __Serializer.prototype.validPrefix.test(p)) {
+    if (!__Serializer.prototype.validPrefix.test(p)) {
       p = 'n' // Otherwise the loop below may never termimnate
     }
-    for (var i = 0;; i++) if (canUse(p.slice(0, 3) + i)) return pok
+    for (var j = 0; ; j++) if (canUse(p.slice(0, 3) + j)) return pok
```
dan-f: This loop is pretty opaque

src/serializer.js
```diff
@@ -197,8 +197,8 @@ var Serializer = function () {
         // @param force,    "we know this is a root, do it anyway. It isn't a loop."
         function dummyObjectTree(obj, subjects, rootsHash, force) {
             // dump('dummyObjectTree('+obj+'...)\n')
-            if (obj.termType == 'BlankNode' && (subjects[sz.toStr(obj)]  &&
-                (force || (rootsHash[obj.toNT()] == undefined )))) {// and there are statements
+            if (obj.termType === 'BlankNode' && (subjects[sz.toStr(obj)]  &&
```
dan-f: Is this just dead code? If so, it should be deleted. If not do we have a ticket to do something with it?

dmitrizagidulin:
👍 Nice work on the serializer!

src/fetcher.js
timbl: a) when theer could be and you don't want your code to crash and 
b) See the comment ...  e.g. rel="shortcut icon"    was found in real life

dan-f: I'm for `a)`, but isn't `b)` misusing the iana link relations vocabulary? Furthermore `http://www.iana.org/assignments/link-relations/shortcut%20icon` dereferences to a 404.

dan-f: I agree that `NamedNode`s should url encode spaces in their values before serialization. But code that constructs named nodes should be providing URIs which resolve to real data.

Alternatively, I think we should have [`NamedNode.toString`](https://github.com/linkeddata/rdflib.js/blob/master/src/named-node.js#L55-L57) do the url encoding.

src/serializer.js
timbl: I was trying to remember whether it was python or coffee - -maybe this was when coffee  was the thing... .. as I don't see any pyjslib calls.

src/fetcher.js
timbl: 1) The way things evolve in things like HTML tags and link relations and HTTP headers is always a mixture of standards first and code first.
2) yes, and IANA should by linked data principles should provide RDF at the end of the link.  This is a longstanding issue over years.  (That's why the classes derived from mime types I code to be in W3c space.
A to-do is to populate that space in fact with triples  the triples to go int it, they need t e put  in files )
3) Another thing we cna do is make a mirror of what we sould tlike they cto have, and have a list of things whih are not maintained well with pointers to where pothers mainatin a better copy -- just for core things like tis

dan-f: Thanks for the background :smile:

What do you think about having the url encoding done during NamedNode serialization rather than during the construction of NamedNodes?

timbl: Because any string which is not a URL is nighter here nor there, not a well-defined thing.   You can only compare URLs.  Things which random code in places converts into URLs, like fileames, people names, titles of documents, HRRP headers, link relations, whatever.... there is no standard for them, they onlly have well-defined properties in their own defined spaces like file system etc etc.   
When they all come together as URIs they are part of a single universal namespace.

src/serializer.js
timbl: Good point. Yes, I re-coded it, as  (canUse(p.slice(0, 3) + i)) return p.slice(0, 3) + i)   etc

dan-f:
:+1: Thanks @timbl for reminding me that it shouldn't be possible to create a `NamedNode` from an invalid URI in the first place.

PR #172, opened 2017-03-08
Timbl serializer: change to format

timbl:
This makes a number of changes to the way Turtle/N3 is serialized.
When you upgrade, your datasets will go through a large change in their version control systems.

Stability of serialized format: (turtle and rdf/xml)
- Sort the statements by order of subject, then predicate, then object 
- rdf:type at the top of the predicates 

Readability: (turtle only)
- If everything fits put it on one line of 80 chars
- Put predicate and object normally on same line
- indenting
- Use of default namespace so :this instead of <#this>
- Use of turtle numeric literals like 23.6

Added a few tests for different structures and datatypes.


dmitrizagidulin:
:+1:, looks good to me.

Couple of questions, though:

- What does `:this` expand to? Is that some kind of literal or bnode notation?
- What's with the changing timestamps in test `tests/serialize/t8-ref.nq` ?

src/serializer.js
```diff
@@ -5,6 +5,9 @@
 **
 ** Bug: can't serialize  http://data.semanticweb.org/person/abraham-bernstein/rdf
 ** in XML (from mhausenblas)
+**
+** Todo: repeatable order of statments, with rdf:type as first pred.
```
dan-f: Is this TODO (and the above "Bug:") still relevant? It could get forgotten if we don't file an issue

src/util.js
```diff
@@ -310,23 +310,24 @@ function heavyCompare (x, y, g) {
...
-  if ((x.termType === 'BlankNode') || (y.termType === 'BlankNode')) {
+  if ((x.termType === 'BlankNode') && (y.termType === 'BlankNode')) {
     if (x.compareTerm(y) === 0) return 0 // Same
     if (signature(x) > signature(y)) return +1
     if (signature(x) < signature(y)) return -1
     return x.compareTerm(y)  // Too bad -- this order not canonical.
     // throw "different bnodes indistinquishable for sorting"
   } else {
+    if (uriMap && x.uri && y.uri){
+      return (uriMap[x.uri] || x.uri).localeCompare(uriMap[y.uri] || y.uri)
```
dan-f: With the logical change above on L313, `x` or `y` could be a `BlankNode`, in which case `x.uri` might be undefined, which would result in a `TypeError`.

tests/serialize/structures.n3
```diff
@@ -0,0 +1,35 @@
...
+      a :Marriage;
+      :spouse :MLK;
+      :offspring ( :MLK3 :DSK :YK [:name "Bernice King"]);
+    ]
+  ];
+
+  :wikipediapage <https://en.wikipedia.org/wiki/Martin_Luther_King_Jr.>.
+
+:MLK3 :name "Martin Luther King III".
+:DSK :name "Dexter Scott King".
```
dan-f: If we're now sorting by s, then p, then o, then shouldn't this be the first statement in the file?

tests/serialize/structures.n3
```diff
@@ -0,0 +1,35 @@
...
+
+[] :loves [].  # A floating arc between two bnodes
+
+:MLK is :author of  [a :Speech ; :date 1963-08-23; :title "I have a dream"] .
+
+# Datatypes
+:ZooReading1
+  :enabled true;
+  :disabled false;
+  :count 16;
```
dan-f: Same sort concern as above, but now relating to the predicate.

timbl:
@dmitrizagidulin re: "What does :this expand to? Is that some kind of literal or bnode notation?" -- the default allocation of the default namespace is to the local document so  :this is slightly shorter for <#this>

src/serializer.js
```diff
@@ -726,8 +765,9 @@ var Serializer = (function () {
       var results = []
       var type, t, st, pred
       var sts = stats.subjects[this.toStr(subject)] // relevant statements
-      if (typeof sts === 'undefined') {
-        throw new Error('Serializing XML - Cant find statements for ' + subject)
+      if (typeof sts === 'undefined') { // empty bnode
+        return propertyXMLTree(subject, stats)
```
timbl: Yes, an addition to structure.* of a a bnode with a sole incoming arc   like  { :MLK :friend [].  }.  That used not to serialize in XML without throwing the errors.  Added  also another combo:  { [] :loves []. } as a related test.

src/serializer.js
```diff
@@ -533,7 +568,11 @@ var Serializer = (function () {
       }
 
       if (uri.slice(0, j + 1) === this.base + '#') { // base-relative
-        return '<#' + uri.slice(j + 1) + '>'
+        if (canSplit){
+          return ':' + uri.slice(j + 1) // assume deafult ns is local
```
timbl: Good point.  I can't find it in https://www.w3.org/TR/turtle/

tests/serialize/Makefile
```diff
@@ -36,10 +36,12 @@ t7:
 	diff ,t7.nt t7-ref.nt
 
 	# NQUADS t8 ends up including todays datetime in data so not reproducable
-t8:
-	node ./data.js -in=t5.n3  -format=application/n-quads -dump=,t8.nq
-	diff ,t8.nq t8-ref.nq
-	#   NTriple output
+#t8:
```
timbl: It is dumping the entire contents of the store including HTTP metadata.  This doesn't work as a test as it contains the date of the document in a header, which always changes.   The test is not included.  There is no test for the HTTP medatadata in hte store, and no spec either AFAIK. so yes an issue would be appropriate.

tests/serialize/t10-ref.ttl
```diff
@@ -1,47 +1,52 @@
...
-    a    sch:SchedulableEvent;
-    sp:inspiration    det:event1.
-<#event1>
-    dc:author    <#id1442316021316>;
-    dc:title    "TAG 2016 Q1 Meeting ";
-    rd:comment    "Three day meeting, starting date:";
-    ic:location    "Melbourne Australia";
-    sch:availabilityOptions
-       sch:YesNoMaybe;
+:event
```
timbl: Yes, assumption is as though docs all start with
```
@prefix : <#>.
```
Very handy, but can't find any docs

src/serializer.js
timbl: The todo: is now done and the bug is now fixed.  todo: delete the todo

src/serializer.js
```diff
@@ -440,12 +466,21 @@ var Serializer = (function () {
...
       case 'Literal':
+        var val = expr.value.toString() // should be a string already
         if (expr.datatype && this.flags.indexOf('x') < 0) { // Supress native numbers
           switch (expr.datatype.uri) {
+
             case 'http://www.w3.org/2001/XMLSchema#integer':
-              return expr.value.toString()
+              return val
+
+            case 'http://www.w3.org/2001/XMLSchema#decimal': // In urtle must have dot
```
timbl: IN structures.n3 there is the a decimal:     :count 16;  
(maybe would be good to add negative and zero and large tests too, but that tests the functin exists)

src/serializer.js
```diff
@@ -440,12 +466,21 @@ var Serializer = (function () {
...
+              if (val.indexOf('.')  < 0) val += '.0'
+              return val
 
-            // case 'http://www.w3.org/2001/XMLSchema#double': // Must force use of 'e'
+            case 'http://www.w3.org/2001/XMLSchema#double': // Must force use of 'e'
```
timbl: In structure.n3   :sensitivity -1.6507e-6;

src/serializer.js
```diff
@@ -726,8 +765,9 @@ var Serializer = (function () {
       var results = []
       var type, t, st, pred
       var sts = stats.subjects[this.toStr(subject)] // relevant statements
-      if (typeof sts === 'undefined') {
-        throw new Error('Serializing XML - Cant find statements for ' + subject)
+      if (typeof sts === 'undefined') { // empty bnode
+        return propertyXMLTree(subject, stats)
+        // throw new Error('Serializing XML - Cant find statements for ' + subject)
```
timbl: No.

src/serializer.js
```diff
@@ -915,7 +955,9 @@ var Serializer = (function () {
     }
     for (var ns in namespaceCounts) {
       if (!namespaceCounts.hasOwnProperty(ns)) continue
-      str += '\n xmlns:' + this.prefixes[ns] + '="' + escapeForXML(ns) + '"'
+      // Rel uris in xml ns is not strictly allowed in the spec but needed in practice often
+      var ns2 = (this.base && this.flags.indexOf('Z') < 0) ? Util.uri.refTo(this.base, ns) : ns
```
timbl: The people who agreed on the XML namespaces spec insisted on two things against my better judgement.  One was "IT is not a goal to ereference namespace URIs" (which was just unhelpful" and that "all namespace URIs must be absolute".   The problem was the NS spec was done RDF people by XML people who didn't get RDF needs. In the cwm code, a special "z" flag ion the serializer allowed one to break this rule, which is sometimes really valuable.   Here I switched to a "Z" (cap) flag if you want to be strict, default is break the rule. - flip case, flip sense.  We could offer a requested change to the spec at this late date o we could publish a set of spec changes assumed in solid.

tests/serialize/structures.n3
timbl: Yes.. missed that .. it was at one point!!

tests/serialize/structures.n3
timbl: agreed

tests/serialize/structures.n3
timbl: Hang on -- that is the input file, not the output file.

tests/serialize/Makefile
```diff
@@ -1,6 +1,6 @@
 # Crude tests for particular serializer bugs
 
-all: t1 t2 t3 t4 t5 t6 t7
+all: t1 t2 t3 t4 t5 t6 t7  t10 t11 t12 t13
```
timbl: Yes

tests/serialize/Makefile
dan-f: Captured here: https://github.com/linkeddata/rdflib.js/issues/176

tests/serialize/t10-ref.ttl
dan-f: If this is not an official spec, then should we be serializing all of our graphs in this way?

src/serializer.js
dan-f: > todo: delete the todo

:smile:

src/serializer.js
dan-f: Thanks. I should have been more specific by mentioning that the line *below* this one could use a test case (the appending of `'.0'`)

src/serializer.js
dan-f: Could be useful to have a `val` of, e.g. 10 to see it generate `10.0e0`.

src/serializer.js
dan-f: I see.

src/serializer.js
dan-f: Ok thanks for the explanation.

> We could offer a requested change to the spec at this late date o we could publish a set of spec changes assumed in solid.

The latter, at the least, sounds like a good idea. I think it's important that we document places where we "break" from the specs we use so that we can track our divergence.

tests/serialize/structures.n3
dan-f: Whoops! My mistake.

tests/serialize/structures.n3
timbl: That's a hand-written input file, not an output file ... do not expect it to be ordered.

dan-f: Yes, I see now. Whoops!

src/serializer.js
timbl: Yes

src/util.js
timbl: The condition `x.uri && y.uri `on L320 filters out any condition except when we are comparing two things with a `.uri`, i.e. `NamedNode` objects

src/serializer.js
timbl: Good idea. Added a bunch of datatype tests including 10.0e0 in rev 553fb98

src/util.js
dan-f: ah I overlooked the preceding line.

src/serializer.js
dan-f: Excellent!

src/serializer.js
dan-f: Captured here: https://github.com/solid/solid-spec/issues/97

dan-f:
Looking good, @timbl. Nice to see those extra test cases :)

My only remaining question is regarding the assumption of a 'default prefix' of the empty string which maps to the uri of the current named graph. I like the idea of this feature but want to make sure it's included in the specs of the various serialization formats for which we're implementing it.

(Also, to make sure our git tags keep up with the package version specified in the `package.json` it's nice to use `npm version` on `master` after PRs merge rather than as a part of the PR itself)

tests/serialize/t10-ref.ttl
timbl: In the test suite, yes, as otherwise we build in knowledge of what directory the test is run in, an that breaks when someone else runs it, like Travis.
In the systems we build on top of the library, no we should use Turtle not RDF/XML.
In the systems random developers, the subset who decide for some reason to use RDF/XML, then it is hard to say which will be more likely to cause unexpected problems.  After all by default we do in XML attributes use relative URIs (I assume) so they will have to pass a base URI to their parser.  It is just if their parser does not use it on the namespace URIs.   I guess the Jena folks could have systems like that, as IIRC they didn't use the rule of always serializing relative to a base (so they would get massive issues when they moved systems).  Yes, I suppose to be safe we should flip the flag back and pass it in the test suite program data.js (which is an early version of rabel the command line tool)

dan-f: Right, I don't think we should necessarily be serializing the full URIs for the subject in every graph. My question is whether or not it is safe, when using the `:thing` syntax, to omit the `@prefix : <#> .` statement.

timbl:
Ok, I removed the default prefix assumption.  Turtle now includes an explict `@prefix : <#>.`
That clears all comments I think.

dan-f:
nice!

PR #179, opened 2017-03-28
Typescript declaration file (documentation and static typing)

Neur0mante:
This is the first draft of a declaration file in typescript that regroups all the documentations for the public items available when importing $rdf, and also defines the types required and returned from the available functions and methods.
This change will only consist on adding the index.d.ts file in the main folder and the 
` "types": "./index.d.ts"`
 line in the package.json, there is no need to include a typescript library in the project for this to work.

In most IDEs when using typescript this provides strong, statically typed checks at compile time, aside from more fine grained documentation.
In vscode and visualstudio the documentations and type informations will also be available in .js files. 
screenshot_7
screenshot_6

This approach will keep the documentation separate so it can be distributed aside the babeled or minified version and still work.
I could publish this in the definitelyTyped repository if you do not wish to have it inside your project, but that will require some more steps on the user's side while this way is much cleaner.

I have introduced a couple of alias types
```
 export type LiteralTypes = string | Date | number | boolean
 export type Nodeable = LiteralTypes | Node | CollectionArray
 interface CollectionArray extends Array<Nodeable> { } 
```

because some methods (statementsMatching and similar) were becoming too verbose to read. Nodeable just indicates things that are accepted from the `Node.fromValue()` (it also allows for proper checking of the circular definition, as Nodeable can be an array of Nodeable)
I haven't been too strict on the typing of most methods cause that would require a lot of manual inspection, and if I'm wrong the method would not work in typescript.
Also, tagging class members and methods as private will surely improve the code usability through intellsense but again I'm not completely sure which ones are only intended for internal usage.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,1442 @@
+import { XMLHandler } from './ts-src/handlers/XMLHandler';
+
+export = rdflib
+
+declare namespace rdflib {
+  export type LiteralTypes = string | Date | number | boolean
+  export type Nodeable = LiteralTypes | Node | CollectionArray
+  interface CollectionArray extends Array<Nodeable> { }
+  /**
+   * Blank node class. Contains the incremental unique id.
+   * 
+   * @export
+   * @class BlankNode
+   * @extends {Node}
+   */
+  export class BlankNode extends Node {
+    static nextId: number;
+    static readonly termType: string;
+    static readonly NTAnonymousNodePrefix: string;
+    readonly classOrder: number;
+    readonly isBlank: number;
+    isVar: number;
+    id: number;
+    termType: string;
+    value: string;
+    constructor(id?: string);
+    /**
+     * Compare this with another terms.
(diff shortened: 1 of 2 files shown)


woutersl:
Hi, I am using rdflib.js in Typescript, thanks for the great work.
Currently I have to provide my own type definitions to make it work though. The integration of this PR would be a great help.

Would it be possible to clarify what holds back the integration of this proposal and, if appropriate, what should be done to facilitate it?
Thanks!

kjetilk:
Interesting! I'm not the right person to evaluate, but I'm adding @RubenVerborgh as reviewer.

RubenVerborgh:
Me neither; the person who knows the API best is @timbl.

I suspect the API will have changed since this PR.

RubenVerborgh:
Actually, I think the rdflib.js API is changing too fast at the moment to commit to TypeScript, definitely without tests. I would suggest to put this into https://github.com/DefinitelyTyped/DefinitelyTyped

PR #180, opened 2017-04-07
Serialize in mocha

Neur0mante:
I wanted to convert the serialize tests to mocha, and while doing that I had to perform some fixes here and there so this ended up as quite a wide PR.
List of modifications:

- Added a .eslintrc configuration file extending standard. I've tried to specify the rules so it would have the least impact with existing code. I think it would be a good idea to have this shared in the github.

- Moved all the files we are using for the serialization tests in sample_files folder for cleanliness.

- There are a few changes to the fetcher. Most were brought up by the linter.
There were various instances of 
```
if (sf.fetchCallbacks[xhr.resource.uri]) {
                  if (!sf.fetchCallbacks[newURI]) {
                    sf.fetchCallbacks[newURI] = []
                  }
                  sf.fetchCallbacks[newURI] === sf.fetchCallbacks[newURI].concat(sf.fetchCallbacks[xhr.resource.uri])
                  delete sf.fetchCallbacks[xhr.resource.uri]
                }
```
I don't think the triple equal is intended am I missing something?

I added a check for windows so that we feed the xhr with two slashes URIs while keeping three slashes in the kb. It's still possible to insert file URIs with two slashes, but it's no longer necessary to make it work correctly under windows.
```
      if (process) {
        if (process.platform.slice(0, 3) === 'win') {
          actualProxyURI = actualProxyURI.replace(/\/\/\//g, '//')
        }
      }
```

- I've written a test helper which incorporates the functionalities from data.js operating mostly using promises so it can be used asynchronously under mocha.

- I've converted all the current serialization tests into mocha. They are all under the serialization.spec.js file. They work under win and Linux, they will show diffing in case of errors, they will exit with error in case of a mismatch.
This is a pretty bare conversion but there are some more accurate testing for the loaded files in it. 
chai-as-promised allow to work seeminglessly with promises so it's pretty clean and it should all run asynchronously so the suite is relatively fast overall.

- I made the 9th test work through promises since the json+ld converter is async


dmitrizagidulin:
@Neur0mante thank you!

Tell us more about the `.eslintrc` extensions -- what's the intended effect?

package.json
```diff
@@ -30,9 +30,11 @@
   "dependencies": {
     "async": "^0.9.x",
     "jsonld": "^0.4.5",
+    "mz": "^2.6.0",
     "n3": "^0.4.1",
     "xmldom": "^0.1.22",
-    "xmlhttprequest": "^1.7.0"
+    "xmlhttprequest": "1.8.0",
+    "bluebird": "3.5.0"
```
dmitrizagidulin: Why the usage of Bluebird, instead of native promises?

package.json
```diff
@@ -30,9 +30,11 @@
   "dependencies": {
     "async": "^0.9.x",
     "jsonld": "^0.4.5",
+    "mz": "^2.6.0",
```
dmitrizagidulin: Since you're only using `mz` in tests, this should go in devDependencies. Also, I'd prefer it if you used `fs-extra` instead.

package.json
```diff
@@ -41,8 +43,13 @@
     "babelify": "^7.3.0",
     "browserify": "^13.0.0",
     "chai": "^3.5.0",
+    "chai-as-promised": "^6.0.0",
     "diff": "^3.2.0",
+    "eslint": "^3.18.0",
```
dmitrizagidulin: What are these for? The `standard` dep pulls these in, usually.

tests/serialize/sample_files/diff.js
```diff
@@ -24,7 +24,9 @@
 
   var data1 = fs.readFileSync(file1, 'utf-8')
   var data2 = fs.readFileSync(file2, 'utf-8')
-
+  data1 = data1.replace(/\r\n?/g, '\n')
+  data2 = data2.replace(/\r\n?/g, '\n')
+  console.log(data1)
```
dmitrizagidulin: Stray console.log()?

src/fetcher.js
```diff
@@ -585,17 +585,18 @@ var Fetcher = function Fetcher (store, timeout, async) {
   // Returns promise of XHR
   //
   //  Writes back to the web what we have in the store for this uri
-  this.putBack = function (uri, options = {}) {
```
dmitrizagidulin: Why the removal of default arguments here and on line 597?

Neur0mante: I thought this came from a linter warning but itactually probably just came from a diff with an older version of the fetcher.

package.json
Neur0mante: I wasn't sure if you could define .eslintrc files by using the `standard` rep so I was doing it backwards by extending the standard-plugin, but yeah I can reverse that.

package.json
Neur0mante: I'm using mz for the promisified version of the load and write. We could promisify fs-extra by:
```
const Promise = require('bluebird')
const fs = Promise.promisifyAll(require('fs-extra'))
```
(which I think was probably the reason why I started to use bluebird promises in the first place)

Neur0mante:
I'll break down the .eslintrc here

This are just to get support for the various environment:
  "env": {
    "node": true,
    "es6": true,
    "mocha": true
  },
it's worth to point out that this doesn't add support for chai, and the plugins I've found wasn't completely satisfying, so I'm using 
`/* eslint no-unused-expressions:0 */`
inside the tests. This must be enforced cause these
          testHelper.kb.statementsMatching(undefined, undefined, lit).should.not.be.undefined
  "globals": {
    "tabulator": true,
    "Components": true
  },
  "rules": {
    "array-bracket-spacing": [
      "error",
      "always"
    ],
    "space-before-function-paren": "error",
    "camelcase": "off",
    "no-unused-vars": "warn"
  }

package.json
dmitrizagidulin: Yeah, I do hate that `fs-extra` doesn't offer a native promise-based interface (and instead outsources that to promisify). :(
Ok, so, might as well keep `mz` (though move it to devDependencies), and remove the Bluebird dep.

dmitrizagidulin:
@Neur0mante re `.eslintrc` -- all of that stuff is handled by `standard`, so we don't need that file (or the various eslint deps and plugins).

Neur0mante:
Sorry I'm used to ctrl+enter to insert a new line from everywhere I wasn't actually done breaking it down

Neur0mante:
es6 is also needed, I'm testing without it and it's complaining about const, import etc...

dmitrizagidulin:
What's complaining about `const`/import etc? Standard is?

Neur0mante:
nevermind. It was because I removed the extends="standard" from the .eslintrc

Neur0mante:
This should do 
```
{
  "extends": "standard",
  "env": {
    "mocha": true
  },
  "globals": {
    "tabulator": true,
    "Components": true
  },
  "rules": {
    "array-bracket-spacing": [
      "error",
      "always"
    ],
    "camelcase": "off",
    "no-unused-vars": "warn"
  }
}

```

dmitrizagidulin:
Walk me through it, though. What is the .eslintrc for, over out of the box standard?

Neur0mante:
Without the mocha env describe and it get flagged as global variables (which are forbidden in standard)
it's worth to point out that this doesn't add support for chai, and the plugins I've found weren't completely satisfying, so I'm using
/* eslint no-unused-expressions:0 */
inside the tests. This must be enforced cause these
testHelper.kb.statementsMatching(undefined, undefined, lit).should.not.be.undefined
are considered unused.


Same problem here, these are for the tabulator plugin for firefox that pop out here and there:
```
"globals": {
    "tabulator": true,
    "Components": true
  }
```
is it still supported by the way? Otherwise we could remove all those checks.

`rules": {`
standard doesn't enforce a particular spacing in arrays but git/github obviously complains about inconsistencies. This will just place spaces at the start and end of all arrays.
```
    "array-bracket-spacing": [
      "error",
      "always"
    ],
```
There are so many variable/classes that are not camelcased that this is mostly an annyoiance. Same thing with unused vars.
"camelcase": "
"no-unused-vars": "warn"
}

dmitrizagidulin:
The mocha env part and the warnings about the globals are handled in the standard: globals section of `package.json`.

In the case of `testHelper.kb.statementsMatching(undefined, undefined, lit).should.not.be.undefined` considered unused -- which part of it does it complain about being unused? We're widely using chai with standard across most our libs, and hasn't been a problem.

👎 to enforcing array spacing, or silencing camelcasing or unused warnings.

dmitrizagidulin:
To clarify, you should instead add "it" and "describe" to the standard: globals: section of package.json.

dmitrizagidulin:
(I see what you mean about camelcasing warnings, there are a bunch of snake-cased variables in this library due to it being transpiled from Python.)

elf-pavlik:
@Neur0mante for sdandardjs & chai see: https://github.com/feross/standard/issues/690#issuecomment-278533482

Neur0mante:
That looks good, turning all those calls into actual functions makes more sense. With should it plugs in automatically with chai.should() right? It seems to be working.

"no-unused-vars": "warn" was just turning those from errors to warning, but it's not a big issue.
I might work on refactoring the python library out, it doesn't look too hard.

Neur0mante:
It should be fine now.
The correction on the triple equals in the fetcher is fine right?

jeswr:
Closing with thanks: both things this PR set out to do have since happened by other routes — the test suite runs under mocha (`test:unit` in package.json), and the fetcher it patched was completely rewritten in TypeScript (src/fetcher.ts), so none of the diff can be rebased onto the current tree.

PR #184, opened 2017-05-11
Media Type Fix

### LINKED ISSUE #15: Charsets not handled with contentTypes
It seems in many cases contentType contains a charset : 

https://www.w3.org/International/articles/http-charset/index

However this would appear to break node-webid with a rather cryptic error message.  This means that many webids will fail to log in to an unpatched node solid server, leading to inconsistent behaviour and possible user frustration

Possible code to fix may be along the lines of :  

    if (mimeType && mimeType.length) {
      mimeType = mimeType.split(';')[0]
    }

At : https://github.com/linkeddata/node-webid/blob/master/lib/parse.js#L7

webr3:
Fixes https://github.com/linkeddata/node-webid/issues/15 at a rdflib level

### DIFF
--- a/tests/unit/parse-test.js
+++ b/tests/unit/parse-test.js
@@ -9,7 +9,7 @@ describe('Parse', () => {
     describe('literals', () => {
       it('handles language subtags', () => {
         let base = 'https://www.wikidata.org/wiki/Special:EntityData/Q2005.ttl'
-        let mimeType = 'text/turtle'
+        let mimeType = 'text/turtle; charset="UTF-8"'
         let store = graph()
         let content = '<http://www.wikidata.org/entity/Q328> <http://www.w3.org/2000/01/rdf-schema#label> "ангельская Вікіпэдыя"@be-x-old .'
         parse(content, store, base, mimeType)
--- a/src/parse.js
+++ b/src/parse.js
@@ -18,7 +18,7 @@ const Util = require('./util')
  * Hence the mess below with executeCallback.
  */
 function parse (str, kb, base, contentType, callback) {
-  contentType = contentType || 'text/turtle'
+  contentType = Util.baseMediaType( contentType || 'text/turtle' )
   try {
     if (contentType === 'text/n3' || contentType === 'text/turtle') {
       var p = N3Parser(kb, kb, base, base, null, null, '', null)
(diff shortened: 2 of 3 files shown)


dmitrizagidulin:
@webr3 - great work! Any chance you can also add some tests for these?

webr3:
test added

dmitrizagidulin:
thanks!

tests/unit/parse-test.js
```diff
@@ -9,7 +9,7 @@ describe('Parse', () => {
     describe('literals', () => {
       it('handles language subtags', () => {
         let base = 'https://www.wikidata.org/wiki/Special:EntityData/Q2005.ttl'
-        let mimeType = 'text/turtle'
+        let mimeType = 'text/turtle; charset="UTF-8"'
```
elf-pavlik: this tests verifies that parsing *handles language subtags*, i suggest creating another *it* to verify proper handling of mimeType

webr3: this test verifies that `;` and anything that follows is stripped and not considered.

`$rdf.parse` expects contentType to be a string of only the `type-name` and `subtype-name` for example `text/turtle` or `text/html` and so forth, with NO parameters,

Thus this test verifies that parameters, whatever they may be, are stripped and not considered

elf-pavlik: I meant that instead of just modifying one line in existing test, which tests something else - *literals: handles language subtags*. Creating a new dedicated test can make more sense.
Imagine that someone introduces change that will break handling content type headers with parameter, but this test results will show failing *literals: handles language subtags*.
I contributed similar fix to handle content type headers with parameter to solid-web-client and wrote 2 dedicated tests https://github.com/solid/solid-web-client/pull/12/files

webr3: gotcha, yes this would be better - I'll add some more explicit tests when in an IDE, this was just a quick fix and check to unblock somebody - but it turns out solid server is test failing any way so wasn't fruitful :(

dmitrizagidulin: @webr3 can you add a couple more tests when you have a chance?

jeswr:
Closing with thanks: fixed on main — `parse()` strips media-type parameters before dispatching on the content type (src/parse.ts:36, `contentType.split(';')[0]`), so `text/turtle; charset=utf-8` and friends parse correctly, which is what this PR set out to fix.

PR #190, opened 2017-06-30
Fix serialize test instructions

RubenVerborgh:
They would not run on OS X 10.9.5 and Node 8.


dmitrizagidulin:
👍 , merging. Thanks @RubenVerborgh

PR #191, opened 2017-07-18
Timbl downstream - avoid duplicate downstream actions for the same target

timbl:
The code to do this used to compare two functions using (fa === fb)   but that doesn't work as they are in fact different equivalent functions (or maybe function comparison is not a thing) .  So now it makes strings for each one, and compares those.  This is a kludge.

Alternatives include making an action not a function at all, but make it an object like { action: "Refresh"; target: myTable }  so the actions are declarative (generally better but less extensible).

### DIFF
--- a/src/blank-node.js
+++ b/src/blank-node.js
@@ -7,12 +7,12 @@ class BlankNode extends Node {
     super()
     this.termType = BlankNode.termType
     if (id && (typeof id !== 'string')) {
-      throw "Bad id argument to new bkank node: " + id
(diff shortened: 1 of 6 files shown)


dmitrizagidulin:
@timbl I tweaked the BlankNode constructor semantics slightly, to remain compatible with the RDFJS api. Does this logic work for you?

dmitrizagidulin:
Closing - this PR is folded into Fetcher refactor PR #192

PR #192, opened 2017-07-20
Refactor Fetcher to use whatwg fetch()

dmitrizagidulin:


dmitrizagidulin:
@timbl - shall we merge this into master?

PR #193, opened 2017-07-31
Tag42git patch to removeStatements

tag42git:
Patch to removeStatements in indexed-formula.js .  Some statements were not being removed.  Tests are in tests/unit/remove-matches-test.js.


tag42git:
The fix works, but a more durable solution probably requires a deeper understanding of the code.

dmitrizagidulin:
Thanks, @tag42git !

tag42git:
You're welcome.  Glad I could help!

Issue #198, opened 2017-08-24
"Request failed: Cannot find module 'xmldom'"-  when loading data


retog:
I included `https://solid.github.io/releases/rdflib.js/rdflib-0.16.2.min.js`as only script. When invoking `fetcher.nowOrWhenFetched` the third argument of the callback function is: Object {ok: false, error: "Request failed: Cannot find module 'xmldom'", status: undefined}.

retog:
The following code

```
<html>
    <head>
        <script src="https://solid.github.io/releases/rdflib.js/rdflib-0.16.2.min.js"></script>
        <script>
            var store = $rdf.graph()
var timeout = 5000 // 5000 ms timeout
var fetcher = new $rdf.Fetcher(store, timeout)

fetcher.nowOrWhenFetched("https://www.w3.org/People/Berners-Lee/card.rdf", function(ok, body, xhr) {
    console.log("now or when fetched");
    if (!ok) {
        console.log("Oops, something happened and couldn't fetch data");
    } else {
        console.log("Got data");
        //do something with data
    }
})
        </script>
    </head>
    <body>
        <div>content</div>
    </body>
</html>
```
Will first output "Got Data" but when trying to access the store it will output "Oops, [...]" and fail

dmitrizagidulin:
@retog Thanks for the heads up!

I've committed a fix for this issue in PR #199. If you'd like to help test it out, do a git fetch, `git checkout switch-to-webpack`, and run `npm run all`.  You should be able to import dist/rdflib.min.js in your script and try it out.

Also, instead of `nowOrWhenFetched()`, try:

```js
var store = $rdf.graph()
var fetcher = new $rdf.Fetcher(store)

fetcher.fetch('https://www.w3.org/People/Berners-Lee/card.rdf')
  .then(function (response) {
    console.log(response)
    console.log(store.toString())  // print the parsed quads
  })
```

dmitrizagidulin:
@retog Oh, also do an `npm install` after checking out the branch.

dmitrizagidulin:
Fixed in v0.16.3, closing.

PR #199, opened 2017-08-24
Switch from Browserify to Webpack

dmitrizagidulin:
Fixes issue #198


RubenVerborgh:
Should we have the distribution in the repository?

What I usually do is set up a step in Travis to create the build and commit it to a separate (disconnected) branch.

dmitrizagidulin:
Don't have a strong opinion either way. Dist was previously committed, so I kept with that pattern in this PR

RubenVerborgh:
Sure, I just noticed that in this PR. Reasons for not including dist: keeping repository small, no merge conflicts, no wrong commits without matching dist version. Reasons for including it: convenience—but that can be addressed with the CI.

Perhaps I can make a separate PR once this one is accepted.

webpack.config.js
```diff
@@ -0,0 +1,28 @@
+const path = require('path')
+// const webpack = require('webpack')
+
+module.exports = {
+  entry: [
+    './lib/index.js'
```
RubenVerborgh: Why does webpack depend on the compiled `lib`? Since [`babel` is ran along webpack](https://github.com/linkeddata/rdflib.js/pull/199/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR18) anyway, we might want to run webpack on `src` and do additional magic there when needed for the browser.

Drawback: slower compilation time. Advantage: not having to run `npm run lib` first (which we probably want to only run on distribution anyway).

RubenVerborgh:
Another problem is that `package-lock.json` is not updated.

webpack.config.js
dmitrizagidulin: Thanks, will fix.

dmitrizagidulin:
Pointed Webpack main to `src/`, added package-lock.json.

RubenVerborgh:
Ah, and as a bonus, we now also have the right source map (to src instead of lib).

PR #201, opened 2017-08-25
Run all scripts and build dist on Travis

RubenVerborgh:
This pull request introduces two major changes:
- npm scripts are reorganized and renamed for consistency
- the `dist` folder is removed (and ignored) from `master`, and instead built automatically by Travis in another branch

On every push to `master`, an automated commit to the other branch is performed, labeled `Update to {commit_hash}.` This avoids the need of the heavy `dist` folder in the development code, reducing weight and merge conflicts.

Open question: to what branch should the automatic build be pushed?
1. to `gh_pages`: in this case, I propose clearing the existing `gh_pages` branch. Benefit: the resulting script is accessible from http://linkeddata.github.io/rdflib.js/dist/rdflib.min.js
2. to another branch: in this case, could you suggest a name?


.travis_publish
```diff
@@ -0,0 +1,28 @@
...
+# Update pages
+rm -r * 2> /dev/null
+cp -r $REPO_PATH/dist .
+
+# Commit and push latest version
+git add .
+git config user.name  "Travis"
+git config user.email "travis@travis-ci.org"
+git commit -m "Update to $TRAVIS_COMMIT."
+git push -fq origin $PUBLICATION_BRANC 2>&1 > /dev/null
```
dmitrizagidulin: I suspect that's missing an 'H'?

dmitrizagidulin:
Does this build dist on every commit to master, CI style? Or does it only build on npm version changes?

.travis_publish
RubenVerborgh: Aw, yes. Good catch. Trouble is I can't test this thing from a branch.

RubenVerborgh:
Yes, it will build on every commit to master. I can make it only build on tags if you want.

dmitrizagidulin:
I don't know, what do you think is best? On the one hand, seems like every commit is overkill. On the other, hey, Travis has the "CI" part right in the name :)

RubenVerborgh:
In my projects, I build on every commit to master. However, I tend to do development in a separate branch, so in practice, master is only updated on every npm version.

dmitrizagidulin:
Seems legit.

timbl:
I agree pushing every master commit to cleared out gh-pages

RubenVerborgh:
Okay, I will prepare `gh-pages` shortly, double-check, and then merge this PR. (Might still require some tweaks, but I can only see the effect once the CI starts running.)

RubenVerborgh:
Worked on first pass 😄  Here's the first automated commit: 63feeaa20e6bb9e3f3f3de4142778f56854e8231

@dmitrizagidulin Could you please set the `github.io` branch to `gh_pages`  (it's currently `master`, I believe).

dmitrizagidulin:
@RubenVerborgh it's now set to `gh_pages`.

RubenVerborgh:
@dmitrizagidulin Thanks!

PR #211, opened 2017-12-06
Fix parsing of content-type with parameters fixes #210

### LINKED ISSUE #210: content-type "text/turtle;charset=UTF-8" is not handled by parse.js
It is typical for a content-type to contain parameters after a semicolon e.g.

    text/turtle;charset=UTF-8

Currently [parse.js](https://github.com/linkeddata/rdflib.js/blob/master/src/parse.js#L21) does a switch based on the media type and the semicolon will cause the matching to fail

Solution is to introduce a test for a content-type with a parameter, and strip off everything after the semicolon in parse.js

melvincarvalho:
Fixes #210 

Test included

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -19,6 +19,7 @@ const Util = require('./util')
  */
 function parse (str, kb, base, contentType, callback) {
   contentType = contentType || 'text/turtle'
+  contentType = contentType.split(';')[0]
   try {
     if (contentType === 'text/n3' || contentType === 'text/turtle') {
       var p = N3Parser(kb, kb, base, base, null, null, '', null)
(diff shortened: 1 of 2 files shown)


melvincarvalho:
Thanks!

Something in node 6 is breaking the generation of the source map via webpack.

However there is a bug in webpack which doesnt show which file it is : https://github.com/webpack/webpack/issues/5931

Source map generation, however, does work fine in node 8, as reported by travis.

melvincarvalho:
FYI: changing the [devtool](https://github.com/linkeddata/rdflib.js/blob/master/webpack.config.js#L27) setting from source-map to [cheap-source-map](https://webpack.js.org/configuration/devtool/#devtool) seems to fix the build error.

melvincarvalho:
Spoke to @ericprud about the node 6 failing test. We're going to push this fix and add an issue for that. It's webpack related, but hard to track down.

PR #218, opened 2018-01-16
Use child nodes of literal DOM instead of itself (#75)

akoptelov:
This is an alternative fix for the issue #75. Adding dom itself as a literal value is definitely an error since it contains literal definition tag with all its properties.


jeswr:
Closing with thanks: your fix landed via 1f5d00fe (2018) — `frame.addLiteral(dom.childNodes)`, exactly as proposed here — and was later refined by 2eaee8e7 to prefer `innerHTML` in browsers (current code: src/rdfxmlparser.js:334). Issue #75 is fixed and closed.

PR #223, opened 2018-03-12
Looking at status and statusText returned reliably

timbl:


timbl:
After code review with @ericprud mreging this so we can also merge his EGP* branch soon

melvincarvalho:
Perhaps bump the version?

PR #225, opened 2018-03-28
Fix to issue#75 parseType Literal only includes child nodes

jamsden:
RDF/XLM elements with parseType=Literal should not include the RDF/ XML property, only its value in the resulting XML source.


jamsden:
Turtle and RDF/XML representations of the same resource no produce the same internal parsed results.

PR #227, opened 2018-04-09
Fetcher response

timbl:
load() and nowOrWehnFetched() must error (return error) on HTTP errors.  Also the local fetch() renamed as it is not a fetch level thing it is a load, and has different semantics, API, etc


src/fetcher.js
```diff
@@ -911,10 +891,11 @@ class Fetcher {
    * @param options {Object}
    * @param errorMessage {string}
    * @param statusCode {number}
+   * @param response {Response}  // when an fetch() error
    *
    * @returns {Promise<Object>}
    */
-  failFetch (options, errorMessage, statusCode) {
+  failFetch (options, errorMessage, statusCode, response) {
```
ericprud: Is it worth making sure that {RDFXML,XHTML,XML,N3}Handler pass the response? Could add it to their signatures and to ['handleResponse`](https://github.com/linkeddata/rdflib.js/blob/fetcher-response/src/fetcher.js#L1485). This expands the API but has no performance impact. Also, I think it could be marked optional in the API because it's optional in failFetch.

src/fetcher.js
```diff
@@ -251,7 +256,7 @@ class XMLHandler extends Handler {
     // We give up. What dialect is this?
     return fetcher.failFetch(options,
       'Unsupported dialect of XML: not RDF or XHTML namespace, etc.\n' +
-      responseText.slice(0, 80))
+      responseText.slice(0, 80), 901)
```
RubenVerborgh: Magic constants? :smile:

tests/unit/fetcher-egp-test.js
```diff
@@ -35,11 +36,26 @@ describe('Fetcher', () => {
...
       fetcher.nowOrWhenFetched(kb.sym(goodServer + path), {force: true}, trywrap(done, function (ok, statusOrErrorText, resp) {
+        console.log('@@@@@@ ok is ' + ok)
+        console.log('@@@@@@ statusOrErrorText is ' + statusOrErrorText)
+        console.log('@@@@@@ resp is ' + resp)
+        console.log('@@@@@@ resp.status is ' + resp.status)
+
         expect(ok).to.be.false
         expect(statusOrErrorText).to.include(404)
-        expect(resp.error).to.match(/404/)
+        expect(resp.status).to.match(/404/)
```
RubenVerborgh: equal?

tests/unit/fetcher-test.js
```diff
@@ -62,9 +64,36 @@ describe('Fetcher', () => {
...
+      rterm = rdf.namedNode('https://example.com/original.ttl')
+      options = {}
+      userCallback = () => {}
+
+      fetcher = new Fetcher(rdf.graph())
+
+      fetcher.load = sinon.stub().resolves({ ok: true, status: 200, statusText: "Dummy stub"})
+    })
+/*
+    it('should invoke userCallback with caught error', done => {
```
RubenVerborgh: Doesn't work, or not needed?

tests/unit/fetcher-test.js
```diff
@@ -132,16 +161,16 @@ describe('Fetcher', () => {
       uri = 'tel:+1-816-555-1212'
       options = fetcher.initFetchOptions(uri, {})
 
-      return fetcher.fetchUri(uri, options)
-        .then(result => {
-          expect(result.ok).to.be.false()
-          expect(result.error).to.equal('Unsupported protocol')
+      return fetcher.load(uri, options)
+        .then(result => {}, err => {
+          //expect(result.ok).to.be.false()
```
RubenVerborgh: Doesn't work, or not needed?

tests/unit/fetcher-test.js
```diff
@@ -196,10 +225,10 @@ describe('Fetcher', () => {
       it('should fail if the uri fetch previously failed', () => {
         fetcher.getState = sinon.stub().withArgs(uri).returns('failed')
 
-        return fetcher.fetchUri(uri, options)
-          .then(response => {
-            expect(response.ok).to.be.false()
-            expect(response.error.startsWith('Previously failed:')).to.be.true()
+        return fetcher.load(uri, options)
+          .then(response => {}, error => {
+            // expect(response.ok).to.be.false()
```
RubenVerborgh: Doesn't work or not needed?

tests/unit/fetcher-test.js
```diff
@@ -238,39 +267,51 @@ describe('Fetcher', () => {
...
-          expect(result.ok).to.be.false()
-          expect(result.status).to.equal(statusCode)
-          expect(result.error).to.equal(errorMessage)
+    it('should return a result object on HTTP error', () => {
+      return fetcher.load(uri, options)
+        .then(result => {}, err => {
+          console.log('@@@@@@ err is ' + err)
+          console.log('@@@@@@ err.response is ' + err.response)
+          console.log('@@@@@@ err.response.status is ' + err.response.status)
+          console.log('@@@@@@ err.response.statusText is ' + err.response.statusText)
```
RubenVerborgh: Tests instead of logs?

src/fetcher.js
timbl: It would mean that every parser would have to have a copy of the response to it could call it on error?  Or we would wrap each one with a  try/catch which would then add that  information?

tests/unit/fetcher-test.js
RubenVerborgh: In the future, you can always change it into `it.skip`. Then it remains part of the test suite, but is explicitly ignored ("xxx passed tests, 1 pending").

RubenVerborgh:
Let me look into the failing CI; seems an NPM issue.

tests/unit/fetcher-test.js
timbl: This was debugging a test.  Removed.  The test was wrong, in that it tried to nock() the statusText using .reply(status, statusText)  but the API is actually .reply(status, body).  So can't check that the statusText comes through by injecting one.

RubenVerborgh:
I think I have a solution for npm; testing now, will report.

src/fetcher.js
```diff
@@ -788,22 +768,29 @@ class Fetcher {
...
                 ok: false,
                 body: null,
                 bodyUsed: false,
                 size: 0,
                 timeout: 0
-              }, docuri, options)
-      )
+              }
+              console.log('Fetcher: Non-HTTP fetch error ' + error)
+              return this.failFetch(options, 'fetch failed: ' + error, 999, dummyResponse) // Fake status code: fetch exception
```
fletcher91: It might be useful to have a reference to the error when downstream wants to do some logic based on the exception type.

src/fetcher.js
```diff
@@ -1110,8 +1113,42 @@ class Fetcher {
...
+            })
+          } else {
+            resolve(response)
+          }
+        } else {
+          let msg = 'Web error: ' + response.status
+          if (response.statusText) msg += ' (' + response.statusText + ')'
+          msg += ' on ' + method + ' of <' + uri + '>'
+          if (response.responseText) msg += ': ' + response.responseText
+          reject(new Error(msg))
```
fletcher91: Extract to a class?

```
class FetcherError extends Error {
  constructor(uri, options, response) {
    super();
    this.uri = uri;
    this.options = options;
    this.response = response;
    this.message = this.toString();

    // Optionally with improved console logging (https://stackoverflow.com/a/32749533/1630540)
    this.name = this.constructor.name;
    if (typeof Error.captureStackTrace === 'function') {
      Error.captureStackTrace(this, this.constructor);
    } else { 
      this.stack = (new Error(message)).stack; 
    }
  }

  toString() {
    let msg = 'Web error: ' + this.response.status
    if (this.response.statusText) msg += ' (' + this.response.statusText + ')'
      msg += ' on ' + method + ' of <' + uri + '>'
    if (this.response.responseText) msg += ': ' + this.response.responseText
    return msg
  }
}
```

src/fetcher.js
```diff
@@ -1110,8 +1113,42 @@ class Fetcher {
...
+        }
+      }, err => {
+        let msg = 'Fetch error for ' + method + ' of <' + uri + '>:' + err
```
fletcher91: Not sure if the suggested class above should handle this as well, but having a reference to the original error object is useful for further processing. The `toString` function could be updated to return this message if an exception rather than a response was given.

src/fetcher.js
timbl: Good idea.. Maybe HTTPError subclass of FetcherError (WebError?) subclass of Error.
I'd like to land this branch, though now, and do that in a separate branch.    And include where there is one an HTTP error body, parsed if it is turtle, and including the nested error objects where they exist.  (Is there a convention for this/  error.error.erro... ? error.cause ?)

fletcher91: HTTPError < FetcherError seems appropriate.

The property `original` comes to mind since its usage in event targets. Calling it something along the lines of `error` or `serverError` begs the question whether the response body of a failed request actually is an error object or some other resource. So perhaps using the `body` property like the Response class is better.

As an example use-case around body parsing; Our API currently returns custom error resources via nquads, with their subject set to the resource in question, so the client renders an error view. When the request succeeds at a later point, the data is overwritten so the actual resource is rendered. The design enables the client to render errors without having to implement logic to retrieve the status of past requests and extrapolate whether an error has occurred.

Maybe something like the following pattern; If the response Content type is a valid RDF format, parse the body. If the body contains a document-relative resource (ttl `_:` IIRC), add that resource (statement array or converted to js?) to the error object under some key e.g. `resource`. Process additional statements included in the response like the request succeeded (Flagged opt in/out).

On a slightly less related note, I came across [RFC 7807](https://tools.ietf.org/html/rfc7807) which defines a JSON and XML objects for "HTTP API problems" which could act as a class model reference (though SHACL validation reports might also be a common body type).

PR #230, opened 2018-05-13
Allow query() functionality to test for the quad not just the triple.

timbl:
The query() system only worked a triple store not a quad store: if you put a specific value in for the graph (.why) or a variable, it was ignored.  This branch fixes that.  Note that IndexFormula, when you do not specify a graph, adds a default of <chrome:session>.   This acts then as a wildcard when it turns up in a query pattern graph position.

There had been no tests on the query() function. Added 9 basic ones to test particularly around this issue.

### DIFF
--- a/src/fetcher.js
+++ b/src/fetcher.js
@@ -1052,8 +1052,8 @@ class Fetcher {
   putBack (uri, options = {}) {
     uri = uri.uri || uri // Accept object or string
     let doc = new NamedNode(uri).doc() // strip off #
-    options.data = serialize(doc, this.store, doc.uri,
-      options.contentType || 'text/turtle')
+    options.contentType = options.contentType || 'text/turtle'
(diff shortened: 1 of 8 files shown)


RubenVerborgh:
I don't know to what extent the internals follow RDF.js, but there is a DefaultGraph interface, which you might want to use instead of `chrome:theSession`.

elf-pavlik:
I still don't know which interpretation of *Default Graph* rdflib.js follows, see: https://github.com/rdfjs/representation-task-force/issues/117 - *Dedicated Unnamed Default Graph vs. Union Default Graph*

timbl:
@elf-pavlik rdflib.js uses the union interpretation of default graph.  And null or undefined may be used as the default graph name.  store.statementsMatching(s,p,o) gives you all the quads matching the triple pattern s p o no matter where they were loaded from.

PR #234, opened 2018-06-20
Don't pack Node's webcrypto, use browser's crypto instead

maxf:
When using webpack to compile a browser version, we shouldn't rely on Node's `@trust/webcrypto`'s module. Instead, we can use the browser's implementation through the `crypto` object.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -22,7 +22,8 @@ module.exports = {
   },
   externals: {
     'node-fetch': 'fetch',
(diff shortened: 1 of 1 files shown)


fer22f:
Appears they added this with https://github.com/linkeddata/rdflib.js/commit/69c23afc52a194fd0d61d487fb5a27a27eea8a0b, but it would be certainly better if it was properly credited...

PR #235, opened 2018-06-21
Only the first namespace is taken in account

pasqLisena:
That `return this` inside the loop stops it after the first iteration.
It should go AFTER the loop.


melvincarvalho:
Thanks for noticing this.

Travis CI seemed to fail.

Do the tests pass when you run them locally?  (Might be an issue with the tests!)

pasqLisena:
> Do the tests pass when you run them locally? (Might be an issue with the tests!)

Maybe. I have the same error BEFORE and AFTER applying my patch.

And from the text, this is not related to that part of the code.

```
  1) Fetcher nowOrWhenFetched should handle dns error:
     Error: done() invoked with non-Error: trywrap AssertionError: expected 'OK' to match /ENOTFOU
ND/
```

pasqLisena:
@melvincarvalho I see the test. The failing one is [test:serialize:10](https://github.com/linkeddata/rdflib.js/blob/master/package.json#L84) BUT looking at the error in [Travis](https://travis-ci.org/linkeddata/rdflib.js/jobs/452910370#L996) they are:
1. different order in the prefixes list (not influent)
2. different prefixes (e.g. `<http://www.w3.org/2001/XMLSchema#>` is assigned now to `xsd` rather than `XML`).

The point n. 2 actually is the expected behaviour! The `XML` prefix has no sense, considering that in the source file the specified prefix is `xsd`.
And also reading the code, `n3parser.js` [saves in the namespaces](https://github.com/linkeddata/rdflib.js/blob/master/src/n3parser.js#L490) all the prefix-uri coulples, and `serialise.js` [try to apply them](https://github.com/linkeddata/rdflib.js/blob/c0f921e1cfe3722a86549174828d32e31b5358db/src/serialize.js#L21) in output.

In summary, in this case is the test that should change, in particular `t10-ref.ttl`. Should I apply this modification?

pasqLisena:
I applied the described modification to the test file.
Please let me know if further actions are required from my side.

pasqLisena:
Hello @melvincarvalho, any update on this?

pasqLisena:
Hello everyone, this pull request has now been open a long time ago in June 2018.

There is no apparent reason for which it has not been accepted, while the project is going on with new features.

This is starting to be blocking for me. I would like to know if this request will ever be approved or rejected.

megoth:
Just noting that this looks very similar to https://github.com/linkeddata/rdflib.js/pull/389, and they should probably be looked at reviewed together.

As I did with that one, I've added @timbl as reviewer, and notified him of it. I'll update once I know more reg his availability to review this.

pasqLisena:
Yes, indeed, they are 2 PR about exactly the same problem.
Hoping that this is going to be solved (with one PR or with the other).

Thank you for this.

timbl:
This one is indeed more or less the same as https://github.com/linkeddata/rdflib.js/pull/389/   and so let us declare https://github.com/linkeddata/rdflib.js/pull/389/ a duplicate and merge this.

pasqLisena:
Thank you :)

PR #240, opened 2018-07-11
Fix circular dependency & expose a basic namespace map

rescribet:
The function was using a non-existent variable `ns` assuming it's available globally (which it might not).

This was taken from [the map from link](https://github.com/fletcher91/link-lib/blob/master/src/utilities.ts#L55), but adapted and extended with some of the larger [vocabs from lov](https://lov.okfn.org/dataset/lov). Notice that map has been frozen to prevent (accidental) change in behaviour by external sources, which is omitted in this commit.

I was wondering whether the following addition would be useful as well;
```JavaScript
toJS () {
    return Node.toJS(this)
}
```


fletcher91:
Oh, I've caused a circular dependency

kjetilk (commented):
Looks good to me, only comment I have is that I think mostly the 2007 HTTP ontology is used in Solid, and personally, I prefer the dc prefix for dcterms. I'll let others decide whether to merge.

michielbdejong:
LGTM, can you resolve the conflict?

fletcher91:
The conflict has been resolved, I've also added the LDP namespace which was missing

The problem of Node#toJS was fixed in https://github.com/linkeddata/rdflib.js/commit/62da35e6b2518c345b580cc517d77ef89f26ce8b (though that created a circular dependency by a top-level import of `ns` in `Node`), so I've renamed the PR to reflect the changes better

michielbdejong:
@megoth can you review this PR?

kjetilk (changes requested):
Actually, I think I'll request some changes now after all this time. I think the patch should be reviewed for consistency with https://github.com/solid/solid-namespace/blob/master/index.js and I think we should review the way we manage these maps, if that implies deprecating solid-namespace in favor doing it in rdflib like here, or if it should be split into modules of its own. But the first stage, IMHO, is to have consistency between this and solid-namespace.

jeswr:
Closing with thanks: both halves of this are covered on main by other means. The Node/ns circular-import problem was eliminated structurally — the term base class now lives in src/node-internal.ts, with its `fromValue`/`toJS` hooks injected from node.ts specifically to prevent the cycle — and the bundled namespace map is provided by the `solid-namespace` dependency; src/ns.js itself no longer exists, so there is nothing left for this diff to apply to.

Issue #239, opened 2018-07-09
Caching Solid ACL group listings


kjetilk:
In Solid, we want to be able to be able to cache group listings used for ACLs: Tim wrote:

> if you store stuff in RDFlib, then it will be cached for next time if you keep the same quad store. The 
> metadaa from the headers isalso cached so one could check the expiry daye in the header and refresh 
> it if past the expiry date. How to manage expiry dates on group listtings? While the group is changing 
> maybe they should be set to 10 minutes, and then if it becomes static changed to a day or something ..

and 

> rdflib.js has code to refresh the quadstore for a document. It doesn’t deal with expiry date. My 
> preference would be to write the code to check the expiry date and if necessary call the refresh 
> function, and then migrate it into rdflib fetcher.js

timbl:
https://github.com/linkeddata/rdflib.js/pull/241 adds a Fetch method refreshIfExpired 
It refreshes also if there is no record of the expiry date.  No test for it yet

timbl:
Using this issue as a way to discuss cache related issues a bit more broadly, functionality which would  be useful includes
 - Remove old cached documents from the store as a function of age since loaded (and size?)
- Remove old cached documents from the store as a function of age since last used ?? 
  (we don't currently track last-used times)
- Refresh a cache *after* it is used rather than before, if is is quite fresh anyway,  for speed of use
- Refresh old documents (or remove from cache) as a function of HTTP expiry date on the document

There are probably others.

jeswr:
Closing: the concrete ask here landed in 2018 via #241 — `Fetcher.refreshIfExpired` (src/fetcher.ts:1807) refreshes a document when its recorded expiry has passed, and also when no expiry was recorded.

The broader cache-management ideas from the 2020 comment (age-based eviction, refresh-after-use, size limits) never got scoped and don't belong to this issue's original question; if still wanted, please re-file them as a fresh, scoped issue so they can be triaged on their own merits.

PR #241, opened 2018-07-18
Stop cacheing  non-404 errors, and allow webOperation to take a NamedNode

timbl:


timbl:
Should address linkeddata/rdflib.js#250  and also https://github.com/linkeddata/rdflib.js/issues/247

timbl:
There is no test for the new functionality.

timbl:
Now also should address https://github.com/linkeddata/rdflib.js/issues/239
Anyone fancy writing tests for these features to the test set?

kjetilk:
OK, it's a start, but isn't there a more comprehensive caching implementation in some HTTP library? I think it would belong in such a library.

src/update-manager.js
```diff
@@ -621,6 +627,7 @@ class UpdateManager {
   update (deletions, insertions, callback, secondTry) {
     try {
       var kb = this.store
+      var i
```
kjetilk: Why declaring `i` there? It seems like it isn't used outside of the `for` loops further down, and that there is no reason why it shouldn't be scoped to those loops?

RubenVerborgh:
@timbl Highly critical now, but why are we even caching 404?
Because I can imagine issues like "woops, forgot to create that, let me do that right now" but then the initial 404 would still be cached.

Shouldn't we just leave caching to the browser?

RubenVerborgh:
…Or is this mainly for non-browser contexts?

src/fetcher.js
```diff
@@ -1290,13 +1296,32 @@ class Fetcher {
...
   }
 
+ /* refreshIfExpired   Conditional refresh if Expired
+ **
+ ** @param  {NamedNode} term -  An RDF Named Node for the eodcument in question
+ ** @param  {function } userCallback - A function userCallback(ok, message, response)
+ */
+  refreshIfExpired (term, userCallback) {
+    let exp = this.getHeader(term, 'Expires')
+    if (!exp || (new Date(exp).getTime()) <= (new Date().getTime())) {
```
RubenVerborgh: Note for the future: `new Date().getTime() === Date.now()`

src/update-manager.js
```diff
@@ -1,4 +1,4 @@
-// Joe Presbrey <presbrey@mit.edu>
+UpdateManager// Joe Presbrey <presbrey@mit.edu>
```
RubenVerborgh: Surely a mistake?

src/update-manager.js
```diff
@@ -11,8 +11,17 @@ const Serializer = require('./serializer')
...
+** The update manager is a helper object for a store.
+** Just as a Fetcher provides the store with the ability to read and write,
+** the Update Manager provides functionality for making small patches in real time,
+** and also looking out for concurrent updates from other agents
+*/
 class UpdateManager {
   constructor (store) {
+    if (!store) {
+      store = new IndexedFormula() // If none provided make a store
+    }
```
RubenVerborgh: Minor nit: could just be an `store ||  new IndexedFormula()` as above.

src/update-manager.js
```diff
@@ -297,12 +306,12 @@ class UpdateManager {
...
    */
   contextWhere (context) {
-    var sparql = this
+    var updater = this
     return (!context || context.length === 0)
       ? ''
       : 'WHERE { ' +
       context.map(function (x) {
-        return sparql.anonymizeNT(x)
+        return updater.anonymizeNT(x)
```
RubenVerborgh: Minor nit: you dont't need the `this` pointer anymore if using arrow functions.

src/update-manager.js
```diff
@@ -660,11 +670,21 @@ class UpdateManager {
...
       var protocol = this.editable(doc.uri, kb)
-      if (!protocol) {
-        throw new Error("Can't make changes in uneditable " + doc)
+      if (protocol === false) {
+        throw new Error("Update: Can't make changes in uneditable " + doc)
       }
-      var i
-      if (protocol.indexOf('SPARQL') >= 0) {
+      if (protocol === undefined) { // Not enough metadata
+        if (secondTry) {
```
RubenVerborgh: Maybe good to document why we want a second try.

src/update-manager.js
```diff
@@ -660,11 +670,21 @@ class UpdateManager {
...
+          throw new Error("Update: Loaded " + doc + "but stil can't figure out what editing protcol it supports.")
+        }
+        console.log(`Update: have not loaded ${doc} before: loading now...`)
+        this.store.fetcher.load(doc).then( response => {
+          this.update(deletions, insertions, callback, true) // secondTry
```
RubenVerborgh: So we are just guessing the update protocol here then?

tests/unit/update-manager-test.js
```diff
@@ -0,0 +1,119 @@
...
+      options = {}
+      userCallback = () => {}
+
+      updater = new UpdateManager()
+      updater.store.fetcher.webOperation = sinon.stub().resolves({ ok: true, status: 200, statusText: "Dummy stub 1"})
+      // updater.store.fetcher.load = sinon.stub().resolves({ ok: true, status: 200, statusText: "Dummy stub 2"})
+
+      loadStub = sinon.stub(updater.store.fetcher, 'load')
+       .callsFake( doc => {
+         console.log('@@@ Dummy load called 3')
```
RubenVerborgh: Remove?

tests/unit/update-manager-test.js
```diff
@@ -0,0 +1,119 @@
...
+        // expect(updater.store.fetcher.load).to.have.been.calledWith(doc)
+        expect(updater.store.fetcher.webOperation).to.have.been.called()
+        expect(ok).to.be.true()
+        done()
+      })
+    })
+
+    it('Should patch an insert triple with no proior load', done => {
+      updater.update([], [st1], (uri, ok, text) => {
+        console.log(`update callback uri = ${uri}, ok = ${ok}, text = <<<${text}>>>` )
```
RubenVerborgh: Remove?

tests/unit/update-manager-test.js
```diff
@@ -0,0 +1,119 @@
...
+         loadMeta(updater.store)
+         return Promise.resolve({ ok: true, status: 200, statusText: "Dummy stub 5"})
+      })
+
+    })
+
+    it('Should patch an insert triple', done => {
+      loadMeta(updater.store)
+      updater.update([], [st1], (uri, ok, text) => {
+        console.log(`update callback uri = ${uri}, ok = ${ok}, text = <<<${text}>>>` )
```
RubenVerborgh: Remove?

src/fetcher.js
timbl: Meaning?   NO, we are not comparing equal we are comparing one greater than the other.    don't understand that comment.  The above is the stackoverflow recommended way of compating dates

src/update-manager.js
timbl: No, Joe wrote the original module -- I guess we could make it clearer

RubenVerborgh: Sure, but `UpdateManager` is not commented, it's a JavaScript statement. It's harmless because it just evaluates to the class below, but it doesn't do anything so should be removed.

src/fetcher.js
RubenVerborgh: I meant that `(new Date().getTime())` is equivalent to `Date.now()`, so in the future you can just write `new Date(exp).getTime() < Date.now()`.

src/update-manager.js
timbl: No, we are doing a GET to get get the HTTP headers in between, so the second try will normally work as it will know what protocol to use.

tests/unit/update-manager-test.js
timbl: ok

src/update-manager.js
timbl: Fixed all declarations of i and j to be local 'let'.

timbl:
Re @RubenVerborgh "Highly critical now, but why are we even caching 404?
Because I can imagine issues like "woops, forgot to create that, let me do that right now" but then the initial 404 would still be cached.  Shouldn't we just leave caching to the browser?"

This seems to be confused as to the role of RDFlib.  Its function is a cache of the linked data web.  It works by (Fetcher) loading part of it, then operating on it: navigating the graph and doing queries, and then makjng changes with UpdateManager.   

The data is stored parsed and indexed. If we just used the browser's cache, we'd have to parse it evry time we accessed it, and it wouldn't be indexed. 

When it comes to caching 404,s, the way the system works is that there are few files, like all ACL files and index.ttl files, which ca be either there or not there. And bits of code just may need at any time possibly frequently i a lop, check that status. To have to go back across the net to to check that the file wasn't there would be just as ,uch a waste of time as going back across the net to get a new copy when it is there.  Yes, its 404 state can change, but so can its state when it is 200.  Thinks like the "refresh" button in the UI do a force to ignore the cache. But normal run of the mill code uses it.

src/fetcher.js
timbl: Ah I see

src/update-manager.js
timbl: Oh yes that was a typo from missig the Command/F  I suspect ..

src/update-manager.js
timbl: This is atomic to the extent that we need it, I think -- that the thing either succeeds or fails.  If it fails 409 that means the thing you wanted to delete wasn't there.   So you can use things to make a locl, or a semaphore, or a P/V flag as they used to be called.
The server has to implement the atomicity.

RubenVerborgh:
~ If we just used the browser's cache, we'd have to parse it evry time we accessed it, and it wouldn't be indexed.

Makes sense, got the intent now.

> When it comes to caching 404

That said, 404s likely do not need to be parsed, so there are no parsing costs to be saved. And if the server has enabled caching on the 404, the browser will deal with it anyway. So not a whole lot is gained by caching them.

timbl:
And this change is not adding caching for 404.  It is about removing caching for 401 500 999 etc

RubenVerborgh:
> Well, 6 network round tripes are saved when finding the ACL file /.acl for file /a/b/c/d/e/f/g .

Doesn't the browser already save you from those roundtrips? It should cache the 404 itself.

Unless you're saying that a 404 for /a also means that /a/b and /a/b/c will not be attempted.

> It is about removing caching for 401 500 999 etc

Yes, but I'm now wondering at how necessary it is to cache non-RDF representations (which 4xx and 5xx currently are). The argument above was that the parsing was cached, but that's not the case for error pages since there is nothing to parse. And regular response caching is done by the browser.

PR #244, opened 2018-08-02
Fix prefixed namespaces not working beyond rdf

fer22f:
The suggestNamespaces function was just prefixing "RDF" because it's the default one, and quitting on the first result due to a incorrectly placed return.


fer22f:
Ah, ok, I didn't see this was addressed by https://github.com/linkeddata/rdflib.js/pull/235.

PR #246, opened 2018-08-14
Update growl dep

kjetilk:
Github notified of a critical security issue on our version of `growl`, 1.9.2. Not only did we depend on it, but the `mocha` version also  did. `mocha` needed a major version upgrade to fix this (you know, I'm very skeptical that this is good practice), but I figured that most probably, just setting the version for `mocha` to 1.10.5 would also work. So, that is what I tried in the second commit of this pull request.

I hope this looks reasonable.


package.json
```diff
@@ -33,6 +33,7 @@
   "bugs": "http://github.com/linkeddata/rdflib.js/issues",
   "dependencies": {
     "async": "^0.9.x",
+    "growl": "^1.10.5",
```
RubenVerborgh: Grown should not be a dependency; it is a transitive dependency.

Instead of doing this, run an `npm audit`, which will give you the right command to update the transitive dependency.

kjetilk: Uhm, right... `npm audit` is a feature of `npm` 6?

RubenVerborgh: Yes, be sure to be on npm 6 for all Solid projects.

theRealImy:
Dependencies just updated

PR #248, opened 2018-08-22
mocha dependency was updated to avoid vulnerabilities securities

jairo-campos-JD:
Hi @RubenVerborgh, I updated mocha module to fix security issues


jairo-campos-JD:
@RubenVerborgh sorry I just added it.

PR #253, opened 2018-09-17
Build and publish docs on Travis

RubenVerborgh:


RubenVerborgh:
@timbl I can merge in your documentation changes from the current jsdoc branch.

I am opposed to include the /doc folder in the repository itself, since this will easily get out of sync. Travis will always get us the latest docs if this one lands in master.

timbl:
Not sure of your logic that the doc should not be included with the source. After all, 
- we include lib as well as src and it is generated
- it is good o check thjs doc you have written in a new function has worked just as you test the doc. 
- Just as you run npm test you can run npm run doc 
- Useful to compare doc in different branches 
But II'll merge the travis mechanics

RubenVerborgh:
>     * we include lib as well as src and it is generated

Actually, we don't include lib in the git repository: https://github.com/linkeddata/rdflib.js/tree/v0.18.0 (It is included in the npm package.)

>     * it is good o check thjs doc you have written in a new function has worked just as you test the doc.

Agree, but the doc doesn't need to be in the repo to do that.

>     * Just as you run npm test you can run npm run doc

You definitely should; just no need to commit that doc.

>     * Useful to compare doc in different branches

But not so useful to deal with the resulting merge conflicts :wink:

PR #254, opened 2018-09-20
fixed typos in rdflib-block-diagram.svg

TallTed:


TallTed:
@timbl @RubenVerborgh 
The current state of this page is very confusing. The order of actions is not preserved in the presentation, and I cannot tell whether all my typo fixes have been applied (making deleting my branch the right tidying step).

PR #255, opened 2018-09-25
Do not bundle solid-auth-client in browser version

RubenVerborgh:
This pull request brings RDFlib to the same build configuration as Mashlib (https://github.com/linkeddata/mashlib/pull/170).

In addition, it does not bundle solid-auth-client anymore with the browser version generated by webpack, for the reasons detailed below.

### Bundled solid-auth-client is unusable by others
Currently, RDFlib bundles solid-auth-client in order to be able to perform authenticated fetch. However, it does not expose solid-auth-client, so any page using `rdlib.min.js` also needs to include `solid-auth-client.bundle.js` at an additional 182kB for the same functionality.

### Loading multiple instances of the same solid-auth-client version leads to issues
The scenario described above leads to multiple instances of solid-auth-client running on the same page. These instances will interfere with each other, as they will all try to react on a login callback etc.

An example is the [profile viewer tutorial](https://github.com/solid/profile-viewer-tutorial/blob/b5eed339521cf4a74cc9766cd8b1b85360119b7f/index.html#L32-L33), which uses the current RDFlib and solid-auth-client, and thus contains two copies of solid-auth-client.

### Loading multiple different versions of solid-auth-client leads to issues
An even worse case happens when the solid-auth-client version bundled with RDFlib is different from the one on the page. Then they might attempt to write incompatible things to localStorage.

### It ties RDFlib to Solid
The extra 182kB in RDFlib (roughly 50% of its size) do not make sense for the many scenarios where RDFlib is used without Solid.

### In Node.js
Nothing changes—this PR only affects the browser build.

### In the browser, with a library compiled through webpack
Nothing changes, since you define your own build.
In particular, Mashlib is not affected and will still bundle solid-auth-client.

### In the browser for cases where you're not using Solid
Just do as you usually would:
```html
<script src="scripts/rdflib.min.js"></script>`
```
This will use the browser's built-in `fetch`.

### In the browser for cases where you're using Solid
Just do as you usually would, but load solid-auth-client beforehand:
```html
<script src="scripts/solid-auth-client.bundle.js"></script>
<script src="scripts/rdflib.min.js"></script>`
```
Then the authenticated `fetch` from solid-auth-client will be used.

## How this change fits into the Solid ecosystem
Within Solid, we recently started exposing functionality through the `window.solid` namespace. There is:
- `solid.auth` for solid-auth-client
- `solid.data` for @solid/query-ldflex
- `solid.react` for @solid/react

@solid/query-ldflex uses solid-auth-client but does not bundle it; rather, it looks for `solid.auth`.
@solid/react uses both solid-auth-client and @solid/query-ldflex but does not bundle either; rather, it looks for `solid.auth` and `solid.data`.

By preparing (only) our browser distributions to fit into and use these namespaces, we avoid multiple inclusions of the same libraries and the problems arising from that.
So people can just include scripts in their browser and start hacking.

For people developing with Node and webpack, nothing changes: their code is compiled into a single bundle with only a single export, which contains all needed libraries.


RubenVerborgh:
Interesting question from another conversation:

> If people include rdflib.js by itself in a JS file, shouldn’t that be self-sufficient?

It is self-sufficient, for RDFlib functionality.

If they also want Solid auth with it, they need the solid-auth-client lib as well. This makes perfect sense, because just RDFlib bundled with solid-auth-client is not useful: you cannot log in the user, since solid-auth-client is not exposed. So you would need to include an extra copy of solid-auth-client anyway. By not bundling solid-auth-client, only that single copy remains.

Issue #269, opened 2018-10-30
The JSDoc is very broken


timbl:
Much of the doc which is in files like for IndexedFormula should be in the JSDoc stuff, and is not, and a huge number of confusing references to internals within the XML parser are displayed in the JSDoc as globals.  It isn't clear what JSDoc does when it works, so hard to fix this.

timbl:
There is a file generated called "global.html#IndexedFormula" [sic] 
There is a link to http://linkeddata.github.io/rdflib.js/doc/global.html#IndexedFormula

timbl:
Tried changing the name of the file indexed-formula.js to store.js
Does not change what happens.
But a good idea anyway

timbl:
I have a branch jsdoc for this

timbl:
Should have been fixed by https://github.com/linkeddata/rdflib.js/pull/270

PR #270, opened 2018-10-30
Jsdoc working, method moved into subclass where logical, more doc.

timbl:
Also added a turtle intro designed for JS programmers


src/query.js
```diff
@@ -17,7 +17,7 @@
 // Also, users name variables and want the same name back when stuff is printed
 /* jsl:option explicit*/ // Turn on JavaScriptLint variable declaration checking
 
-const IndexedFormula = require('./indexed-formula')
+const IndexedFormula = require('./store')
```
RubenVerborgh: I find this renaming confusing. Is it a store, an indexed formula, both, or are both the same? Seems that they are the same, but would use one term.

timbl: Both  "IndexedFormula" is a formula which has indexes and be queries.   A formula in N3 is a thing in the grammar which you can put in a n3 file or between two { }.  I think devs find that nomenclature .  If we change the names of these classes then we have to leave the old ones in forever more or less.  But that's ok.  I don't know how Store works with the RDFJS folks.   I think Store is just much the most logical for devs.  So we should add Store as === IndexedFormula and move toward it.

RubenVerborgh: It makes sense to me, but that is given my N3 background. At the moment, the mixed import is confusing:

```
const IndexedFormula = require('./store')
```

I would expect either one of these:
```
const Store = require('./store')
const IndexedFormula = require('./indexed-formula')
```

> If we change the names of these classes then we have to leave the old ones in forever more or less.

The name of the `const` is up to the one doing the require.

src/store.js
```diff
@@ -63,8 +65,8 @@ function handleRDFType (formula, subj, pred, obj, why) {
   }
   return done // statement given is not needed if true
 }
-/**
- * IndexedFormula
+/*
+ * PanicPanic
```
RubenVerborgh: ? 😄

kjetilk:
This is the solution to #269, right?

src/store.js
timbl: That was a debug trace for trying to understand  jsdoc

timbl: That was a debug trace for trying to understand  jsdoc  Good catch

src/query.js
timbl: We need people to be able to say 
```
const kb = new $rdf.Store()
or
const kb = $rdf.IndexedFormula()
```
is what I meant

RubenVerborgh: Just changing this in the exports should be fine then.

timbl:
Yes, should solve https://github.com/linkeddata/rdflib.js/issues/269 in removing major brokenness... **doesn't** make the jsdoc complete.   Does allow you to point at things like  fetcher methos and updater methods and store methods

Issue #219, opened 2018-01-24
Functional-style datamodel & performance


rescribet:
In using this library in the browser building our new front-end, I've come across some performance problems regarding to the OO-based types which are used in this library. I'm not sure whether to post this here, or in the task force repo, but since our implementation [uses rdflib.js](https://github.com/fletcher91/link-redux) directly I'll opt for this issue tracker.

In some scenarios, when the data source is to be trusted, incoming data can't just be loaded into the store directly, but it needs to convert all plain objects to instances of `Node`, which can mean some significant performance penalties for slower devices (e.g. mobile browsers).

I ran into this problem when using web workers and when [restoring a graph from indexedDB](https://github.com/fletcher91/persisted_formula). The incoming objects already have memory allocation, but the following conversion into Node objects resulted in a lot of garbage collection.

So I was wondering whether there'd be enthusiasm to approach the inner implementations in a more functional (maybe even immutable) style implementation. If my expectations are correct, it would enable the store to directly reuse incoming objects from response bodies, re-use nodes with the same values, and allow IRI references to be single instance (saving memory). In addition, computational heavy operations (e.g. calculating the hashString) could be handed off to a worker thread (or even the server) and stored in-value (if they are immutable).

From my dives into the code of this library, the changes seem rather straightforward, since the nodes already have a discriminating property which indicates their type (`termType`). But I'm probably missing some important insights. Changing functions like `Node.equals(other: Node)` to `(Node.equals(one, other))` where the arguments are of the form `{ termType: x, value: y}` would at least allow objects from web workers, indexedDB, and fetch to be reused.

timbl:
Interesting.  If it easy and speeds things up then I thinkj we woul dbe open to it so long as it is well documented

timbl:
It would mean though that  x.uri would no longer work for named nodes which would be a pain.
We have the conflict between a clean logical design and a hack to allow parser to save time.  A pity the difference can't be compiled out!

ericprud:
Can you share these changes, perhaps in the form of a PR? It would also be great if you could share your benchmarks so we can acquire those skills.

fletcher91:
Yeah sure, I'll be glad to do that. Might be a few weeks though

jeswr:
[...]

PR #272, opened 2018-11-09
Term memoization

rescribet:
As [per request](https://github.com/linkeddata/rdflib.js/issues/219#issuecomment-435590674) I've integrated some of [the external changes to rdflib](https://github.com/fletcher91/link-lib/blob/v1.0.0-beta27/src/utilities/memoizedNamespace.ts) into the library.

This ~~is still a WIP~~ still contains some minor bugs, but the tests succeed and it works properly when integrated into our [rdflib app](https://app.argu.co/) with minimal changes, ~~but [some additional code](https://github.com/fletcher91/link-lib/blob/v1.0.0-beta27/src/RDFStore.ts#L48) isn't ported yet, which might come this weekend~~. But the overall mechanics and changes are already visible.

The largest change is having to use factory methods rather than constructor calls, IIRC this could be overcome by using ES5 functional types rather than ES6 classes, but I can see value in using factories and leaving the constructor semantics unchanged to prevent unexpected behaviour (e.g. `new Literal('a') === new Literal('a') // => true`).

TODO:
- [x] Check if [remaining proxy code](https://github.com/fletcher91/link-lib/blob/v1.0.0-beta27/src/RDFStore.ts#L48) is still necessary, and port it if true
- [x] Memoize the literals
- [x] Performance benchmarks to [back-up the claims](https://github.com/linkeddata/rdflib.js/issues/219#issuecomment-432998907)
- [ ] (after approval) Update general documentation
- [ ] (after approval) Look into releasing items from memory for extremely long-running processes


fletcher91:
So, I've created a turtle parsing benchmark, and [the results](https://gist.github.com/fletcher91/7d2b06546369b1e47449976bf29e30a1) show about a 40% decrease in memory usage (~55% after garbage collection), and an 8% reduction in processing time.

src/literal.js
```diff
@@ -14,22 +14,10 @@ class Literal extends Node {
       datatype = XSD.langString
     }
     // If not specified, a literal has the implied XSD.string default datatype
-    if (datatype) {
-      this.datatype = NamedNode.fromValue(datatype)
-    }
+    this.datatype = datatype ? NamedNode.fromValue(datatype) : XSD.string
   }
   copy () {
```
fletcher91: This method seems rather nonsensical now, since indexing is done with the value, so either reindexing logic has to be created, or the user should pass intended modifications to this function.

src/literal.js
```diff
@@ -106,7 +93,8 @@ class Literal extends Node {
     } else {
       datatype = XSD.double
     }
-    return new Literal(strValue, null, datatype)
+    return Term.literalByValue(strValue, null, datatype) ||
+      Term.addLit(Term.literalByValue(strValue, null, datatype))
```
fletcher91: `Term.literalByValue` already calls `Term.addLit`, so this call shouldn't be necessary

src/namespace.js
```diff
@@ -1,8 +1,16 @@
 const NamedNode = require('./named-node')
+const Term = require('./term');
 
 function Namespace (nsuri) {
+  Term.namedNodeByIRI(nsuri)
+
   return function (ln) {
-    return new NamedNode(nsuri + (ln || ''))
+    const fullIRI = nsuri + ln
+    if (Term.nsMap[fullIRI] !== undefined) {
```
fletcher91: this can be replaced by `return Term.namedNodeByIRI(fullIRI, ln)`

src/term.js
```diff
@@ -0,0 +1,218 @@
...
+   * @return {NamedNode} The updated NamedNode instance
+   */
+  static addNN(nn, ln) {
+    if (nn.sI) {
+      throw new Error(`NamedNode ${nn} already registered`)
+    }
+
+    nn.sI = ++this.termIndex
+    nn.term = ln
+    this.termMap[nn.sI] = this.nsMap[nn.value] = nn
```
fletcher91: Checking whether `this.nsMap[nn.value]` already contains a value can aid in debugging, but introduces a runtime overhead. If rdflib has a 'dev build' it'd certainly be a good addition

fletcher91:
`Collection` still is missing from memoization since, due to its mutable nature, it might require a different strategy. Giving it a store index (should be system index? since the class spans multiple stores) would allow more performant indexedformula indexing (due to the way number indexes are handled vs string indexes)

michielbdejong:
Awesome work! Trying to write down what you, @vinnl and I just discussed f2f:

I do find it a bit scary that when you do:

```js
var a = new A();
var b = new A();
a.value = 'x';
b.value ='y';
console.log(a.value);
```
You would get 'y' instead of 'x'.

So I think that we might be better off going for nodes that are just objects. So instead of:

```js
var node = new NamedNode('http://example.com/friend')
```
we could maybe just do:
```js
var node  = {
  termType: 'NamedNode',
  value: 'http://example.com/friend'
}
```

That would also be a lot cheaper for passing these to webworkers and back.

elf-pavlik:
Not sure how this PR affects conforming to RDF/JS Data Model. As a note, one of latest changes to that draft provides [`DataFactory#fromQuad`](http://rdf.js.org/data-model-spec/#dom-datafactory-fromquad) which can also take JSON serializable object and upgrade it to RDF/JS conformant object.

To go other way and 'downgrade' RDF/JS conformant object to JSON serializable object I recall issue opened by @RubenVerborgh https://github.com/rdfjs/data-model-spec/issues/94 which might need revising.

> That would also be a lot cheaper for passing these to webworkers and back.

Currently I experiment with prototype based on https://github.com/PolymerLabs/actor-helpers which actually aims at moving as much as possible into web workers. Easy round trip between RDF/JS conformant objects and JSON serializable objects will come there as requirement.

jeswr:
Closing with thanks — not because the idea is wrong, but because the diff is unsalvageable: every file this touches was moved or rewritten in the TypeScript / data-factory migration, so none of it can be rebased. The measured wins were real (~40% memory in your benchmark) and the idea is worth keeping: today the right seam for term interning/memoization is the pluggable data factory (src/factories/), where it can be added without rewriting the term classes. I'll re-file that as a scoped issue under the #824 perf track so the idea isn't lost with this PR.

PR #273, opened 2018-11-10
repairs to 'Literals' in browser [linkeddata/rdflib.js#75]

JeffCave:
Update to allow browser to parse `literal` elements correctly. Basically uses `innerHTML` on the browser, and if it can't find `innerHTML` assumes its on node and uses the current technique

Addresses issue [linkeddata/rdflib.js#75]


jeswr:
Closing with thanks: this landed in substance via 2eaee8e7 — src/rdfxmlparser.js:334 now reads `frame.addLiteral(dom.innerHTML || dom.childNodes)`, i.e. `innerHTML` in browsers with the child-nodes fallback under Node, which is exactly the approach proposed here. The underlying issue (#75) is fixed and closed.

PR #274, opened 2018-11-10
Adds documentation for webpacked web applications

hovenko:
Adds documentation that some transitive dependencies need to be externalized when building webpacked web application.

Related to https://github.com/solid/solid/issues/217

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,26 @@ installed.
 npm install --save rdflib
 ```
 
+#### Webpack web applications
+
+This library has some transitive dependencies to other libraries that are primarily designed for Nodejs servers.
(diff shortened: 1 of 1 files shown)


jeswr:
Closing as stale, with thanks. The 2018 bundling advice no longer matches the tree: the transitive chain it worked around is gone (jsonld@9 removed ky-universal / web-streams — see #685), the build moved to webpack 5, and `@xmldom/xmldom` is already externalized in rdflib's own webpack config (webpack.config.js:31-35). Bundler guidance is being rewritten from scratch as part of the documentation pass under #824, so merging seven-year-old advice would mislead more than help.

PR #277, opened 2018-11-15
store/indexed-formula/formula.anyJS() is like any but converts into native JS - also sync query

timbl:
The superclass Formula does this basic API things
and indexedForumla implements them more efficiently

eg use this to get a number of a Date object from a dataset.

uses Node.toJS()


src/fetcher.js
```diff
@@ -1159,6 +1159,9 @@ class Fetcher {
     return new Promise(function (resolve, reject) {
       fetcher._fetch(uri, options).then(response => {
         if (response.ok) {
+          if (method === 'PUT' || method === 'PATCH'|| method === 'DELETE') {
+            delete fetcher.requested[uri] // invalidate read cache
```
RubenVerborgh: `POST` too (in case you cache folders)?

src/n3parser.js
```diff
@@ -1526,17 +1526,18 @@ __OLD_BadSyntax.prototype.toString = function() {
...
     return "Line %i of <%s>: Bad syntax (%s) at ^ in:\n\"%s%s^%s%s\"" % new pyjslib_Tuple([ ( this.lines + 1 ) , this._uri, this._why, pre, str.slice( st, i), str.slice( i,  ( i + 60 ) ), post]);
-};
+}
+
 function BadSyntax(uri, lines, str, i, why) {
-    let msg =  (  (  (  (  (  (  (  ( "Line " +  ( lines + 1 )  )  + " of <" )  + uri )  + ">: Bad syntax: " )  + why )  + "\nat: \"" )  + str.slice( i,  ( i + 30 ) ) )  + "\"" ) ;
-    let e = new Error(msg)
-    e.lineNumber = lines + 1
+    let lineNo = lines + 1
+    let msg =  (  (  (  (  (  (  (  ( "Line " +  ( lineNo )  )  + " of <" )  + uri )  + ">: Bad syntax: " )  + why )  + "\nat: \"" )  + str.slice( i,  ( i + 30 ) ) )  + "\"" ) ;
```
RubenVerborgh: Maybe the parentheses can go? 😉

src/n3parser.js
```diff
@@ -1526,17 +1526,18 @@ __OLD_BadSyntax.prototype.toString = function() {
...
+    let e = new SyntaxError(msg, uri , lineNo)
+    e.lineNo = lineNo
```
RubenVerborgh: For future reference, interesting trick:
```
return Object.assign(new SyntaxError(msg, uri, lineNo),
  { lineno, characterInFile: i, syntaxProblem: why })
```

src/node.js
```diff
@@ -74,23 +78,27 @@ Node.fromValue = function fromValue (value) {
...
+const Namespace = require('./namespace')
+const ns = { xsd: Namespace('http://www.w3.org/2001/XMLSchema#') }
+
+Node.toJS = function toJS (term) {
   if (term.elements) {
     return term.elements.map(Node.toJS) // Array node (not standard RDFJS)
   }
   if (!term.datatype) return term // Objects remain objects
-  if (term.datatype.equals(ns.xsd('boolean'))) {
+  if (term.datatype.sameTerm(ns.xsd('boolean'))) {
```
RubenVerborgh: Just wondering about why this changed, whereas RDF/JS is still using `equals`.

src/update-manager.js
```diff
@@ -349,13 +349,15 @@ class UpdateManager {
 
         console.log('UpdateManager: update Ok for <' + uri + '>')
 
-        callback(uri, response.ok, response.responseText, response)
+        callbackFunction(uri, response.ok, response.responseText, response)
       })
       .catch(err => {
-        callback(uri, false, err.message, err)
+        callbackFunction(uri, false, err.message, err)
```
RubenVerborgh: Might want to consider `(error, result)` for callbacks in the future, where `result` is an object that combines the multiple things.

src/fetcher.js
timbl: good point

src/node.js
timbl: I guess sameTerm I understand but equals looks like it should be the same as === or == which could require the object to be the same object ... not sure why I ended up changing it.   Happy either way

RubenVerborgh: I would recommend `equals`; it's not `===` but rather "do these objects represent the same thing", i.e., will they reasonably behave the same in all circumstances?

src/n3parser.js
timbl: another day

PR #282, opened 2018-11-21
Fix an issue where serializer flags didn't affect strings

rescribet:
The `stringToN3` method didn't account for flags set on the object, which caused multiline strings to be serialized in n-triples.


michielbdejong:
LGTM

jeswr:
Closing with thanks: this exact fix landed on main via #652 (0d9e8a81, fixes #650) — `atomicTermToN3` now passes `this.flags` into `stringToN3` (src/serializer.js:583), so serializer flags affect string escaping as this PR intended six years earlier. Sorry it sat so long.

Issue #296, opened 2019-02-04
Fetcher.load saves metadata to graph even though `noMeta` is set to true


megoth:
The metadata for the response is saved in the store irregardless of noMeta being set or not.

PR #297, opened 2019-02-04
A very crude fix, but should do the trick

megoth:
Should fix https://github.com/linkeddata/rdflib.js/issues/296


megoth:
As this hasn't been merged yet, I'm choosing to close it.

PR #298, opened 2019-02-06
Ignore deletes that do not match

RubenVerborgh:
Delete operations that do not match any triples should silently succeed instead of erroring.

See https://github.com/solid/node-solid-server/issues/1085

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -141,34 +141,18 @@ class IndexedFormula extends Formula { // IN future - allow pass array of statem
         // console.log('ds before substitute: ' + ds)
         if (binding) ds = ds.substitute(binding)
         // console.log('applyPatch: delete: ' + ds)
-        ds = ds.statements
-        var bad = []
-        var ds2 = ds.map(function (st) { // Find the actual statemnts in the store
+        ds.statements.forEach(function (st) { // Find the actual statemnts in the store
           var sts = targetKB.statementsMatching(st.subject, st.predicate, st.object, target)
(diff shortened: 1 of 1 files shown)


src/store.js
```diff
@@ -141,34 +141,18 @@ class IndexedFormula extends Formula { // IN future - allow pass array of statem
...
           var sts = targetKB.statementsMatching(st.subject, st.predicate, st.object, target)
-          if (sts.length === 0) {
-            // log.info("NOT FOUND deletable " + st)
-            bad.push(st)
-            return null
-          } else {
-            // log.info("Found deletable " + st)
-            return sts[0]
+          if (sts.length > 0) {
+            targetKB.remove(sts[0])
```
RubenVerborgh: @timbl Unsure why we are only removing the first matching statement—I think it should be all. But this follows the original code.

RubenVerborgh:
From a discussion with @timbl:

> It is a critical part of the architecture that patches which delete non existent triples fail.

> Otherwise we have no control of concurrent editing

RubenVerborgh:
Alternative implementation in https://github.com/linkeddata/rdflib.js/pull/299

jeff-zucker:
There was lengthy discussion on this with the consensus being that an attempt to delete a triple that doesn't exist should return a 409 because it might indicate multiple editors on the same resource, see https://github.com/solid/node-solid-server/issues/1085#issuecomment-464892054

PR #299, opened 2019-02-07
Add mustExist flag to patch object

RubenVerborgh:
This PR is an alternative to #298.

Delete semantics are _not_ changed by default. Instead, a patch flag `mustExist` is added; if that flag is set to a falsy value, then a DELETE with a non-matching WHERE clause will silently not cause any effect.

Needed to address the incorrect SPARQL UPDATE semantics in https://github.com/solid/node-solid-server/issues/1085.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -135,36 +135,12 @@ class IndexedFormula extends Formula { // IN future - allow pass array of statem
     const { delete: deletions, insert: insertions, where } = patch
     const mustExist = 'mustExist' in patch ? !!patch.mustExist : true
 
-    function doPatch(binding, onDonePatch) {
-      if (deletions) {
-        const graph = !binding ? deletions : deletions.substitute(binding)
-        const toRemove = []
-        const notFound = []
-        graph.statements.forEach(function (st) { // Find the actual statemnts in the store
-          const sts = targetKB.statementsMatching(st.subject, st.predicate, st.object, target)
-          if (sts.length === 0) {
-            notFound.push(st)
-          } else {
-            toRemove.push(sts[0])
-          }
-        })
-        if (mustExist && notFound.length !== 0) {
-          return patchCallback('Could not find to delete: ' + notFound.join('\n or '))
-        }
-        toRemove.forEach(function (st) {
-          targetKB.remove(st)
-        })
-      }
-      if (insertions) {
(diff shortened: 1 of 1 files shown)


timbl:
Where is the flag passed?  What is its default value?

RubenVerborgh:
> Where is the flag passed?

On the `patchObject`, so by whatever parses that object.
The idea would be to set it to `false` for SPARQL UPDATE patches, and to an explicitly provided value in N3 patches.

> What is its default value?

`true`, so existing behavior is preserved.

timbl:
What will stop this flag breaking other implementations?

timbl:
So NJ3 patches would work as semaphores and Sparql Update patches woud not -- the mime type would be the key?  And Solid-ui code would have to switch to n3 patches immediately ?

timbl:
Who would change Gold?

RubenVerborgh:
> What will stop this flag breaking other implementations?

It is set to `true` by default, which is the old behavior.

> So N3 patches would work as semaphores

N3 patches be able to specify a property `solid:mustExist`.
If set to `true` (default), they work as semaphores.
If set to `false`, they follow the SPARQL UPDATE behavior.

> Sparql Update patches would not

Correct.

> the mime type would be the key?

Correct (but behavior for N3 is overridable with `solid:mustExist`).

> And Solid-ui code would have to switch to n3 patches immediately ?

They need to switch to N3 patches, as a correct implementation of SPARQL UPDATE does not have the semaphore behavior.

> This means that any old solid-ui apps will fail -- can use user data -- with new servers.

Maybe, it depends how they do their updates.
If they use the rdflib.js update manager (i.e., not directly create SPARQL bodies themselves), they should be alright.

> Who would change Gold?

I'm afraid I cannot help with that—but do we have many instances running still?

jeswr:
Closing as superseded, with thanks. The question this PR answered ad hoc — what should happen when a patch's WHERE/conditions don't match — has since been settled at the spec layer: the Solid Protocol specifies N3 Patch, whose condition-matching semantics (reject on non-match) are defined server-side, so a client-side per-patch `mustExist` flag is no longer the right mechanism. The file this patched (src/store.js) was also rewritten in the TypeScript migration, so the diff no longer applies.

PR #306, opened 2019-03-17
Update fetcher.js to conditionally require solid-auth-cli in browserless mode

jeff-zucker:


megoth:
Maybe add `solid-auth-cli` as dependency in package.json as well?

RubenVerborgh:
Will rebase on top of the anyJS branch as requested by @timbl.

RubenVerborgh:
I had a crash building it locally; let's see what Travis does.

PR #312, opened 2019-04-22
Add NamedNode.id as the local part of the URI, and fix getHeaders to …

timbl:
…not look at error responses.


michielbdejong:
Sorry that i didn't find time to review it on Friday. I will be faster next time!

PR #321, opened 2019-06-06
Update webapp-intro.html

TallTed:
a little light cleanup

### DIFF
--- a/Documentation/webapp-intro.html
+++ b/Documentation/webapp-intro.html
@@ -33,24 +33,24 @@ td: { background-color: #eee; }
 <body>
 <h2>What is rdflib?</h2>
 
-<p>The easiest and best way to work with linked data in Solid is to use a library called rdflib. Rdflib is a general toolbox for doing most things for linked data. It can store data, parse and serialize data into various formats, and keep track of changes to the data coming from the app or from the server.</p>
+<p>The easiest and best way to work with linked data in Solid is to use a library called <code>rdflib</code>. <code>rdflib</code> is a general toolbox for doing most things with linked data. It can store data, parse and serialize data into various formats, and keep track of changes to the data coming from the app or from the server.</p>
 
 <h3>Glossary of Terms</h3>
 
 <p>Here are some basic vocabulary terms we'll be using throughout this document.</p>
 
-<ul><li><b>Store</b> - data structure to store graph data and perform queries against. This is the simplest way to work with linked data in rdflib. You can store data from Javascript, dump data out from it, or perform raw queries.</li>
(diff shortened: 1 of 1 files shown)


Documentation/webapp-intro.html
```diff
@@ -92,17 +92,17 @@ <h3>Using the Store</h3>
 <code class="language-javascript">const VCARD = new $rdf.Namespace(‘http://www.w3.org/2006/vcard/ns#‘);
 </code></pre>
 
-<p>If we don’t  know which vocabulary to use, various groups have their favorite lists. One is the <a href="https://github.com/solid/solid-ui/blob/master/src/ns.js">solid-ui list of namespaces</a>.</p>
+<p>If we don’t know which vocabulary to use, various groups have their favorite lists. One is the <a href="https://github.com/solid/solid-ui/blob/master/src/ns.js">solid-ui list of namespaces</a>.</p>
```
TylerRick: This link is outdated now (as of https://github.com/solid/solid-ui/commit/95e298d289e5750cfc40c153fa248b93da5f0909#diff-3f43ebf4bdca62f4bf680703836e1e88).

It should link to: https://github.com/solid/solid-namespace/blob/master/index.js

TallTed: Funny how PRs that are ignored for several weeks get a little stale...

Although, as I didn't add that link, it is even staler than that, and fixing it seems like it ought to be its own PR.

TallTed:
@angelo-v @RubenVerborgh -- Almost 2 years on, I wonder whether this might get merged, so that further patches can be made?  The branch it's based on was inadvertently deleted, making it difficult if not impossible to tweak this PR (which I think leaves the target in better shape than it is now, even without any further tweaks).

RubenVerborgh:
I’m not the right person to tag here.

TallTed:
Sorry, @RubenVerborgh, looked at the wrong commit history.

Untagged in my previous, but can't inform you of that without tagging you again :-/

bourgeoa:
@jeff-zucker I think you have write access

Documentation/webapp-intro.html
jeff-zucker: @TallTed  can you resubmit the PR with the better URL for the namespace mentioned by above : https://github.com/solid/solid-namespace/blob/master/index.js Then I will merge it.

TallTed: @jeff-zucker - I will replace the URL in another PR later today since no-one else seems to want to. (I don't understand why TylerRick, and now you, didn't simply make another PR to fix that stale URL, since it's not related to my PR at all.) 

Please merge this one now, as I cannot change it, as the source repo was inadvertently deleted, as I said earlier today.

Issue #324, opened 2019-07-12
Fetcher.putback breaks for contentType JSON-LD


akashdeep-singh:
The `putback` method breaks if contentType is `application/ld+json`. This is because it calls the `serialize` function which accepts an optional callback which seems to be required only in case of JSON-LD. Refer the following lines:

https://github.com/linkeddata/rdflib.js/blob/a15d8768d82c452400f11e4d9747a8e43a846db0/src/fetcher.js#L1072

https://github.com/linkeddata/rdflib.js/blob/a15d8768d82c452400f11e4d9747a8e43a846db0/src/serialize.js#L44

jeff-zucker:
Fixed by https://github.com/linkeddata/rdflib.js/commit/a24dc44ef7bb31ce230f1fb96d44b85580ddc311 which encorporates the changes suggested by @akashdeep-singh .

PR #325, opened 2019-07-12
Fix for #324: Fetcher.putback breaking for JSON-LD 

akashdeep-singh:
Also contains one minor typography correction in fetcher.js (fata -> data).


jeff-zucker:
This PR had gotten stale because rdflib switched to Typescript.  The PR is needed though, I added a minor improvement and merged the changes directly into the current rdflib main branch. @akashdeep-singh - thanks for your contribution!

PR #330, opened 2019-08-02
One line untested fix to problem of no quad.why in jssonld parsing

timbl:
See https://gitter.im/linkeddata/rdflib.js?at=5d407b4784bc7a717fee19cd

I notice parsing turtle with rdflib yields a 'why' named node, but parsing jsonld does not. I think the why named node is important for app development. Is this intentional or a bug?

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -20,6 +20,8 @@ const Util = require('./util')
 function parse (str, kb, base, contentType, callback) {
   contentType = contentType || 'text/turtle'
   contentType = contentType.split(';')[0]
+  const doc = kb.sym(base)
+
   try {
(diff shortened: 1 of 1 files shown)


megoth:
It would be good to have tests verifying these changes. It seems that no tests broke with this commit, so I suspect that no tests covered this originally.

If you don't have time to do it, perhaps you could describe the steps you did to verify that this works for you, and I can try to code that into a test.

ajmacdonald:
This PR solved an empty results problem I was getting when parsing JSON-LD and serializing to RDF/XML.

megoth:
Ok, we'll merge this into the next version of rdflib sounds like it's very valuable ^_^

timbl:
yes, no reason not to merge as far as I can see

Issue #332, opened 2019-08-14
Add WAC-Allow header awareness in updater's editable() function


timbl:
The solid spec requires servers to send headers like

`WAC-Allow: user="read",public="read"`

which give the user an insight into the operations which are allowed to them, and to the public, by the Access Control system.   Where this header is present, the RDFLIB client should not assume it can modify the document if it is not allowed to.

This will cause UIs to be simpler by removing modification tools, and prevent users from being offered editing when they are not allowed.

 So, if the resource is currently editable (there is a defined editing protocol), and when the AC-header  *is* present, and when it does *not* include "write" in the user permissions, then the editable function should return something false-like

timbl:
Used by various features already, but also should be in future  in all forms https://github.com/solid/solid-ui/issues/88

jeswr:
Closing: implemented by 5365067 (2019). On main, `UpdateManager.editable()` parses the recorded `WAC-Allow` header at src/update-manager.ts:178-186 — when the header is present and the user's permission group contains neither `write` nor `append`, `editable()` returns false, exactly as requested here. (It is also consulted for non-HTTP resources at :212-213.)

PR #333, opened 2019-08-14
Check wac-allow header in case it disallows write

timbl:
Addressing

https://github.com/linkeddata/rdflib.js/issues/332

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "rdflib",
   "description": "an RDF library for node.js. Suitable for client and server side.",
-  "version": "0.21.1-1",
+  "version": "0.21.1",
   "engines": {
     "node": ">=6.0"
   },
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "rdflib",
   "description": "an RDF library for node.js. Suitable for client and server side.",
-  "version": "0.21.1",
+  "version": "0.21.1-1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "rdflib",
   "description": "an RDF library for node.js. Suitable for client and server side.",
-  "version": "0.21.1",
+  "version": "0.21.1-1",
```
megoth: Unless you've published a prerelease to NPM, I don't think we should merge this.

src/update-manager.js
```diff
@@ -100,10 +100,11 @@ class UpdateManager {
     var definitive = false
     var requests = kb.each(undefined, this.ns.link('requestedURI'), docpart(uri))
 
-    // Hack for the moment @@@@ 2016-02-12
+    /* "Hack for the moment" @@@@ 2016-02-12 - Dimitri added (why?); timbl removed 2019-08
```
megoth: @dmitrizagidulin could we perhaps remove this hack altogether?

megoth: Always nice to remove code if we can ^_^

dmitrizagidulin: I'm sorry, I don't remember what the context for that was :) Feel free to remove.

megoth: Thank you, I'm removing this in the latest commit ^_^

PR #335, opened 2019-08-15
Migrate source to ES6 import/export

RubenVerborgh:
This commit migrates the `src` folder to use `import`/`export` rather than `require`/`module.exports`. In addition to just using more modern semantics, it allows tree shaking by webpack and others.

While it uses `default` exports for _internal_ files, the external exports (`index.js`) do _not_ use `default`. Hence, external code using `require` keeps on working.

The only breaking change is that external code that was incorrectly using `import rdf from 'rdflib'` should use `import * as rdf from 'rdflib'`. Preferably, such code imports the exposed subcomponents of `rdflib`, such that tree shaking becomes possible.

### DIFF
--- a/src/updates-via.js
+++ b/src/updates-via.js
@@ -1,9 +1,9 @@
 /*
  * Updates-Via
  */
-const namedNode = require('./data-factory').namedNode
+import { namedNode } from './data-factory'
 
-class UpdatesSocket {
+export class UpdatesSocket {
   constructor (parent, via) {
     this.parent = parent
     this.via = via
@@ -103,7 +103,7 @@ class UpdatesSocket {
   }
 }
 
-class UpdatesVia {
+export class UpdatesVia {
   constructor (fetcher) {
     this.fetcher = fetcher
     this.graph = {}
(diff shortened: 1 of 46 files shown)


timbl:
Note that ES6 import/export if you use the 'default' version is NOT COMPATIBLE with nodejs/commonjs module systems.  Therefore if default export is used, this will be a breaking change to the API.    That would take a high level strategic discussion.   So I assume that you are not using export default.

RubenVerborgh:
Note that the usage of defaults only happens in internal files; our exports
in index.js so not use default. Unless libraries directly depend on
internal files (which I hope not), this should be fine.

src/updates-via.js
```diff
@@ -1,9 +1,9 @@
 /*
  * Updates-Via
  */
-const namedNode = require('./data-factory').namedNode
+import { namedNode } from './data-factory'
```
Vinnl: There's a bug here, reported here: https://gitter.im/linkeddata/rdflib.js?at=5d5c1e77950718247717e691

`namedNode` is not a named export of `data-factory` - instead, it offers a default export of an object with a property `namedNode`.

RubenVerborgh: True; the export is correct, the import is not. On it.

RubenVerborgh: Fix in e7c5680122def0a2b1298fb5c7158203dd24c707.

PR #346, opened 2019-08-27
Automatically publish prereleases

Vinnl:
This PR configures Travis to automatically publish a branch to npm. For example, you could run

    npm install rdflib@auto-prerelease

to try out new features in this branch, if there had been any.

`npm install rdflib` will still only install the stable version, as is currently the case. The pre-release versions will also not be shown when you run `npm outdated`.

Some advantages are:

- we're able to test an exact production-like (built, minified, packaged) artefact just as if it were a stable release, _before_ making a stable release
- we would no longer have to manually cut a pre-release in advance
- we cannot forget to cut a pre-release in advance
- it also lays the foundation for running automatic integration tests later, if we're so inclined.

A disadvantage of the current setup is that, if you were to visit the rdflib npm page and then click through to [the versions tab](https://www.npmjs.com/package/rdflib?activeTab=versions), you'd also see the prerelease versions listed. Likewise if you run `npm info rdflib versions`, although you can still list just stable releases using `npm show rdflib@* version`.

/cc @RubenVerborgh, who had an opinion on this, to chime in.


Vinnl:
@RubenVerborgh Right, pre-release isn't the right term. I believe the more common term is [Review App](https://about.gitlab.com/product/review-apps/).

RubenVerborgh:
Thanks for looping me in.

So just for clarity, this PR does not just "automatically publish pre-releases". When fixed, it would "automatically _create_ a pre-release for every single commit on every single branch, and publish it to npm". And it is this creation that worries me, not the publication.

---

In general, I don't see an inherent need for publishing such a huge number of versions. It results in a large number of extra artifacts published on a public server that need to be managed. Some of these might be published with security bugs and might need deprecation or retraction.

It pollutes the list of npm version for a set of reasons that can also be achieved in different ways, and it is not sure if those reasons were requirements or needs in the first place.

Also, it seems to be wasteful to put an unreasonable burden on npm. The vast majority of such automatic pre-releases would not even be used once. For the cases we need it, we can just ask Travis to roll one, or roll one ourselves.

---

I will go through the list of advantages, but I'd rather see a list of requirements. Because requirements allow us to agree on what problems we are actually experiencing, and then explore multiple solutions with their strengths and weaknesses.

> * we're able to test an exact production-like (built, minified, packaged) artefact just as if it were a stable release, _before_ making a stable release

Such testing does not require a pre-release on every commit.

> * we would no longer have to manually cut a pre-release in advance

Yes, but that does not mean we need releases on every commit.

> * we cannot forget to cut a pre-release in advance

Other protection mechanisms can also do that, without all the releases.

> * it also lays the foundation for running automatic integration tests later, if we're so inclined.

Not sure why per-commit pre-releases help there.

---

> Right, pre-release isn't the right term

Then definitely a _no_ from me. Automated _release_ creation on a per-commit basis defeats the purpose of a release in the first place.

Is there any precedent for this on npm?

RubenVerborgh:
> Right, pre-release isn't the right term.

So the generation of versions like `1.2.3build456` is intended? Probably not even allowed, given that `3build456` is not an integer.

Vinnl:
> it would "automatically create a pre-release for every single commit on every single branch, and publish it to npm". 

Well, only the latest commit on every push - but yes, there'd be quite a few of them. We do the same for mashlib, solid-panes and solid-ui.

> Some of these might be published with security bugs and might need deprecation or retraction.

Given that they're clearly not stable releases, and people don't automatically get upgraded to them, I don't think people can reasonably expect any security/maintenance guarantees on these. We don't scrub commits with security bugs from our Git history either :)

> Also, it seems to be wasteful to put an unreasonable burden on npm.

I'd wager to guess that npm doesn't really care - especially since these can be stored relatively efficiently, given that the delta's are unlikely to be that large.

> I'd rather see a list of requirements. 

I'm not exactly sure when you consider something a requirements, but reformulated as a list of problem statements:

- Bugs can be introduced in the process of publishing a release to npm.
- Having to manually create releases increases the odds of making a mistake.
- It's easy to forget publishing a beta release.
- Automated integration tests of published npm packages can only be automated if the npm packages it tests are also published automatically.

> Automated release creation on a per-commit basis defeats the purpose of a release in the first place.

They're not actual releases, they're review apps. The purpose is to solve the problems above.

It seems your primary response to the problem statements above is that there are better solutions. Which might well be the case, but this is one that was easy to whip up in practically no time, and is available now.

So the choice is between:

- Merging this PR, with potential minor modifications.
- Leave the situation as is.
- Find time somewhere to implement a different solution.

I don't feel strongly either way, but I can't solve nr 3 :)

RubenVerborgh:
> We do the same for mashlib, solid-panes and solid-ui.

I think we need to discuss these better in the future.

> Given that they're clearly not stable releases,

Point taken; I discovered through your mashlib example that the identifier `1.2.3build456` generated by your script is translated into `1.2.3-build456` and thus _is_ a pre-release.

> can be stored relatively efficiently, given that the delta's are unlikely to be that large.

They are zipped, so delta will be nearly as large as the file. But it's more the principle whether we should publish every single commit we make. I still haven't seen the case for that.

> I'm not exactly sure when you consider something a requirements, but reformulated as a list of problem statements:

Still nothing in there that requires a release on every commit. It's excessive.

> They're not actual releases, they're review apps.

They are (pre-)releases to npm, regardless of what GitLab or others call them.

> but this is one that was easy to whip up in practically no time

…but quite cumbersome to undo.

Let's just not release tons of versions that will never be used. That is not a good signal to the outside world in any way. It's a messy solution.

Consider this a request for removal from mashlib, solid-panes, and solid-ui as well.

Vinnl:
> I think we need to discuss these better in the future.

Suggestions? They were submitted through a PR before being approved and merged.

> But it's more the principle whether we should publish every single commit we make. I still haven't seen the case for that.

> Still nothing in there that requires a release on every commit. 

As said: it's the easiest solution, one we have now.

> They are (pre-)releases to npm, regardless of what GitLab or others call them.

Well great, then I presume the point that they are actual releases is no longer relevant?

> …but quite cumbersome to undo.

That's fair; something to weigh when deciding on which of the three options I mentioned to go for. I think the third one is not realistic, so the choice is between solving the problems I mentioned or solving the problem of potentially having to undo this.

> That is not a good signal to the outside world in any way.

I don't think this affects the outside world, really: you'd have to go looking for the non-stable releases explicitly. The stable releases are as clean as they were.

> Consider this a request for removal from mashlib, solid-panes, and solid-ui as well.

Duly noted.

RubenVerborgh:
> Suggestions? They were submitted through a PR before being approved and merged.

We probably want more reviewers for alterations to publication processes.

Vinnl:
Since no maintainer seems to want automatic prereleases at this time using this code, I'm going to close this. The code is there for the taking though, if anyone changes their mind :)

Issue #355, opened 2019-10-01
Migrate to typescript


joepio:
Migrating to typescript will make the API more dev friendly, especially in TS projects. It will also help to prevent bugs and make refactoring easier. Related to #179. 

Done:

- [x] Configure tsconfig (start not too strict) / webpack / package.json
- [x] Use Babel, not typescript loader (similar config to [solid-panes](https://github.com/solid/solid-panes/pull/75/files?file-filters%5B%5D=.js&file-filters%5B%5D=dotfile#diff-e5e546dd2eb0351f813d63d1b39dbc48R11) and [solid-ui](https://github.com/solid/solid-ui/pull/127/files?file-filters%5B%5D=.json))
- [x] Switch from JSDoc to TSDoc / Typedoc

When converting `.js` files to `.ts`, be sure to:

- Remove 'use strict' from converted files
- Conform to the https://rdf.js.org/data-model-spec/
- Use [DefinitelyTyped types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/rdflib/index.d.ts) or [types from the existing ontola fork](https://github.com/fletcher91/rdflib.js/blob/ontola-fork/index.d.ts) 
- Improve docs
- Work towards stricter tsconfig

PR #356, opened 2019-10-02
Basic setup TypeScript #355

joepio:
Adds support for .ts files. PRs for converted files will follow soon.
More info in issue #355.


webpack.config.js
```diff
@@ -15,10 +15,18 @@ module.exports = (env, args) => {
     },
     module: {
       rules: [
+        {
+          test: /\.tsx?$/,
+          loader: "awesome-typescript-loader"
```
Vinnl: For a gradual transition, it might be a better idea to continue to use Babel for _both_ JS and TypeScript files, to prevent inconsistent output or bugs caused by differing implementations.

To do this, you could enable the TypeScript plugin for Babel, and run typechecking separately using the ForkTSChecker Webpack plugin. This should also make compilation faster, and absolve the need to run tests with ts-node :)

For inspiration, see [the PR on solid-panes](https://github.com/solid/solid-panes/pull/75/files?file-filters%5B%5D=.js&file-filters%5B%5D=dotfile) and [the one on solid-ui](https://github.com/solid/solid-ui/pull/127/files?file-filters%5B%5D=.json).

joepio: Good points, will do!

joepio: Done! Can you review again?

package.json
```diff
@@ -87,18 +94,18 @@
...
   "files": [
     ".babelrc",
     "src",
     "lib",
     "dist"
   ],
-  "main": "lib/index.js",
-  "module": "src/index.js",
+  "main": "lib/index",
+  "module": "src/index",
```
Vinnl: I don't think the extensions should be removed here? One thing you might want to add, though, is [the `types` property](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package), after [configuring TypeScript to also emit a `--declaration`](https://www.typescriptlang.org/docs/handbook/compiler-options.html).

webpack.config.js
Vinnl: Done, but not that I'm not an rdflib maintainer, so my review is more to provide feedback rather than a path towards merging...

PR #357, opened 2019-10-02
Read the Node's value instead of the Node itself

Vinnl:
I'm not really quite sure what this code is supposed to do, but the way it's set up right now, it doesn't look like it would ever work. `status` appears to be an array of `Node`s, so it will never be equal to `200` or `400` - we'd have to read the value for that.

Of course, if it never worked, the question is whether it _should_ work. It appears to mark a Document as not editable if its status code was 200 or 404 - the former is the case for a Document that I _think_ should be editable?


src/update-manager.js
```diff
@@ -138,7 +138,7 @@ export default class UpdateManager {
           var status = kb.each(response, this.ns.http('status'))
           if (status.length) {
             for (let i = 0; i < status.length; i++) {
```
timbl: Values of rdf nodes are strings not integers

Vinnl: Ah of course - good spot. Very happy with #355 ;)

Vinnl:
Request from Tim: add tests.

Issue #362, opened 2019-10-08
Node.fromValue() type assumptions - null / undefined


joepio:
The `Node.fromValue` method (which is called in the Statement constructor) creates an RDF Node (Term) from a JS value. If a new Statement is constructed with Null / undefined values, we'll have a statement with Subject / Predicate / Object values of null or undefined.

This behavior seems intentional, since it's documented: 

```
 RDF Nodes are returned unchanged, undefined returned as itself.
```

This means that the types for subject, predicate and object should be:

```js
  subject: Node | undefined | null;
  predicate: Node | undefined | null;
  object: Node | Collection | Literal | undefined | null;
```

Instead of: 

```js
  subject: Node;
  predicate: Node;
  object: Node | Collection | Literal;
```

These simpler types are easier to deal with, since it makes null / undefined checks unnecessary.
Since a triple without a subject and / or predicate / object seems silly, I think it might be better to throw an error and keep the types strict.

I've used the stricter types in my fork for #355 in the Statement constructor.

joepio:
Just discussed this with @megoth and @timbl. 

We agreed that `new Statement(null, undefined, null)` and other nonsensical statements should throw an error. This also means that `NamedNode.fromValue()` will throw if it is passed `undefined` or `null`, while currently it returns that value.

There was some discussion on how to deal with an `undefined` graph, though. Currently, it sets a `defaultGraph()`. We could throw an error if no explicit graph has been set, but this will could break many clients. Instead, we decided to log a warning to the console that this feature is at risk, linking to this issue.

PR #363, opened 2019-10-08
More Typescript migrations - switch to Typedoc

joepio:
Related issue: #355

- Converted some of the most fundamental classes to typescript, including Node, Literal, BlankNode, NamedNode, Collection, Statement.
- Introduced a `.types` file for shared types.
- Included a _temporary_ `types-temp.ts` file in project root as a reference file for documentation and keeping track of the ts migration process. 
- The `.isVar` method is set to boolean values, instead of 0 or 1. This seemed reasonable, as it's only used for boolean type checks, and the existing types already define it as a `boolean` value.
- JSDoc is switched with Typedoc. Combined with typescript, it makes the documentation far more complete.
- Sometimes I had to make assumptions on date types, e.g. that a user will never create a Statement containing a Collection. These assumptions used to be implicit, but typescript forces us to make them explicit. This happens when you see [type assertions](https://basarat.gitbooks.io/typescript/docs/types/type-assertion.html) `(value as type)`.
- Literals can apparently be `null` or `undefined`, when nodes are created using the `.fromValue` method. This happens in the Statement constructor. See #362.


src/node.ts
```diff
@@ -0,0 +1,65 @@
...
+ * @method fromValue
+ * @param value - Any native Javascript value
+ */
+Node.fromValue = function (value: ValueType): Node | Literal | undefined | null | Collection {
+  if (typeof value === 'undefined' || value === null) {
+    return value
+  }
+  const isNode = Object.prototype.hasOwnProperty.call(value, 'termType')
+  if (isNode) {  // a Node subclass or a Collection
+    // @ts-ignore
```
Vinnl: I'd preferably avoid adding `@ts-ignore` if possible, but if not, please at least add a comment indicating why it is required here, e.g.

    // @ts-ignore: Cannot cast `value` to the the required type because...

I did not add a comment to every `@ts-ignore`, as I'm sure you can do a Find in all Files ;)

src/node.ts
```diff
@@ -0,0 +1,65 @@
...
+    return value
+  }
+  if (Array.isArray(value)) {
+    return new Collection(value)
+  }
+  return Literal.fromValue(value)
+}
+
+import Namespace from './namespace'
```
Vinnl: I think usually it's good practice to put all `import` statements at the top of the file, so that it's clear when there are e.g. duplicate entries?

src/node.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+  // Node remains Node
+  // @ts-ignore
+  if (!term.datatype) return term // Objects remain objects
+  const literalTerm = term as Literal
+  // if (!Object.prototype.hasOwnProperty.call(term, 'dataType')) return term // Objects remain objects
+  if (literalTerm.datatype.sameTerm(ns.xsd('boolean'))) {
+    return literalTerm.value === '1'
+  }
+  console.log("4", term)
```
Vinnl: Forgot to remove this?
```suggestion
```

src/node.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+  if (
+    literalTerm.datatype.sameTerm(ns.xsd('integer')) ||
+    literalTerm.datatype.sameTerm(ns.xsd('float')) ||
+    literalTerm.datatype.sameTerm(ns.xsd('decimal'))
+  ) {
+    let z = Number(literalTerm.value)
+    return Number(literalTerm.value)
+  }
+  console.log("7", term)
```
Vinnl: Forgot to remove this?
```suggestion
```

src/store.js
```diff
@@ -20,7 +20,7 @@ import { ArrayIndexOf } from './util'
 import Formula from './formula'
 import { RDFArrayRemove } from './util'
 import Statement from './statement'
-import Node from './node'
```
Vinnl: Hmm, this might just be my unfamiliarity with the codebase, but is this change intended/valid? Likewise for the same change in `variable.js`, `formula.js`, etc.

src/named-node.ts
```diff
@@ -1,54 +1,58 @@
...
+   * @param iri The IRI for this node
    */
-  constructor (iri) {
+  constructor (iri: NamedNode | string) {
     super()
     this.termType = NamedNode.termType
 
-    if (iri && iri.termType === NamedNode.termType) {  // param is a named node
-      iri = iri.value
+    if (iri && (iri as NamedNode).termType === NamedNode.termType) {  // param is a named node
```
Vinnl: Tip: create a [user-defined type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) (e.g. `isNamedNode()`), then there's no need to do the cast inside the `if` block. (Or even after it, I guess, since `string` is the complementary type.)

src/named-node.ts
```diff
@@ -1,54 +1,58 @@
...
      return new NamedNode(str.slice(0, p + 1))
    }
-   /**
-    * Returns an NN for the whole web site, ending in slash.
-    * Contrast with the "origin" which does NOT have a trailing slash
-    */
-  site () {
+
+  /**
+   * Returns an NamedNOde for the whole web site, ending in slash.
```
Vinnl: Heh, +1 for expanding the acronym :)

src/literal.ts
```diff
@@ -1,13 +1,36 @@
...
+   * The language for the literal
+   */
+  lang!: string
+
+  /**
+   * The literal's datatype as a named node
+   */
+  datatype!: NamedNode
+
+  termType: string
```
Vinnl: Given that this can only be one string, this might be better?
```suggestion
  termType: 'Literal'
```

src/literal.ts
```diff
@@ -108,25 +137,26 @@ export default class Literal extends Node {
...
+
   /**
-   * @method fromValue
-   * @param value
-   * @return {Literal}
+   * Builds a literal node from an input value
+   * @param value The input value
    */
-  static fromValue (value) {
+  static fromValue(value: ValueType): Literal | Node | undefined | null {
```
Vinnl: The if statements in here could also be well-served by some user-defined type guards, I think.

src/literal.ts
```diff
@@ -1,13 +1,36 @@
...
+  /**
+   * Initializes this literal
+   * @param value The literal's lexical value
+   * @param language The language for the literal
+   * @param datatype The literal's datatype as a named node
+   */
+  constructor (value: string, language?: string | null, datatype?: NamedNode) {
     super()
-    this.termType = Literal.termType
+    this.termType = 'Literal'
```
Vinnl: Haha, how on earth did this work before? :P

src/store.js
joepio: Note sure about the implications, I'll revert them

joepio:
Thanks for all the feedback @Vinnl, I've made the changes you requested (except for the type guard in literal.ts, that changes the behavior and was a little tricky).

joepio:
@Vinnl I've added these type guards in [a new branch](https://github.com/ontola/rdflib.js/tree/typescript-3). Shall I update this PR to include the changes from that branch, or shall I submit it as a separate one?

Vinnl:
@joepio Generally, separate PRs for separate functionality is easier to review for me, so I'd prefer a new one, marked as dependent on this one. I see [more changes than just type guards](https://github.com/ontola/rdflib.js/compare/typescript-2...typescript-3) though :)

joepio:
> @joepio I see [more changes than just type guards](https://github.com/ontola/rdflib.js/compare/typescript-2...typescript-3) though :)

Very true - the biggest one is RDFJS Taskforce compatibility.

joepio:
Yesterday discussed to first merge #358. I'll rebase my typescript fork (which includes the changes from this PR + a lot more) on master after that, and then will issue a new PR.

Vinnl:
@joepio #358 has now been merged, so feel free to rebase.

joepio:
Replaced by #373

Issue #370, opened 2019-10-25
Change defaultGraph iri


joepio:
```
// in data-factory-internal.ts
export const defaultGraphURI = 'chrome:theSession'
```

In Firefox:
```
new URL("chrome:thesession")
// TypeError: chrome:thesession is not a valid URL.
new URL("anyOtherRandomScheme:anyvalue")
// URL {..valid}
```

This probably has to do with security considerations in Firefox, since they [mention](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/create#Parameters) it in the `tabs.create` docs.

Creating URL objects from `NamedNode` (including `DefaultGraph`) values is quite common.

Should we change the `defaultGraphURI` to an empty string? This seems most appropriate, since this is what the [RDFJS Taskforce spec](https://rdf.js.org/data-model-spec/#defaultgraph-interface) uses.

dmitrizagidulin:
+1, I too think we should change the `defaultGraphURI` to an empty string.

fletcher91:
`chrome:theSession` is a named graph (so `NamedNode`). It is used as the default _named graph_ (`NamedNode`), not to be confused with the RDF _default graph_ (`DefaultGraph`).

> Should we change the defaultGraphURI to an empty string

An empty string isn't a valid IRI, so that would cause a lot of problems. E.g. the trivial implementation of `Creating URL objects from NamedNode (including DefaultGraph) values is quite common.` would throw a TypeError.

We have used `rdf:defaultGraph` before, which is a valid url and parses in firefox as well. Though no registered scheme `rdf` exists. Though I think it has logical behaviour in practice; 
* Keeps the API and downstream code simple (e.g. code only has to expect named or blank nodes) 
* Calling `putBack` will still fail as intended (since it isn't a document but a concept). 
* Taskforce adhering projects should still work (`factory.defaultGraph().equals(quad.graph) === true`).

joepio:
How about `rdfjs:defaultGraph` instead of `rdf`? Seems to make sense, since defaultGraph is not in the rdf spec, and it refers to something created in RDFjs.

joepio:
Just discussed this with @megoth and @timbl. 

[...]

PR #373, opened 2019-12-03
Typescript

joepio:
This PR converts most of RDFlib.js to Typescript (#355).  It fixes various minor bugs and it adds a lot of documentation, using TypeDoc instead of JSDoc. Since it's such a big PR with most of the code changed, I've added a changelog with some of my thoughts. For the most important ones, I've already opened some issues, but for the less important ones I've just added some bullets below.

## Changes included in PR #363 (already approved)

- Converted some of the most fundamental classes to typescript, including `Node`, `Literal`, `BlankNode`, `NamedNode`, `Collection`, `Statement`.
- Introduced a `.types` file for shared types.
- JSDoc is replaced with Typedoc. Combined with types and comments from `@types/rdflib`, this makes the documentation far more complete.
- I used many of the types and comments from `@types/rdflib` by [Cénotélie](https://github.com/cenotelie/). Added credits in `package.json`, discussed this with Cénotélie.

## New changes

- Migrated `formula`, `variable`, `store`, `update-manager`, `data-factory`, `default-graph`, `namespace`, `parse`,`serialize`, `parse`, `uri` and `utils` tot ts.
- Added, and later removed `fork-ts-checker-webpack-plugin`, which enables errors in the log when ts errors occur. Since this gave errors, we use `tsc` in the `prepare` script. This means that only builds without type errors will pass.
- Added typeguards, e.g. `isTFNamedNode` and `isTFPredicate` in `Utils`, and used these at various locations.
- Use enums for `termType` and `contentType`, without breaking compatibility with regular strings.
- `Formula` Constructor arguments are optional - since some functions initialize empty Formulas.
- In `Formula.fromNT()` `return this.literal(str, lang || dt)` seemed wrong, converted it to
- The various `fromValue` methods conflict with the base Node class, type wise. Since they don't use `this`, I feel like they should be converted to functions.

## Compatibility with RDFJS taskforce and external datafactories

We've worked quite a bit on compliance with the TF (TaskForce) spec. The internal types (including the factory) extend Taskforce types, which live in a separate `tf-types.ts` file. This makes it easier to use RDFLib in combination with external, custom data factories.

- Switched internal calls from `sameTerm` to `equals` in order to comply to TF spec, so that these functions also work with external datafactories. Alias still exists, so nothing changes externally.
- Switched internal calls from `.why` to `.graph`. The `.why` alias still exists, so nothing changes for external users.
- Calls to `kb.sym` have been replaced with `kb.rdfFactory.namedNode`, which makes all these functions more compatible with external datafactories. Added a deprecation warning to `.sym`.

## Minor fixes

- Removed the last conditional of `Formula.holds()`, since it did not make sense
- Removed some unreachable code, unused variables and functions that didn't do anything such as `Node.substitute()`.
- Removed the `justOne` argument from `formula.statementsMatching`, since it was unused.
- The `uri.document` function called `.uri` on a string, I removed that.
- Transformed inline comments to JSDoc, moved them to type declarations instead of constructor.
- Some types are set to any, because I didn't fully understand them. I've added TODO comments for these.
- Removed the fourth argument from the `parser.parse` function in `fetcher.parse`, since the function only takes three.
- Removed the `response` argument from `fetcher.parse`, `XHTMLHandler.parse`, `RDFXMLHander.parse`, `XMLHandler`, since it was not used.
- `Fetcher.failfetch` added strings as objects to the store. Changed that to literals.
- Internal calls to `NamedNode.uri` are changed to `.value` to comply with TF spec. This enables these functions to work with external datafactories.
- Removed unused second argument from `Fetcher.cleanupFetchRequest`
- Created one huge `Options` type for Fetcher. Not sure if this is the way to go.
- In `Node.toJS`, the boolean only returned true if the `xsd:boolean` value is `'1'`, now it it should also work for `'true'`.
- Converted `kb.add(someStrin2g)` to `kb.add(new Namednode(somestring))` to enhance compatibility with other datafactories. This happens in `Fetcher` and
- `Fetcher.refreshIfExpired` passed an array of headers, but it needs only one string.
- `Fethcer` uses `Headers` a lot. I've changed empty objects to empty `new Headers` instances, which enhances compatibility with default `Fetch` behavior.
- `Serializer.tripleCallback` had an unused third argument.
- `UpdateManager.update` checked an undefined `secondTry` variable. Since later in the same function, `.update` is called with a 4th argument, I assume this is `secondTry`. I've added it as an optional argument. Perhaps this is
- `Formula.add()` now uses `this.rdfFactory.defaultGraph` instead of the non-existent `this.defaultGraph`
- `IndexedFormula.replaceWith` now passes a Node instead of a string to `.add` in the `if (big.value)` block
- Added support for `true` values in `xsd:boolean` literals, according to [xsd spec](https://www.w3.org/TR/xmlschema11-2/#boolean).

## Possible bugs discovered, which are not fixed by this PR

- `Formula.substitute` uses `this.add(Statments[])`, which will crash. I think it should be removed, since `IndexedFormula.substitute` is used all the time anyway.
- The `Formula.serialize` function calls `serialize.ts` with only one argument, so without a store. I think this will crash every time. Also`Formula.serialize` uses `this.namespaces`, but this is only defined in `IndexedFormula`. Is it rotten code and should it be removed?
- `IndexedFormula.add()` accepts many types of inputs, but this will lead to invalid statements (e.g. a Literal as a Subject). I suggest we make this more strict and throw more errors on wrong inputs. Relates to #362. We could still make the allowed inputs bigger by allowing other types with some explicit behavior, e.g. in subject arguments, create `NamedNodes` from `URL` objects and `strings` that look like URLs . In any case, I thinkg the `Node.fromValue` behavior is too unpredictable for `store.add`. For now, I've updated the docs to match its behavior.
- The types for `Node.fromValue` and `Literal.fromValue` show how unpredictable these methods are. I suggest we make them more strict (also relates to #362), so they either return a `TFTerm` (`node`) or throw an error - they should not return `undefined` or `null`. Also, I think they should be converted to functions in `Utils`: this would fix the circular dependency issue (why we need `node_internal`) and it would fix the type issues in `Literal.fromValue` (which tends to give issues since it's signature does not correctly extend from `Node.fromValue`)
- In `Fetcher.addtype`, the final logic will allways return `true`, since `redirection` is a `NamedNode`. Should it call `.value`?
- Various `Hanlder.parse()` functions in `Fetcher` return either a `Response` or a `Promise<Error>`. This seems like weird behavior - should it not always return an array?
- The `defaultGraph` iri is set to `chrome:theSession`, but this errors in Firefox. I suggest we change it to something else. See #370.
- The `Parse.executeErrorCallback` conditional logic is always `true`.
- I've added many `// @ts-ignore` comments. Ideally, these should be resolved by fixing the underlying type issues.
- `UpdateManager.update_statement` seems to refer to the wrong `this`. It calls `this.anonimize`, but it is not available in that scope.
- `UpdateManager.updateLocalFile` uses `Component`, but this is not defined anywhere. Is this deprecated?
- `Data-factory-internal.id()` returns `string | undefined`, I feel like undefined should not be possible - it should throw an error. This would resolve the type incompatibility on line 146.
- `IndexedFormula.copy` runs `.copy` on a Collection, but that method is not available there.
- `IndexedFormula.predicateCallback` is checked, but never used in this codebase.

## Other things I noticed

- Literals can apparently be `null` or `undefined`, when nodes are created using the `.fromValue` method. This causes faulty behavior. This happens in the `new Statement()` constructor as well. See #362.
- The `IndexedFormula.add()` method has logic for Statement array inputs and store inputs, but this behavior is not documented. It also refers to `this.fetcher` and `this.defaultGraph`, which both should not be available. I've added types that accept these arrays.
- The filenames of major classes differ from their default exports, e.g. `store.ts` is called `IndexedFormula`.
- Aliases (e.g. `IndexedFormula.match` for `IndexefFormula.statementsMatching`) introduce complexity, documentation and type duplication. I suggest adding deprecation warnings.
- The various calling methods of `Fetcher.nowOrWhenFetched` are quite dynamic. A simpler, stricter input type might be preferable.
- The Variable type (or `TFVariable`) really messes with some assumptions. I feel like they should not be valid in regular quads, since it's impossible to serialize them decently. If I'd add it to the accepted types, we'd require a lot of typeguards in functions.
- `Fetcher` `StatusValues` can be many types in RDFlib: string, number, true, undefined... This breaks compatibility with extending `Response` types, since these only use numbers. I feel we should only use the `499` browser error and add the text message to the `requestbody`. I've created a type for the internal `InternalResponse`; it shows how it differs from a regular `Response`. The `.responseText` property, for example, is used frequently in Fetcher, but
- The `IndexedFormula` and `Formula` methods have incompatible types, such as in `compareTerm`, `variable` and `add`. I've added `//@ts-ignore` lines with comments.
- The fourth `reponse` argument in `.parse()` methods in `Handler` classes was unused (except in N3Handler), so I removed it everywhere it was unused.
- `Serializer`'s fourth `options` argument is undocumented, and I couldn't find out how it worked.
- `Fetcher` `saveResponseMetadata` creates literals
- Many functions in `Fetcher` assume that specific `Opts` are defined. I've included all these in a single `Options` type and added documentation for the props I understood. I've also created an `AutoInitOptions` type, which sets auto-initialized. I extended Options in each function where specific opts seemed to be required. I'm not entirely confident about the types I've set. I feel like the truly required items should never be `Opts`, since they aren't optional. Refactoring this requires a seperate PR.
- `Fetcher.load` allows arrays as inputs. This causes the output types to be more unpredictable. `Promise<Result> | Result[]`. I suggest splitting this in two functions, e.g. add `loadMany`
- `Utils.callbackify` seems to be used only in `Fetcher`.
- `UpdateManager.editable` has a confusing return type (`string | boolean | undefined`). I suggest we refactor it to always return one of multiple pre-defined strings,.
- The `optional` argument in `formula.js` does not seem to be documented, used or tested - should it be removed?

## Need review

- Some of the `Formula` and `IndexedFormula` functions (e.g. `anyStatementMatching`) might have too strict types - perhaps Collections are allowed in some of them.
- `IndexedFormula.declareExistential` & `newExistential` have different assumptions on types - should they be blanknodes or namednodes?
- `IndexedFormula.check` passes a single statement to `checkStatementList`, which expects an array
- I've added many type assertions (e.g. `as TFObject`), but Ideally, these do not exist. Ultimately, these should be replaced by TypeGuards that work on runtime

## Some thoughts on simplifying language

Getting started with Linked Data or RDF can be difficult, since it introduces quite a few new concepts.
Since this library is powerful and generic, it might be one of the first and most important RDF tools that a developer will use.
Therefore, we should try to use consistent langauge and keep synonyms to a minimum.

- The name `Node` and `Term` seem to refer to the same concept. Both are used in this repo. I think Term is slightly more suited, partially because it complies to the TF spec, but also because it seems more sementically correct. A `Literal`, for example, is not really a node in the mathematical sense, it's more of an edge, since it cannot connect to other nodes.
- `Statement`, `Triple` and `Quad` refer to the same concept, at least technically. Maybe we could pick one. I suggest `Statement`, because it covers both triples and quads.
- The concept `graph` is referred to as `why`, `doc` and `graph` in the code and API. I think this might be confusing - should we just call it `graph` everywhere?
- the `IndexedFormula` default export name is different from the `store` filename. It might be easier to just call it `store` everywhere, including where it's called `kb`.

## Notes for PR reviewer

- All my commits have been squashed into [one commit](https://github.com/linkeddata/rdflib.js/pull/373/commits/5e7c08c0c0dcc36420f27ec0c8bc9d9ed41fba77), since rebasing onto @fletcher91's changes proved very difficult. The individual commits can be seen [here](https://github.com/ontola/rdflib.js/commits/typescript-4).
- Github does not show line-by-line changes in the renamed files, which makes reviewing this PR in Github unnecessarily hard. I recommend using an IDE with decent typescript support and a Git blame plugin.


src/tf-types.ts
```diff
@@ -0,0 +1,173 @@
...
+  LiteralTermType,
+  NamedNodeTermType,
+  VariableTermType,
+} from './types'
+
+/**
+ * RDF/JS taskforce Term
+ * @link https://rdf.js.org/data-model-spec/#term-interface
+ */
+export interface TFTerm {
```
RubenVerborgh: Let's get rid of the entire `TF` prefix; it's not the right one, and it doesn't have a place here. rdflib.js terms are RDF/JS-compatible.

src/tf-types.ts
```diff
@@ -0,0 +1,173 @@
+import { SupportTable } from './factories/factory-types'
```
RubenVerborgh: RDF/JS already has types: https://github.com/rdfjs-base/data-model/blob/master/index.d.ts
Shall we just use those?

src/utils/terms.ts
```diff
@@ -0,0 +1,123 @@
...
+/** TypeGuard for RDF/JS TaskForce Terms */
+export function isTFTerm (obj: any): obj is TFTerm {
+  return typeof obj === 'object'
+    && obj !== null
+    && 'termType' in obj
+    && 'value' in obj
+}
+
+/** TypeGuard for RDF/JS TaskForce Literals */
+export function isTFLiteral (value: any): value is TFLiteral {
```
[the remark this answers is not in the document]
joepio: I had to have some distinction between RDFlib.js types and RDF/JS types. I agree that TF is not defining and could be confusing. Perhaps I should add `rdflib` namespace to RDFlib.js types?

tests/unit/fetcher-test.js
```diff
@@ -435,35 +436,35 @@ describe('Fetcher', () => {
...
+  //   it('should default to text/xml for chrome: protocol uris')
+  // })
+  //
+  // describe('handlerForContentType', () => {
+  //   it('should return null when no contentType given')
+  //
+  //   it('should return a handler instance if content type matches')
+  //
+  //   it('should return null when no handler match is found')
+  // })
```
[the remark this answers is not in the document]
joepio: These tests should work, I've reverted this in a fixup.

src/tf-types.ts
joepio: If we remove the TF prefix, how should we distinguish between RDFlib.js NamedNode (which has some extra methods) and RDF/JS NamedNodes? Add a prefix for RDFLib types?

src/tf-types.ts
joepio: I wasn't aware of that implementation. I don't know what the consequences are of replacing our internal types with the ones from this one. I think, since that library is just `js` with one (minimal) typings file, that we might lose quite a bit of type information.

src/tf-types.ts
Vinnl: A prefix for RDFLib types that extend RDF/JS types would make sense to me, since they're the special case in this regard.

tests/unit/util-test.js
```diff
@@ -25,7 +26,7 @@ describe('util', () => {
     })
 
     it ('handles literals', () => {
-      expect(isTerm(new Literal('test'))).to.be.true()
+      expect(isTFTerm(new Literal('test'))).to.be.true()
     });
```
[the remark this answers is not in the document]
joepio: NamedNode, Term and Statement only, currently. I'll add tests for the others.

src/utils/terms.ts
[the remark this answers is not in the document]
RubenVerborgh: Yes, as I wrote in the review, I think the general thing should indeed be the RDF/JS conventions, and then rdflib-specific types. They are subtypes of the RDF/JS types.

src/tf-types.ts
RubenVerborgh: Not the implementation; just the types.

src/tf-types.ts
joepio: RDFLib types, however, are actually Classes. Do you suggest I rename the classes? `NamedNode` => `RDFlibNamedNode`, or perhaps rename the imports?

```js
import {NamedNode as RDFlibNamedNode} from "./namednode"
```

Don't think this is what you'd want.

tests/unit/util-test.js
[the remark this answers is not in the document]
joepio: Added tests!

src/tf-types.ts
joepio: The types are just nine lines of code, they lack pretty much all the methods (which reside in the .js files), or am I missing something?

src/tf-types.ts
RubenVerborgh: We need to distinguish between the export names of the library (as a whole) and the internal naming. The external naming won't change in any case; the exported `NamedNode` will be the rdflib one. I think for internal purposes, `RDFlibNamedNode` makes sense. Or perhaps `RdfLib.NamedNode` or so.

src/tf-types.ts
RubenVerborgh: Looks like it's rather the imported `rdf-js` (https://www.npmjs.com/package/@types/rdf-js) is what we want.

joepio: Ah, these map pretty much 1 on 1. I'll try to replace them.

src/utils.ts
```diff
@@ -0,0 +1,114 @@
+import Fetcher from './fetcher'
```
Vinnl: Am I correct in understanding that `util` contains utility methods written in Javascript, and `utils` methods written in TypeScript? If so, I'm not sure that pluralising it is a good way of discerning between the two - maybe just rename `util.js` to something like `jsUtils.js`, for example?

tests/unit/factories/canonical-data-factory-test.ts
```diff
@@ -0,0 +1,64 @@
...
+    describe('#supports', () => {
+      it('collections', () => { expect(Factory.supports[Feature.collections]).to.be.false() })
+      it('defaultGraphType', () => { expect(Factory.supports[Feature.defaultGraphType]).to.be.false() })
+      it('equalsMethod', () => { expect(Factory.supports[Feature.equalsMethod]).to.be.true() })
+      it('identity', () => { expect(Factory.supports[Feature.identity]).to.be.false() })
+      it('id', () => { expect(Factory.supports[Feature.id]).to.be.true() })
+      it('reversibleId', () => { expect(Factory.supports[Feature.reversibleId]).to.be.false() })
+    })
+
+    describe('equals', () => {});
```
Vinnl: Is this a way of leaving a marker that it would be good if tests for these would be written as well? If so, perhaps add an explicit `// TODO`?

tests/unit/factories/canonical-data-factory-test.ts
```diff
@@ -0,0 +1,64 @@
...
+import Factory from '../../../src/factories/canonical-data-factory';
+import { Feature } from '../../../src/factories/factory-types'
+import NamedNode from '../../../src/named-node'
+import Literal from '../../../src/literal'
+import DefaultGraph from '../../../src/default-graph'
+import Empty from '../../../src/empty'
+
+describe('data-factory', () => {
+  describe('DataFactory', () => {
+    describe('#supports', () => {
```
Vinnl: The idea behind the test functions being called `it` is that you can make somewhat natural-sounding tests, so it'd probably be more descriptive to remove this `describe('#supports`, and instead use `it('supports collections'`, etc.

tests/unit/factories/canonical-data-factory-test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+import Factory from '../../../src/factories/canonical-data-factory';
+import { Feature } from '../../../src/factories/factory-types'
+import NamedNode from '../../../src/named-node'
+import Literal from '../../../src/literal'
+import DefaultGraph from '../../../src/default-graph'
+import Empty from '../../../src/empty'
+
+describe('data-factory', () => {
+  describe('DataFactory', () => {
```
Vinnl: Isn't this redundant?

tests/unit/factories/canonical-data-factory-test.ts
Vinnl: Especially since it appears that some should be `it('does not support defaultGraphType'.

tests/unit/factories/extended-term-factory-test.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Feature } from '../../../src/factories/factory-types'
+import Collection from '../../../src/collection'
+
+/**
+ * @ignore
+ * Inherits from CanonicalDataFactory (internal), so we only have to test the added features
+ * e.g. Collection
+ */
+describe('extended-term-factory', () => {
+  describe('#supports', () => {
```
Vinnl: `it('supports collections'`, etc.

src/factories/canonical-data-factory.ts
```diff
@@ -0,0 +1,210 @@
...
+      return this.id(a) === this.id(b)
+    }
+
+    return false
+  },
+
+  /**
+   * Generates a uniquely identifiably *idempotent* string for the given {term}.
+   *
+   * Equivalent to {Term.hashString}
```
Vinnl:
```suggestion
   * Equivalent to [[Term.hashString]]
```

src/update-manager.ts
```diff
@@ -571,27 +606,29 @@ export default class UpdateManager {
...
     var openWebsocket = function () {
       // From https://github.com/solid/solid-spec#live-updates
       var socket
       if (typeof WebSocket !== 'undefined') {
-        socket = new WebSocket(wssURI)
+        socket = new WebSocket(validWssURI)
+        //@ts-ignore Firefox Addon
       } else if (typeof Services !== 'undefined') { // Firefox add on http://stackoverflow.com/questions/24244886/is-websocket-supported-in-firefox-for-android-addons
+        //@ts-ignore Firefox Addon
         socket = (Services.wm.getMostRecentWindow('navigator:browser').WebSocket)(wssURI)
```
Vinnl: Given that Firefox Add-Ons now use WebExtensions, I presume that this entire clause can be deleted :)

src/store.ts
```diff
@@ -113,41 +202,68 @@ export default class IndexedFormula extends Formula { // IN future - allow pass
...
+    target: TFNamedNode,
+    patchCallback: (errorString: string) => void
+  ): void {
     var targetKB = this
     var ds
-    var binding = null
+    var binding: Bindings | null = null
 
-    function doPatch (onDonePatch) {
+    function doPatch (onDonePatch: Function) {
```
megoth: You could be a bit more specific on the Function, e.g. `() => void`

src/tf-types.ts
megoth: I +1 the use of https://www.npmjs.com/package/@types/rdf-js

tsconfig.json
```diff
@@ -3,13 +3,16 @@
...
+    "target": "es2019",
+    "module": "esnext",
     "strict": true,
     "noImplicitAny": false,
+    "noUnusedLocals": true,
     "lib": [
-      "es2019"
-    ]
+      "es2019",
+      "dom"
```
megoth: Is the introduction of `dom` ok considering the use of rdflib in Node?

src/namespace.ts
```diff
@@ -0,0 +1,15 @@
+import RDFlibNamedNode from './named-node'
+import { TFDataFactory, NamedNode } from './tf-types'
+
+/**
+ * Gets a namespace for the specified namespace's URI
+ * @param nsuri - The URI for the namespace
```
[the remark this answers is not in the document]
Vinnl: You still need `@param` to add a description for a specific parameter - it's just no longer necessary to include the type there :)

joepio:
I tried ([see commit](https://github.com/ontola/rdflib.js/commit/3972399e455e2aa160f15971b25b16e4f79b1dd0)) replacing my TF types with those from @types/rdf-js, but it seems these are incompatible - both with RDFlib and with the original spec.

One of the most fundamental problems is that @types/rdf-js defines `Term` as a Union of its other Terms:
```typescript
export type Term = NamedNode | BlankNode | Literal | Variable | DefaultGraph;
```

The [RDF/JS spec](http://rdf.js.org/data-model-spec/#term-interface) defines Term as an abstract interface.

Since RDFlib uses custom terms (like Collection), it is compatible with the RDF/JS spec, and it works with my types, but it does not work with the types from @types/rdf-js. This problem with Term, unfortunately, trickles down quite a bit and makes it impossible to switch, at the moment.

There are a couple of things we can do here:

1. Issue a PR for [DefinitelyTyped/rdf-js](https://github.com/ontola/DefinitelyTyped), change Term to fit the spec. This does require that these changes do not break anything in rdf-js.
2. Add the types to the [RDF/JS repo](https://github.com/rdfjs-base/data-model).
3. Do not import the @types/rdf-js types, but use my previous implementation.

@RubenVerborgh What are your thoughts on this?

megoth:
To me it sounds like we want to start a process to make sure that the interfaces expressed by RDFJS are compatible with the work in rdflib. That is a lengthier discussion though, so I would suggest removing those parts and keep this PR simple.

We could of course mitigate that for now by having `src/tf-types`, but I think removing them altogether is better as a first step?

joepio:
> To me it sounds like we want to start a process to make sure that the interfaces expressed by RDFJS are compatible with the work in rdflib. That is a lengthier discussion though, so I would suggest removing those parts and keep this PR simple.

I'll make a separate issue for migrating to external types.

> We could of course mitigate that for now by having `src/tf-types`, but I think removing them altogether is better as a first step?

Removing them has quite a big impact, I'd suggest we use the `tf-types` for now and do the migration to an external type definition later. As for now, the types in this PR seem to resemble the spec more closely.

src/utils/terms.ts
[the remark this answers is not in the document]
joepio: I suggest we continue this discussion in #374

src/tf-types.ts
joepio: Changed the names!

src/tf-types.ts
joepio: Turns out it's a bit of a problem to use these types. See #374

tsconfig.json
joepio: The Response type used in rdflib is the one that Fetch uses. I tried the Express response, but it was not compatible. 

Is there a way to only use the Response type, without requiring the `dom` lib?

src/store.ts
joepio: True, fixed.

src/node-internal.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  /** The class order for this node */
+  classOrder!: number;
+
+  /** The node's value */
+  value: string;
+
+  protected constructor(value: string) {
+    this.value = value
+  }
```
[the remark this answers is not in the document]
joepio: Various methods of `node-internal` call this.value, so I thought it had to be instantiated. 
 
Fixed it by using the bang! / non-null assertion operator.

src/node.ts
```diff
@@ -16,30 +14,23 @@ export default Node
...
+ */
+Node.toJS = function (term: Term): Term | boolean | number | Date | string | any[] {
+  if (isCollection(term)) {
     return term.elements.map(Node.toJS) // Array node (not standard RDFJS)
   }
-  if (!term.datatype) return term // Objects remain objects
+  if (!isTFLiteral(term)) return term
   if (term.datatype.equals(ns.xsd('boolean'))) {
-    return term.value === '1'
+    return term.value === '1' || term.value === 'true'
```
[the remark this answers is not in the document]
joepio: I added it because it's in the [`xsd` spec](http://books.xmlschemata.org/relaxng/ch19-77025.html). Should have added it to the PR comments, though!

.npmignore
```diff
@@ -1,6 +0,0 @@
-doc
```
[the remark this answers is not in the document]
joepio: That file was [added](https://github.com/linkeddata/rdflib.js/commits/master/.npmignore) by Vincent to not share incomplete types, I think all should work fine without the file.

src/formula.ts
```diff
@@ -0,0 +1,936 @@
...
+  /**
+   * The accompanying fetcher instance.
+   *
+   * Is set by the fetcher when initialized.
+   */
+  fetcher?: any
+
+  initBindings: ReadonlyArray<any>
+
+  isVar = 0
```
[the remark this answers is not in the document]
joepio: I converted all isVars to booleans too, but Thom pointed out that some tests were failing because of this.

src/fetcher.ts
```diff
@@ -422,12 +593,78 @@ function isXHTML (responseText) {
   return responseText.substr(docTypeStart, docTypeEnd - docTypeStart).indexOf('XHTML') !== -1
 }
 
-function isXML (responseText) {
-  return responseText.match(/\s*<\?xml\s+version\s*=[^<>]+\?>/)
+function isXML (responseText: string): boolean {
+  const match = responseText.match(/\s*<\?xml\s+version\s*=[^<>]+\?>/)
+  return match ? true : false
```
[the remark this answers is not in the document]
joepio: Agreed, dit that

src/fetcher.ts
```diff
@@ -657,29 +915,31 @@ export default class Fetcher {
...
     }
 
-    let docuri = uri.uri || uri
+    let docuri = termValue(uri as NamedNode)
     docuri = docuri.split('#')[0]
 
     options = this.initFetchOptions(docuri, options)
 
-    return this.pendingFetchPromise(docuri, options.baseURI, options)
+    return this.pendingFetchPromise(docuri, (options as AutoInitOptions).baseURI, (options as AutoInitOptions))
```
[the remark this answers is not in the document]
joepio: nice!

src/fetcher.ts
```diff
@@ -706,33 +966,30 @@ export default class Fetcher {
     })
   }
 
-  cleanupFetchRequest (originalUri, options, timeout) {
+  cleanupFetchRequest (originalUri: string, timeout: number) {
```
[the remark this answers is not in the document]
joepio: I'll change it, add a deprecation warning and document it.

tsconfig.json
megoth: Not sure =\ Have you been able to test the transpiled files with Node- (e.g. NSS) and web-based (e.g. mashlib) systems? If yes, I'll approve this PR, if no, I'll try to test it out myself later and report back.

src/utils.ts
joepio: Agreed. Renamed to utils-js.js (so they appear next to eachother alphabetically)

tests/unit/factories/canonical-data-factory-test.ts
joepio: Added tests

tests/unit/factories/canonical-data-factory-test.ts
joepio: Agreed, edited.

src/named-node.ts
```diff
@@ -1,55 +1,58 @@
...
- * @class NamedNode
- * @extends Node
+ * A named (IRI) RDF node
  */
-export default class NamedNode extends Node {
+export default class NamedNode extends Node implements TFNamedNode {
+  static termType = TermType.NamedNode
+
+  classOrder = ClassOrder.NamedNode
+  termType: NamedNodeTermType = TermType.NamedNode
```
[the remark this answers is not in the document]
joepio: If I leave the non-static one out, Typescript says:

```
'import("/Users/joep/dev/src/github.com/linkeddata/rdflib.js/src/named-node").default' incorrectly implements interface 'import("/Users/joep/dev/src/github.com/linkeddata/rdflib.js/src/tf-types").NamedNode'.
  Types of property 'termType' are incompatible.
    Type 'TermType' is not assignable to type '"NamedNode"'.
      Type '"BlankNode"' is not assignable to type '"NamedNode"'.ts(2420)
```

Can't understand why, though...

src/types.ts
```diff
@@ -0,0 +1,106 @@
...
+  DefaultGraph = 'DefaultGraph',
+  Literal = 'Literal',
+  NamedNode = 'NamedNode',
+  Variable = 'Variable',
+
+  // The next ones are not specified by the rdf.js taskforce
+  Collection = 'Collection',
+  Empty = 'Empty',
+  Graph = 'Graph',
+}
```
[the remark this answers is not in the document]
joepio: It works! Required some refactoring, though.

src/named-node.ts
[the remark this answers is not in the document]
Vinnl: That's because `TFNamedNode` defines a non-static member `termType`. Is the `static` version used somewhere? A search for `NamedNode.termType` doesn't turn up any results for me...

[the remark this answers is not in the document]
joepio: The class (static) method isn't used, you're right. Removed the static properties.

joepio:
@megoth 

I believe testing in other applications (e.g. in mashlib, some node env) is smart before doing a definitive release and bumping the version. U agree that a pre-release is the way to go!

src/types.ts
```diff
@@ -0,0 +1,98 @@
...
+  n3 = "text/n3",
+  n3Legacy = "application/n3",
+  nTriples = "application/n-triples",
+  nQuads = "application/n-quads",
+  nQuadsAlt = "application/nquads",
+  jsonld = "application/ld+json",
+  xhtml = "application/xhtml+xml",
+  html = "text/html",
+  sparqlupdate = "application/sparql-update",
+}
```
Vinnl: I think this might work as a union type as well?

src/node-internal.ts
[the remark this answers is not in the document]
Vinnl: Huh, weird, given that `Empty` and `DefaultGraph` didn't use to set it. Ah well :man_shrugging:

src/parse.ts
```diff
@@ -0,0 +1,137 @@
...
+               contentType === ContentType.nQuadsAlt) {
+      var n3Parser = new N3jsParser({ factory: DataFactory })
+      nquadCallback(null, str)
+    } else if (contentType === undefined) {
+      throw new Error("contentType is undefined")
+    } else {
+      throw new Error("Don't know how to parse " + contentType + ' yet')
+    }
+  } catch (e) {
+    executeErrorCallback(e)
```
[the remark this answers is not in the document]
joepio: That code is a bit weird, but it was already there. A comment above mentions that it has to do with JSON-LD parsing, maybe it can be removed? I'll ask Thom en Tim, but it will be a separate issue / PR.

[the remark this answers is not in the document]
Vinnl: Yeah, if it was already there a separate PR makes more sense - or just don't bother, if it's not causing any problems now.

RubenVerborgh:
>     * I used many of the types and comments from `@types/rdflib` by [Cénotélie](https://github.com/cenotelie/). Added credits in `package.json`, discussed this with Cénotélie.

Why do we copy instead of use? I only see disadvantages. There are more stakeholder to the RDF/JS domain than the maintainer of that one package. If there are incompatibilities, the package should be adjusted.

src/types.ts
joepio: Agreed, changed.

RubenVerborgh:
Also, we should use proper naming. The word "taskforce" shouldn't occur in code, comments, or anywhere else. The proper terminology is the RDF/JS Data Model specification: https://rdf.js.org/data-model-spec/

joepio:
@RubenVerborgh These are not copies, the types are slightly different. See #374. I've removed the 'taskforce' mentions.

RubenVerborgh:
@joepio Thanks, lost track of all the issues. Only remnants now are `TF` prefixes as in `TFDataFactory`, which should also be replaced. Perhaps `RdfJsDataFactory` or `RSDataFactory`.

Happy to withdraw my objections after the last commits, and if we are using #374 to ensure that the interfaces are in fact copies and thus straight imports. If we can't something is wrong on either our side or their side.

src/formula.ts
```diff
@@ -0,0 +1,936 @@
...
+
+  /** The additional constraints */
+  constraints: ReadonlyArray<any>;
+
+  /**
+   * The accompanying fetcher instance.
+   *
+   * Is set by the fetcher when initialized.
+   */
+  fetcher?: any
```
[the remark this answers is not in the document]
joepio: Nope! edited it.

PR #388, opened 2020-02-11
fix #364 Unable to serialize JSON-LD

### LINKED ISSUE #364: Unable to serialize to JSON-LD
Using 1.0.4 (and N3 1.2.1) I'm getting `TypeError: n3Writer.addTriple is not a function` when trying to serialize to JSON-LD. (The other formats work correctly.)

Seems that `addQuad` needs to be used. Can't see any tests about serializing to JSON-LD either.

ericprud:
[convertToJson's first waterfall function](https://github.com/linkeddata/rdflib.js/blob/master/src/convert.js#L13) is expected to get [called repeatedly](https://github.com/rdfjs/N3.js#from-an-rdf-document-to-quads). Apparently [async's waterfall](https://caolan.github.io/async/v3/docs.html#waterfall) expects a [single invocation](https://github.com/caolan/async/blob/master/lib/waterfall.js#L72). I moved the second callback into the first:
``` javascript
    n3Parser.parse(n3String, function (error, quad, prefix) {
      if (error) {
        throw error;
      } else if (quad !== null) {
        n3Writer.addQuad(quad);
      } else {
        n3Writer.end(callback);
      }
    });
```

https://github.com/linkeddata/rdflib.js/issues/364#issuecomment-584894816

### DIFF
--- a/src/convert.js
+++ b/src/convert.js
@@ -10,15 +10,15 @@ export function convertToJson (n3String, jsonCallback) {
   })
   asyncLib.waterfall([
     function (callback) {
-      n3Parser.parse(n3String, callback)
-    },
-    function (triple, prefix, callback) {
-      if (triple !== null) {
-        n3Writer.addTriple(triple)
-      }
-      if (typeof callback === 'function') {
-        n3Writer.end(callback)
-      }
+      n3Parser.parse(n3String, function (error, quad, prefixes) {
+        if (error) {
+          throw error;
+        } else if (quad !== null) {
+          n3Writer.addQuad(quad);
+        } else {
+          n3Writer.end(callback);
+        }
+      });
     },
     function (result, callback) {
       try {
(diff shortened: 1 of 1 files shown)


src/convert.js
```diff
@@ -10,15 +10,15 @@ export function convertToJson (n3String, jsonCallback) {
...
-    function (triple, prefix, callback) {
-      if (triple !== null) {
-        n3Writer.addTriple(triple)
-      }
-      if (typeof callback === 'function') {
-        n3Writer.end(callback)
-      }
+      n3Parser.parse(n3String, function (error, quad, prefixes) {
+        if (error) {
+          throw error;
```
RubenVerborgh: Actually we want to `callback(error)` here.

src/convert.js
```diff
@@ -47,15 +47,15 @@ export function convertToNQuads (n3String, nquadCallback) {
...
-    function (triple, prefix, callback) {
-      if (triple !== null) {
-        n3Writer.addTriple(triple)
-      }
-      if (typeof callback === 'function') {
-        n3Writer.end(callback)
-      }
+      n3Parser.parse(n3String, function (error, triple, prefixes) {
+        if (error) {
+          throw error;
```
RubenVerborgh: Idem.

src/convert.js
ericprud: yep, [async docs](https://caolan.github.io/async/v3/docs.html#waterfall) support that. `throw error` throws all the way out of the call to `waterfall` while `callback(error)` calls the last function with whatever was thrown.

nickevansuk:
+1

Could we get this merged, as it's currently preventing RDF -> JSON-LD conversion from working?

megoth:
@ericprud has committed a fix for the requested changes - @RubenVerborgh are you ok with the fix?

PR #390, opened 2020-02-18
Support saveResponseMetadata for non-http(s) responses and allow them to be edited in data-kitchen

jeff-zucker:
If we get headers, we should save them.  Solid-rest provides content-type, link, and other headers for file:// and app:// responses and we need them for data-kitchen.

### DIFF
--- a/src/fetcher.ts
+++ b/src/fetcher.ts
@@ -1666,7 +1666,7 @@ export default class Fetcher implements CallbackifyInterface {
     kb.add(responseNode, this.ns.http('statusText'),
     kb.rdfFactory.literal(response.statusText), responseNode)
 
-    if (!options.resource.value.startsWith('http')) {
+    if (!response.headers) {
       return responseNode
     }
 
--- a/src/update-manager.ts
+++ b/src/update-manager.ts
@@ -106,7 +106,7 @@ export default class UpdateManager {
     uri = termValue(uri)
 
     if ((uri as string).slice(0, 8) === 'file:///' || (uri as string).slice(0,6) === 'app://' ) {
-      if (( typeof window != "undefined" && window.hasOwnProperty('kitchen') ) || kb.holds(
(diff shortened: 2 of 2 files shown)


timbl:
I didn't mean we should do all that for thiis PR - for this PR i think just remove both conditions so all file: are deemed editable.

jeff-zucker:
@timbl - I've run into additional problems with the updater and forms.  I am able to fudge it enough to get forms-ui working on local filesystem but it isn't clean.  So I am inclined to hold off on the PRs and fix solid-rest to return wac-allow and other needed things first, then see where we need exceptions.  Other than wac-allow, what are the priorities for what the filesystem should send to rdflib for it to operate properly?

Currently, here is what solid-rest sends back for a HEAD request on a resource:
```javascript
 {
    url: 'file:///var/www/html/solid/solid-auth-cli/test-folder/rest/deep-folder/test-file2.ttl',
    status: 200,
    statusText: 'OK',
    headers: Headers { [Symbol(map)]: [Object: null prototype] },
    counter: undefined
  }
```
and the output of headers.entries on a HEAD request
```javascript

[ 'allow', 'HEAD, GET, POST, PUT, DELETE' ]
[ 'content-type', 'text/turtle' ]
[ 'date', '2020-03-19T02:53:22.064Z' ]
[
  'link',
  '<test-file2.ttl.meta>; rel="describedBy", <test-file2.ttl.acl>; rel="acl",<http://www.w3.org/ns/ldp#Resource>; rel="type"'
]
[
  'location',
  'file:///var/www/html/solid/solid-auth-cli/test-folder/rest/deep-folder/test-file2.ttl'
]
[ 'x-powered-by', 'solid-rest-file-storage-1.0.0' ]

```

jeff-zucker:
Solid-rest now supplies wac-allow headers so we can remove the old file-extension writing and fully support  file:/// URIs in both the fetcher and the updater.  This PR replaced by PR #444

PR #394, opened 2020-03-01
Remove fetcher logging

RubenVerborgh:
Some `console.log` statements are left in the fetcher, and they basically flood the screen if you're doing HTTP.


timbl:
Looks good to me.

Issue #387, opened 2020-02-09
Use of $rdf.st() as static function broken


timbl:
For example 
https://github.com/solid/solid-panes/blob/master/src/internal/internalPane.ts#L164
breaks
with  `this is not defined` at
https://github.com/solid/solid-panes/blob/master/src/internal/internalPane.ts#L164

PR #395, opened 2020-03-23
Export bound factory methods

RubenVerborgh:
Fixes https://github.com/linkeddata/rdflib.js/issues/387

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,5 @@
+import { IRDFlibDataFactory } from './types'
+
 import BlankNode from './blank-node'
 import Collection from './collection'
 import * as convert from './convert'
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -34,8 +36,13 @@ export * from './utils/terms'
 
 const NextId = BlankNode.nextId
 
-const { fromNT } = Formula.prototype;
 
+// Prepare bound versions of data factory methods for export
+const boundDataFactory = {} as IRDFlibDataFactory
+for (const name in DataFactory) {
+  if (typeof DataFactory[name] === 'function')
+    boundDataFactory[name] = (...args) => DataFactory[name](...args)
```
michielbdejong: I might be wrong, but I think this way of doing it unnecessarily gives those methods access to lexical scope, so I think
```js
boundDataFactory[name] = boundDataFactory[name].bind(boundDataFactory)
```
would be slightly cleaner, but it would definitely work this way too, so that's probably fine.

Thanks a lot for jumping in to fix this one!

RubenVerborgh: They're equivalent, but I seem to remember that anonymous functions a preferred over `bind` in V8 for speed reasons. And it's a `const` so can be propagated.

michielbdejong: Ah ignore that, actually, the scope wouldn't make it into the `DataFactory[name](...args)` call anyway.

RubenVerborgh: (My knowledge is, however, outdated: https://stackoverflow.com/questions/17638305/why-is-bind-slower-than-a-closure/17638540#17638540)

RubenVerborgh: Changed to `bind` give no performance difference; saves the `name` lookup as well.

PR #396, opened 2020-03-25
NPM scripts - Changing prepare to prepublishOnly + adding postpublish

megoth:
Based on the setup we have for mashlib - https://github.com/solid/mashlib/blob/master/package.json - which works well for what we need to automate when packaging for NPM


megoth:
> Moving to `prepublishOnly` is definitely good. Might want to also add `npm ci &&` in front, to avoid that local changes make it into the build (this has happened before).

That's reasonable, so I've added it in the latest commit.

> I don't think we want to tie a `git` workflow to `npm`; there's no reason for that.

I've found it very useful with `git push origin master --follow-tags`, to ensure that the versions pushed to NPM are synced with the GH repo. You don't think it's a reasonable addition?

RubenVerborgh:
> You don't think it's a reasonable addition?

I mean, I don't really _mind_.

It's just that we are unnecessarily tying two different workflows together.
Just as an example, I myself actually _first_ `git push` and wait for the Travis build to succeed before I `npm publish`.

You might instead want a bash alias in your dotfiles that combines those steps, `alias npmpub='git push origin master --follow-tags && npm publish`. More related aliases here: https://github.com/RubenVerborgh/dotfiles/blob/master/config/npm_config

megoth:
> It's just that we are unnecessarily tying two different workflows together.
> Just as an example, I myself actually _first_ `git push` and wait for the Travis build to succeed before I `npm publish`.
> 
> You might instead want a bash alias in your dotfiles that combines those steps, `alias npmpub='git push origin master --follow-tags && npm publish`. More related aliases here: https://github.com/RubenVerborgh/dotfiles/blob/master/config/npm_config

Personally I don't quite see the problem with mixing these two workflows, and I'd hoped to streamline my own process when publishing to NPM, but fair enough, I'll respect different workflows, and I can fix aliasing for my own personal benefit.

I've removed postpublish altogether from the scripts, are you ok with merging what's left?

PR #398, opened 2020-04-01
fix lang parsing for xml and add a test

### LINKED ISSUE #107: literal with .lang set shoudl have datatype rdf:langString
> if and only if the datatype IRI is http://www.w3.org/1999/02/22-rdf-syntax-ns#langString, a non-empty language tag as defined by [BCP47]. The language tag MUST be well-formed according to section 2.2.9 of [BCP47].

https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal

see also: https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md#literal-extends-term

florent-andre:
fix #107 with suggested edit and create an associated test

### DIFF
--- a/src/rdfxmlparser.js
+++ b/src/rdfxmlparser.js
@@ -153,7 +153,7 @@ export default function RDFParser(store) {
 
         this.addSymbol(this.ARC, uri)
       },         /** Add a literal to this frame */'addLiteral': function (value) {
-        if (this.parent.datatype) {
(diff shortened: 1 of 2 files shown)


TommasoBianchi:
Hey there, any news on this? I am moving towards using this library in production, and this issue is now a blocking one for me.

florent-andre:
I don't know, trying to ping recent maintainers : @megoth @RubenVerborgh 

I would also love to see this fixed upstream

PR #401, opened 2020-04-02
[on hold] Trying to add unit test for sparql-update

michielbdejong:
This PR adds a unit test that fails.
The error seems to come from:
https://github.com/linkeddata/rdflib.js/blob/master/src/n3parser.js#L810

Followed by:
https://github.com/linkeddata/rdflib.js/blob/master/src/patch-parser.js#L70

`res2[0]` there is an `IndexedFormula` containing the triple that was inserted, but it's being used as the object in a triple of the form `<query> snqs:insert [IndexedFormula]`. I think the intention of this code must have been to insert a list or a collection or a blank node or something there.

I'll delve into the git history because I'm pretty sure this did work last year.

### DIFF
--- a/tests/unit/parse-test.js
+++ b/tests/unit/parse-test.js
@@ -2,6 +2,7 @@
 import { expect } from 'chai'
 
 import parse from '../../src/parse'
+import { SPARQLUpdateContentType } from '../../src/types'
 import CanonicalDataFactory from '../../src/factories/canonical-data-factory'
 import defaultXSD from '../../src/xsd'
 import DataFactory from '../../src/factories/rdflib-data-factory'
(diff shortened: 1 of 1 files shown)


michielbdejong:
The error
```
Object {<https://example.com/#s> <https://example.com/#p> <https://example.com/#o> .} is not an object type
```
is probably thrown from https://github.com/linkeddata/rdflib.js/blob/master/src/store.ts#L433

michielbdejong:
Indeed, if you update NSS to the latest RDFLib.js, it breaks 85 tests, including:
```sh
21) PATCH
       with insert
         on a non-existing file
           has "Patch applied successfully" in the response:
     AssertionError: expected 'Patch document syntax error: Error: Error: Object {<https://tim.localhost:7777/x> <https://tim.localhost:7777/y> <https://tim.localhost:7777/z> .} is not an object type while trying to parse <https://tim.localhost:7777/new.ttl#patch-9f1fb311b57c6601db19f09bc4bb597f> as text/n3\n' to include 'Patch applied successfully'
      at Function.assert.include (node_modules/chai/lib/chai/interface/assert.js:843:45)
      at Context.<anonymous> (test/integration/patch-test.js:462:16)
      at processImmediate (internal/timers.js:439:21)

**Comment by michielbdejong** [comment]
So is this valid in NT?
```NT
{
  <https://example.com/#query>
  <http://www.w3.org/ns/pim/patch#insert> {
    <https://example.com/#s>
    <https://example.com/#p>
    <https://example.com/#o> .
  } .
}
```
Because it's what the sparlqUpdateParser tries to add to the store in 
https://github.com/linkeddata/rdflib.js/blob/master/src/patch-parser.js#L70, and the store chokes on it at  https://github.com/linkeddata/rdflib.js/blob/master/src/store.ts#L433.

Issue #404, opened 2020-04-07
Identify core areas of code base


michielbdejong:
In order to maximise impact of code-cleanup, I want to find out which areas of the code base are "core" and which areas areas are more expendable (i.e. it would be possible to use rdflib.js without ever touching those).

michielbdejong:
Under src/ there are 5891 lines of JS and 7259 + 446 + 135 = 7840 lines of TS.
Some things we'll definitely need:
* src/blank-node.ts (101)
* src/fetcher.ts (2124)
* src/formula.ts (935)
* src/index.ts (119)
* src/literal.ts (187)
* src/n3-parser.js (1570)
* src/named-node.ts (113)
* src/serializer.js except statementsToXML (965 - 306)
* src/store.ts except applyPatch (1137 - 95)
* src/statement.ts (132)
* src/serialize.ts (102)
* src/uri.ts (214)

Total: 101 + 2124 + 935 + 119 + 187 + 1570 + 113 + (965 - 306) + (1137 - 95) + 132 + 102 + 214 =

7298

So that's 7298 / (5891 + 7840) = 53%.

michielbdejong:
Sorry, as @megoth pointed out I forgot about src/update-manager.ts, so then the estimate would be (7298 + 1142)/(5891 + 7840) = 61%.

There may also be more parts we can skip, even in these files. For example, I already excluded statementsToXML and applyPatch from the count, but there may be more like that, so consider 61% to be an upper-limit estimate.

emmettownsend:
Without going through it here, we already have the write up and so we know the recommendations we need to follow. 
But in general my thoughts on refactoring in this context:

There are many strategies that can be used but some of them are not on the cards for us at the moment because of time constraints. We have dates that we need to deliver dependent products by. 

Some of the files are way too big and need to be broken up. We need to consider cohesiveness, loose coupling, testability and separation of concerns.  I would not be in favour of refactoring a file at a time because 
  - 1) We do not know that the individual files are cohesive and doing a file at a time will risk not breaking it up properly 
  - 2) We most likely do not need all the code in every file right now so lets only refactor what we need first. 

As we create the current product we should refactor rdflib like this:

 Start: 
  - Pull out a function or class that we need to use into a new library
  - Determine the next dependency that function has
  -  Goto Start
 
This way we only need to refactor the code that we need right now. We will eventually refactor it all but we don't have the need or the time to do it all at once. 

You will end up with just the code we need to refactor.

[...]

Issue #397, opened 2020-03-27
Problems with types


megoth:
I've done some initial work to try to fix type errors that pop up when using the work on the master branch (e.g. for Solid UI it outputs 209 errors).

Doing this I came across a couple of issues that can be split into two groups of problems:

1. Classes that do not extend their parents properly
2. Difficulties in combining types that relate to RDFJS and RDFLIB.

# Problem 1

For problem 1 I've had a talk with Tim, and he thinks some reasonable fixes are:

* To match the signature of `Formula#add` conform to `IndexedFormula#add`
* To rename `IndexedFormula#compareTerm` to `IndexedFormula#compareTerms`

We didn't go into depth of other inconsistencies, but it seems that we will have to break some APIs in any case if we want to fix this properly (currently we've used `@ts-ignore` to tell TypeScript to ignore them - but this won't fix the problem for projects that depend on this classes, as TypeScript will complain about inconsistent types here as well, which leads to a lot of unneccessary uses of ts-ignore).

# Problem 2

Problem 2 seems to come from the fact that many outputs from methods now uses the types described in the [RDF/JS Data model specification](http://rdf.js.org/data-model-spec/) (I'll just refer to them as RDF/JS types later in this text) instead of the corresponding rdflib ones, and this causes errors for projects such as Solid UI which has worked with rdflib types till now. This has been discussed earlier at https://github.com/linkeddata/rdflib.js/issues/374 when @joepio worked on this earlier (and an even older thread at https://github.com/linkeddata/rdflib.js/issues/137).

The data model types exported from rdflib (e.g. by doing `import { NamedNode } from 'rdflib'`) still refer to rdflib types, which causes type errors for projects such as Solid UI (one fix is to change all type references to the corresponding RDF/JS types, e.g. by doing `import { NamedNode } from 'rdflib/src/tf-types'`, but I think this is a very bad solution.

One simple fix is to make sure that the types that TypeScript refers to are in fact referring to the RDF/JS types. But this will make it more difficult for people to use the extra functionality that is available in the rdflib types.

Now, Tim wants to make sure that types returned from rdflib are using the rdflib types (e.g. so that developers have access to some extra methods, such as `toNT`, `doc`, and `site`).

Since they are a superset, one theory of mine was to allow RDF/JS types as input for all relevant methods, but that they output rdflib types. The idea is that developers who want to work with RDF/JS can still do it (since rdflib types are a superset of RDF/JS types).

[...]

PR #406, opened 2020-04-20
Improve typing of DataFactory

RubenVerborgh:


RubenVerborgh:
Sorry, I don't fully understand:

> Works mostly ok with maslib - there's one bug though, which is that it requires termToNQ in canonical-data-factory.ts to handle terms that are graphs.

- Is the bug that `termToNQ` requires handling graphs (as in: it should not)?
- Or is the bug that `termToNQ` does not handle graphs?

Do you have a code example of what breaks?

megoth:
> Sorry, I don't fully understand:
> 
> > Works mostly ok with maslib - there's one bug though, which is that it requires termToNQ in canonical-data-factory.ts to handle terms that are graphs.
> 
>     * Is the bug that `termToNQ` requires handling graphs (as in: it should not)?
> 
>     * Or is the bug that `termToNQ` does not handle graphs?
> 
> 
> Do you have a code example of what breaks?

Sorry for my imprecisessness - if I build the data browser with the current changes, I get the following error:

```
Error: Can't serialize nonstandard term type (was 'Graph')
  https://megoth.localhost:8443/mashlib.js:7308:15
    termToNQ  https://megoth.localhost:8443/mashlib.js:7230:23
    id  https://megoth.localhost:8443/mashlib.js:7432:42
    id  https://megoth.localhost:8443/mashlib.js:10100:30
    id  https://megoth.localhost:8443/mashlib.js:19608:38
    canon  https://megoth.localhost:8443/mashlib.js:20227:27
    statementsMatching  https://megoth.localhost:8443/mashlib.js:10197:22
    each  https://megoth.localhost:8443/mashlib.js:67316:24
    outlineObjectTD  https://megoth.localhost:8443/mashlib.js:4114:35
    refresh/<  https://megoth.localhost:8443/mashlib.js:90436:32
    syncTableToArray  https://megoth.localhost:8443/mashlib.js:4107:18
    refresh  https://megoth.localhost:8443/mashlib.js:4121:7
    render  https://megoth.localhost:8443/mashlib.js:67943:56
    renderPane  https://megoth.localhost:8443/mashlib.js:67975:47
    listen/<  undefined
```

I've pushed a possible solution to [feature/extended-typing-toNQ-fix](https://github.com/linkeddata/rdflib.js/compare/feature/extended-typing...feature/extended-typing-toNQ-fix), but not sure if this is correct.

RubenVerborgh:
Eh where does `Graph` even come from? It is most definitely not an RDF type, nor RDF/JS type.
It shouldn't exist. Let me check that.

RubenVerborgh:
Ok seems like we're still doing

```
export const CollectionTermType = "Collection" as const
export const EmptyTermType = "Empty" as const
export const GraphTermType = "Graph" as const
```

stuff, which we really should not.

However, my main mission was to unblock you on #397, which I think I've achieved.

I think the route for fixing the problem is getting rid of `GraphTermType`, as there really is no such thing; they are either named, blank, or default. In the meantime, workaround `extended-typing-toNQ-fix` will do.

megoth:
> Ok seems like we're still doing
> 
> ```
> export const CollectionTermType = "Collection" as const
> export const EmptyTermType = "Empty" as const
> export const GraphTermType = "Graph" as const
> ```
> 
> stuff, which we really should not.

Should we create a ticket to track this?

> 
> However, my main mission was to unblock you on #397, which I think I've achieved.

Yes, thank you again 😸 

> 
> I think the route for fixing the problem is getting rid of `GraphTermType`, as there really is no such thing; they are either named, blank, or default. In the meantime, workaround `extended-typing-toNQ-fix` will do.

Yeah, I think we should create a ticket to track the issue, and then remove the hack introduced in feature/extended-typing-toNQ-fix.

I've created https://github.com/linkeddata/rdflib.js/pull/407 that can be merged when everything's in order.

pmcb55:
Just my 2cents, but perhaps this strange looking 'Graph' type is one of the N3 types that Tim mentioned were extensions to the standard RDF types...!?

RubenVerborgh:
It's a misnomer in any case then. Notation3 has formulas.

timbl:
> It's a misnomer in any case then. Notation3 has formulas.

N3 has things it calls formulas and other people call graphs. So not really a misnomer. (Like Symbol/NamedNode, Formula/Graph, Statement/Arc, etc these are differences between talking about it  as  a logic language and as a graph.  It is useful to be able to look at it either way, as for core RDF a lot of people thing of the graph, but for N3 you have a language which is  clean superset of the RDF but where is more than a graph)

I suspect the solution is to check if the given thing is  of termtype Graph, and if so raise a run time value error, saying ' This is N3 data, it is too complex to be serialized as an NQuad item" or words to that effect.  I always thought we would need run-time errors when we build the system which mix N3 and and base RDF.    It is fine to have people use rdflib.hs most of the time for plain RDF, but it would be be bad for  the types they are using not t allow them to be able to express more complex things wher ethey ned to. JUst f they try to serialize a query or a rule etc in say turtle ot RDF/XML the system will raise a run time error not a  compile time type mismatch.

RubenVerborgh:
> I suspect the solution is to check if the given thing is of termtype Graph, and if so raise a run time value error, saying ' This is N3 data, it is too complex to be serialized as an NQuad item" or words to that effect.

I think a specific termtype for `Graph` neither necessary nor desired.

Graphs are identified by URIs (or blank nodes), so I'm missing arguments as to why they would be a separate thing.

As a note, N3.js parses Notation3 and remains within the RDF/JS model.

elf-pavlik:
I've noticed #404, maybe it would be possible to have rdflib.js stick to RDF and have distinct n3.js depend on it and extend it with N3 specific functionality. I can imagine developers intending to use rdflib.js to handle RDF getting confused if they see some N3 errors thrown at them.

megoth:
Seems that we need to fix one more type error that was hidden with `@ts-ignore`, namely an incorrect extension of Literal: https://github.com/linkeddata/rdflib.js/blob/master/src/literal.ts#L16

Will try to fix this tonight to get this merged and a new version of rdflib released.

megoth:
> Will try to fix this tonight to get this merged and a new version of rdflib released.

An ugly fix possible, pushed as https://github.com/linkeddata/rdflib.js/pull/414. If this is ok, I think I should be able to merge this to master and release a new version of rdflib.

PR #409, opened 2020-04-27
Fixing tests

megoth:
Reverting some of the changes done earlier - this is hacky, since we're saying one thing (typing defaultGraph as DefaultGraph) while doing one thing (creating defaultGraph as a NamedNode).

Probably needs a more thorough fix, but this was the quickest way I found, and I think it's good enough for now.

### DIFF
--- a/src/factories/canonical-data-factory.ts
+++ b/src/factories/canonical-data-factory.ts
@@ -17,6 +17,7 @@ import {
   GraphTermType
 } from '../types'
 import DefaultGraph from '../default-graph'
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
I can't sign off on this unfortunately; it breaks RDF/JS compatibility.

megoth:
> I can't sign off on this unfortunately; it breaks RDF/JS compatibility.

Ok, that's fair. I'll try implementing a proper fix.

PR #412, opened 2020-04-28
Bug fix and bring up to date `editable` method

timbl:
Remove a kludge which was put in to counter a long gone NSS bug. 

Also fix a bug with getting editable status of a file which is 404 - ie  would the file be editable if it existed?


src/update-manager.ts
```diff
@@ -4,6 +4,7 @@
 ** 2010-08-08 TimBL folded in Kenny's WEBDAV
 ** 2010-12-07 TimBL addred local file write code
 */
+/* source-trace 20209521a */
```
megoth, line 7: Do we need this in master?

src/update-manager.ts
```diff
@@ -769,7 +758,11 @@ export default class UpdateManager {
         (this.store.fetcher.load(doc) as Promise<Response>).then(response => {
           this.update(deletions, insertions, callback, true)
         }, err => {
-          throw new Error(`Update: Can't read ${doc} before patching: ${err}`)
+            if (err.response.status === 404) { // nonexistent files are fine
+              this.update(deletions, insertions, callback, true)
```
megoth, line 762: Maybe create a test to cover this branch?

timbl: Done

src/update-manager.ts
timbl: Removed

PR #414, opened 2020-04-29
An ugly hack to make Literal.fromValue compatible with Node.fromValue

megoth:
Possible solution, although very ugly...


megoth:
Talked Ruben, and we agree that this is a good enough solution for now.

PR #415, opened 2020-04-30
Adding rdflib data types to Store

megoth:
Once I started using the newest version with the new types in SolidOS I realized that the Store methods were not working properly, as they return RDF/JS data types, and SolidOS relies heavily on rdflib data types.

I think the changes should conform to RDF/JS data types, but have some doubts about some of the code I changed, especially IndexedFormula#canon (which, as I understand it, must cast RDF/JS data types to rdflib data types).

I hope this works, as it makes most of the codebase in SolidOS work (there are still _some_ inconsistencies, but I'll create separate issues/PRs for these)

### DIFF
--- a/src/formula.ts
+++ b/src/formula.ts
@@ -122,8 +122,8 @@ export default class Formula extends Node {
   /** Add a statment object
    * @param {Statement} statement - An existing constructed statement to add
    */
-  addStatement (statement: Quad): number {
-    return this.add(statement) as number
+  addStatement (statement: Quad): Statement | null | this | number {
+    return this.add(statement)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/formula.ts
```diff
@@ -137,7 +123,7 @@ export default class Formula extends Node {
    * @param {Statement} statement - An existing constructed statement to add
    */
   addStatement (statement: Quad): number {
-    return this.statements.push(statement)
+    return this.add(statement) as number
```
RubenVerborgh, line 126: The `as number` is weird. If the output type of `add` depends on the input parameter type, then we might want to issue multiple signatures for `add` (which is possible in TypeScript).

PR #427, opened 2020-07-13
Store requestedURI as NamedNode

michielbdejong:
This fixes a bug where https://github.com/linkeddata/rdflib.js/blob/master/src/fetcher.ts#L1600 would return zero results and the update manager would incorrectly end up at https://github.com/linkeddata/rdflib.js/blob/master/src/update-manager.ts#L593.

### DIFF
--- a/src/fetcher.ts
+++ b/src/fetcher.ts
@@ -1597,7 +1597,8 @@ export default class Fetcher implements CallbackifyInterface {
     header: string
   ): undefined | string[] {
     const kb = this.store
-    const requests = kb.each(undefined, this.ns.link('requestedURI'), doc) as Quad_Subject[]
+    const docuri: string = doc.value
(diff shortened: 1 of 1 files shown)


michielbdejong:
Another option for fixing this would be to leave the storing code as it is (storing a literal), and then change the querying code to look for a literal that has the doc's URI as its value, but that felt less clean.

Not sure if there are other consideration or other places that depend on storing the requestedURI as a literal.
I did find `addType` which seems to use a NamedNode for the requestedURI as a subject in a type triple.

michielbdejong:
@timbl I made the changes you requested.
Tested this using our new https://github.com/solid/mashlib-dev#debugging-rdflib-using-solid-pane-tester and it works.
I'll interpret this as good to merge now, if someone knows of further improvements we can do them in follow-up.

Issue #408, opened 2020-04-22
Breaking change in JSON-LD parsing?


howarddierking:
Hi - apologies in advance if I missed this somewhere, but was there a breaking change introduced to how JSON-LD is parsed in one of the last versions. 

The following was working in `0.19.0` 

Test file
```
const explicit = `{
  "@context": { 
    "@base": "http://schema.howarddierking.com/",
    "@vocab": "http://schema.org/"
  },
  "@type": "Person",
  "@id": "/howard-dierking",
  "name": "Howard Dierking",
  "givenName": "Howard",
  "familyName": "Dierking",
  "url": "https://www.howarddierking.com"
}`;

describe('index', () => {
	describe('#getFirstName', () => {
		it('should return expected value for explicit triples', (done) => {
			R.pipeP(subject.graphFor, subject.getFirstName)(explicit)
			.then(actual => {
				actual.should.eql('Howard');
				done();
			});
		});
		
                // more tests ...
	})
})
```

Relevant src file
```
const jsonLd = 'application/ld+json';
const parseRdf = P.promisify(rdf.parse);

function graphFor(s){
	let g = rdf.graph();

	return parseRdf(s, g, null, jsonLd);
}
```

Under `1.1.0`, this code now fails because what looks to be zero items being loaded into the store.  Is there a known change here in rdflib (or the jsonld lib)?

josephguillaume:
It looks like this issue is because the change to reduce code bloat in Oct 2019 https://github.com/linkeddata/rdflib.js/issues/349 did not convert `@type` to `rdf:type` (see comment on code: https://github.com/linkeddata/rdflib.js/commit/16daad7d68d805aa29e8e01bb4a16d77b28b6afe#r41313390), and errors were also accidentally suppressed (#435)

Here's a simpler test case (though note that parse is async for json-ld)
```
      g = $rdf.graph();
      const explicit = `{
        "@context": {
          "@base": "http://schema.howarddierking.com/",
          "@vocab": "http://schema.org/"
        },
        "@type": "Person",
        "@id": "/howard-dierking",
        "name": "Howard Dierking",
        "givenName": "Howard",
        "familyName": "Dierking",
        "url": "https://www.howarddierking.com"
      }`;
      $rdf.parse(explicit, g, null, "application/ld+json");
      g.statements
```

Issue #343, opened 2019-08-27
Fetching JSON-LD does not appear to work


dantman:
Fetching a JSON-LD document does not appear to put anything into the store.

I tested this out with 1.0.4 by trying to fetch various urls with `nowOrWhenFetched` and dumping the store with toString when fetched.

I verified my test for fetching data worked by fetching Turtle (application/x-turtle) data. This worked, dumping a pile of data from Book.ttl into the store.
```js
var $rdf = require('rdflib');
var store = $rdf.graph();
var fetcher = new $rdf.Fetcher(store, 5000);
fetcher.nowOrWhenFetched('http://schema.org/Book.ttl', () => { console.log(store.toString()); });
```

Fetching RDF data failed, however that's fine because schema.org is incorrectly returned a application/xml content-type.
```js
var $rdf = require('rdflib');
var store = $rdf.graph();
var fetcher = new $rdf.Fetcher(store, 5000);
fetcher.nowOrWhenFetched('http://schema.org/Book.rdf', () => { console.log(store.toString()); });
```

But trying to fetch the JSON-LD version of the document also failed. This failure was not expected because the server correctly sends back a `application/ld+json; charset=utf-8` content-type.
```js
var $rdf = require('rdflib');
var store = $rdf.graph();
var fetcher = new $rdf.Fetcher(store, 5000);
fetcher.nowOrWhenFetched('http://schema.org/Book.jsonld', () => { console.log(store.toString()); });
```

angelo-v:
Fixed in https://github.com/linkeddata/rdflib.js/pull/468

bourgeoa:
@angelo-v #468 only soved the predicate part.
see PR #469 that should solve the predicate and object when using @type as a node type or a value type per JSON-LD specification https://json-ld.org/spec/latest/json-ld/#typed-values.

PR #430, opened 2020-07-19
WIP: Fetching triples from a JSON-LD document

angelo-v:
I added a failing test for a problem I observed fetching data from JSON-LD documents. The PR does not include a fix yet. Reviews, if the test actually uncovers a bug that needs fixing are welcome. The same is true for ideas how to fix.

### DIFF
--- a/tests/unit/fetcher-jsonld-test.js
+++ b/tests/unit/fetcher-jsonld-test.js
@@ -1,0 +1,40 @@
+import * as rdf from '../../src';
+
+const nock = require('nock')
+import chai from 'chai'
+
+const {expect} = chai
+
+describe('Given a JSON-LD resource', () => {
+
+    const uri = "http://localhost/jsonld#it"
+
+    beforeEach(() => {
+        const docContents = `
+        {
+            "@id": ${uri},
+            "https://predicate.example": "value"
+        }
+        `
+        nock('http://localhost').get('/jsonld').reply(200, docContents, {
+            'Content-Type': 'application/ld+json',
+        })
+    });
+
+    describe('when it is fetched to the store', () => {
+
+        let fetcher, store;
+        beforeEach(() => {
+            store = rdf.graph();
+            fetcher = rdf.fetcher(store);
+        });
+
+        it('then the triples from the document can be found in the store', async () => {
+            await fetcher.load(uri);
+            const match = store.anyStatementMatching(rdf.sym(uri))
+            expect(match).to.be.ok()
+        });
+    });
+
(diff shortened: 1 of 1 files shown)


tests/unit/fetcher-jsonld-test.js
```diff
@@ -0,0 +1,40 @@
...
+const {expect} = chai
+
+describe('Given a JSON-LD resource', () => {
+
+    const uri = "http://localhost/jsonld#it"
+
+    beforeEach(() => {
+        const docContents = `
+        {
+            "@id": ${uri},
```
josephguillaume, line 15: There's a small typo here:
```suggestion
            "@id": "${uri}",
```
I haven't run the test, but if it passes now, then maybe the issue you were facing was #408? 
Note also that another bug caused jsonld errors to be suppressed: #435 

I've added some annotations to the relevant commit https://github.com/linkeddata/rdflib.js/commit/16daad7d68d805aa29e8e01bb4a16d77b28b6afe#r41313390

josephguillaume: Or is there a separate issue related to fetching, like in https://github.com/linkeddata/rdflib.js/issues/343?

angelo-v: Good catch! Fixed the syntax. But test is still failing for the same reason. I think the issue is in the fetching / response handling.

There is just no default handler for JSON-LD?

```
const defaultHandlers = {
  RDFXMLHandler, XHTMLHandler, XMLHandler, HTMLHandler, TextHandler, N3Handler
}
```

michielbdejong:
This also makes NSS fail the content negotiation tests I'm adding in https://github.com/solid/solid-crud-tests/commit/9d4928b8abedb0f791b0f2a92a6a9a9e0d4eb774

michielbdejong:
I'll see what I can do to try to fix this bug.

michielbdejong:
Indeed it cannot find a JSON-LD handler, these are my debug statements from `handlerForContentType`:
```
handle response application/ld+json
checking pattern /application\/rdf\+xml/ null
checking pattern /application\/xhtml/ null
checking pattern /(text|application)\/(.*)xml/ null
checking pattern /text\/html/ null
checking pattern /text\/plain/ null
checking pattern /(application|text)\/(x-)?(rdf\+)?(n3|turtle)/ null
no handler
```

michielbdejong:
Continued in https://github.com/linkeddata/rdflib.js/pull/439

PR #431, opened 2020-07-24
Update many. Fix getHeader problem

timbl:
UpdateMany is a function we had said we would would try to get people using so that when Solid is extended to allow it in one HTTP request, that can all take advantage of it.  In the mean time, we ned to test it and see whether people can manage the error handling when some fail and some don't.

The bug in fetcher was introduced I suspect during the type script conversion.
A URI is a string not a node.

### DIFF
--- a/src/fetcher.ts
+++ b/src/fetcher.ts
@@ -1596,8 +1596,8 @@ export default class Fetcher implements CallbackifyInterface {
     doc: NamedNode,
     header: string
   ): undefined | string[] {
-    const kb = this.store
-    const requests = kb.each(undefined, this.ns.link('requestedURI'), doc) as Quad_Subject[]
+    const kb = this.store // look for the URI (AS A STRING NOT A NODE) for a stored request
+    const requests = kb.each(undefined, this.ns.link('requestedURI'), doc.uri) as Quad_Subject[]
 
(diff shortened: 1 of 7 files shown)


src/fetcher.ts
```diff
@@ -1596,8 +1596,8 @@ export default class Fetcher implements CallbackifyInterface {
     doc: NamedNode,
     header: string
   ): undefined | string[] {
-    const kb = this.store
-    const requests = kb.each(undefined, this.ns.link('requestedURI'), doc) as Quad_Subject[]
+    const kb = this.store // look for the URI (AS A STRING NOT A NODE) for a stored request
+    const requests = kb.each(undefined, this.ns.link('requestedURI'), doc.uri) as Quad_Subject[]
```
SharonStrats, line 1600: I am wondering if we should add uri to tf-types.ts under export interface NamedNode extends Term..  line 30.  This might solve the build issue, looking at how we handled this in solid-ui.

SharonStrats: index.d.ts in solid-ui has uri as a member of the class NamedNode which extends Node....

michielbdejong: This was already fixed in the master branch.

PR #436, opened 2020-08-12
Fix JSON-LD tests

michielbdejong:
Not sure why this test was failing but I edited it until it worked in this branch.


michielbdejong:
Ah actually they fail on Travis now... :) 
I was running this test in isolation, I guess it's not resetting things properly.

michielbdejong:
Fixed in #439

PR #438, opened 2020-08-31
Default to cross-fetch

michielbdejong:


michielbdejong:
@jaxoncreed can you review?
I think this is the way to go, but it would be a semver-major change for people who currently use the fact that solid-auth-client / solid-auth-cli are the defaults.

This is the branch I use in https://github.com/solid/solid-crud-tests/blob/master/package.json#L36

I'll see if I can fix the tests on this branch.

jeff-zucker:
I would imagine that  most apps/libraries/users rely on the solid-auth* fetches - certainly for all authenticated fetches.  If I understand the change, users will need to pass an alternate fetch in via the options each time they create a fetcher which means possibly multiple changes per script and would mean that any intermediate libraries between the app and rdlib would also need to have made the changes.  I'm wondering whether a global window.solidFetcher variable would make sense as an alternate way to specify rdflib's fetch.

jeff-zucker:
If a global variable is used, we may want to think about consolidating other global variables into a solidEnvironment object - rdflib uses window.$solidTestEnvironment, solid-client-authn-browser uses window.authSession.

michielbdejong:
A global would work, yes, but I would prefer to pass a fetcher object in the options whenever you instantiate rdflib, it feels cleaner. For instance in the Solid test suite, multiple tests run in parallel and I would like to avoid having to use globals there, as it feels like it would risk effects bleeding from one test to another. I'm already using this branch there.

I also see @jeff-zucker's point though, and don't want to cause you work having to update your scripts if we can avoid ti.

I would be open to doing both: check for an explicit fetcher in options first, if missing, check a global variable, if that's missing too, fall back to something like cross-fetch. Or even throw an error saying 'please specify a fetcher' if we think cross-fetch is probably not what people want.

@jaxoncreed can you find a way to get rid of these hard-coded peerDependencies, but also keep @jeff-zucker's code running, as well as the code in https://github.com/solid/solid-crud-tests/blob/master/package.json#L36 which already uses the branch from this PR?

michielbdejong:
I'm still struggling to let `tsc --noEmit --target es2019 tests/types/*.ts` find cross-fetch (it uses esnext module resolution).

michielbdejong:
@joepio Is https://github.com/linkeddata/rdflib.js/pull/438/commits/b2fc8ad59fc437af16ac34792a85604448d975f4 OK? I read that `tsc` will default to [Classic module resolution](https://www.typescriptlang.org/docs/handbook/module-resolution.html#classic) if you set the target to es2019, and then it doesn't look in node_modules, which I guess can't be what we want, right? VSCode also complains about the line where I include cross-fetch in src/fetcher.ts.

jaxoncreed (approved):
I agree this is the right way to go. If you want an authenticated fetcher, you should pass it in yourself (especially because the new interface requires one instance of an authentication client).

timbl:
Do you really mean window.solidFetcher not window.solidFetch?  It is the fetch function, is it  not?

PR #439, opened 2020-09-01
Add JSON-LD handler to fetcher

michielbdejong:


tests/unit/fetcher-jsonld-test.js
```diff
@@ -0,0 +1,40 @@
...
+        let fetcher, store;
+        beforeEach(() => {
+            store = rdf.graph();
+            fetcher = rdf.fetcher(store);
+        });
+
+        it('then the triples from the document can be found in the store', async () => {
+            await fetcher.load(uri);
+            const match = store.anyStatementMatching(rdf.sym(uri))
+            expect(match.subject.value).to.equal(uri)
```
angelo-v, line 35: I prefer to expect the specific triple from above here

src/fetcher.ts
```diff
@@ -479,6 +480,39 @@ class HTMLHandler extends Handler {
 }
 HTMLHandler.pattern = new RegExp('text/html')
 
+class JsonLdHandler extends Handler {
```
angelo-v, line 483: ok, we actually did not have a JsonLdHandler anywhere, yet? :eyes:  Thanks for writing one :+1:

tests/unit/fetcher-jsonld-test.js
```diff
@@ -0,0 +1,40 @@
...
+        });
+    });
+
+
+});
```
timbl, line 40: Good to have newlines at end of files. (Some unix utilities ignore the widow line)

tests/unit/fetcher-test.js
```diff
@@ -372,7 +372,7 @@ describe('Fetcher', () => {
     it('should set the Accept header', () => {
       options = fetcher.initFetchOptions(uri, options)
 
-      let expectedHeader = 'image/*;q=0.9, */*;q=0.1, application/rdf+xml;q=0.9, application/xhtml+xml, text/xml;q=0.5, application/xml;q=0.5, text/html;q=0.9, text/plain;q=0.5, text/n3;q=1.0, text/turtle;q=1'
+      let expectedHeader = 'image/*;q=0.9, */*;q=0.1, application/rdf+xml;q=0.9, application/xhtml+xml, text/xml;q=0.5, application/xml;q=0.5, text/html;q=0.9, text/plain;q=0.5, text/n3;q=1.0, text/turtle;q=1, application/ld+json;q=0.4'
```
timbl, line 375: I think should be `application/ld+json;q=0.9` as, given a JSDON/LD file we parse it completely with no data loss.  The only reason we don't put it as 1.0 is for systems which can provide any equally well we prefer turtle as it tends to be smaller, simpler, and should be faster to parse.

Issue #204, opened 2017-10-18
File:/// access has been dropped


timbl:
ooops ... running rabel the command line client
https://github.com/linkeddata/rabel
using the current rdflib, it gives a "status 0" error to the console, and debugging reveals that the access to fileL URIs no longer works at all, as the fetcher now uses, in a node.js environment, node-fetch module which includes:

```
	return new Fetch.Promise(function(resolve, reject) {
		// build request object
		var options = new Request(url, opts);

		if (!options.protocol || !options.hostname) {
			throw new Error('only absolute urls are supported');
		}

		if (options.protocol !== 'http:' && options.protocol !== 'https:') {
			throw new Error('only http(s) protocols are supported');
		}
```
Maybe this a question of switching to a different one of many fetch modules.  File:// access is important, for command line apps, test suites, and browser extensions also can also access files.
Surprised we had no tests.

dmitrizagidulin:
The reason that the node fetch modules dont support file:// urls is because the in-browser native fetch implementations also forbid file urls. (And they do that for security reasons).

So, this is unlikely to be able to be changed.

ericprud:
I don't know if other programs use node-fetch but Fetcher uses node-fetch only on node:

    if (!this._fetch)
      if (typeof window !== 'undefined')
        this._fetch = window.fetch.bind(window);
      else
        this._fetch = require('node-fetch');

This implies that node-fetch could be useful for e.g. file.

timbl:
The logic that file access isn't allowed in browsers and so it won't happen on the command line I don't follow.   No reason to make restrictions from the browser environment constrain the command line. The command line environment is really valuable, for testing, for utility programs, and for batch processing, and so on.  It used to work with XHR.  The fact that it doesn't now is a serious bug introduced by the refactoring.

timbl:
Just as we have a need to have a separate fetch for file: URIs, also @thewebalyst was asking about allowing other URI schemes for other protocols such as decentralized storage spaces.  Suggests having a top layer that dispatches based on URI scheme.

timbl:
I note rabel works with rdflib@0.14.1

dmitrizagidulin:
The reason that file access was dropped in browser's `fetch` API is the same reason most Node fetch libs don't support it -- it's a serious security risk (with people being able to potentially request arbitrary files on the server). With Solid servers specifically, the risk is somewhat mitigated with the ACL mechanism, but I still think it should not be in any way encouraged.

[...]

PR #444, opened 2020-10-16
support file:/// URIs

jeff-zucker:
Adds support for wac-allow and other headers in fetcher and updater.  This will allow rdflib to make use of solid-rest or other package to provide full access to file:/// (and other non-http requests such as cloud or in-browser storage).   Should close issue #204.


jeff-zucker:
Closing this and resumbitting against 2.1.2 master.

PR #447, opened 2020-10-17
Use default Header type for request options

### LINKED ISSUE #446: fetcher forces use of Headers type in request options
To achieve https://github.com/solid/solid-spec/blob/master/api-rest.md#creating-containers-directories

one can do the following in Javascript:
```javascript
const request = {
  headers: {
    Slug: 'Untitled Folder',
    Link: '<http://www.w3.org/ns/ldp#BasicContainer>; rel="type"',
    'Content-Type': 'text/turtle',
  },
};

return this.fetcher.webOperation('POST', folderAddress, request);
```

However in typescript this is not possible because the headers type of `Record<string, string>` is not assignable to fetchers' AutoInitOptions that extend RequestInit and force the use of `Headers` type instead of `Headers | Record<string, string> | string[][]`

Is there a specific reason why headers was overwritten in AutoInitOptions? I get errors when using the Headers constructor for the headers and I would like to use the more loose typing of headers again...

ludwigschubi:
Closes #446

### DIFF
--- a/src/fetcher.ts
+++ b/src/fetcher.ts
@@ -191,7 +191,7 @@ interface AutoInitOptions extends RequestInit{
   req: BlankNode
   // Might be the same as Options.data
   body?: string
-  headers: Headers
+  headers: HeadersInit
   credentials?: 'include' | 'omit'
 }
 


michielbdejong:
LGTM but I would like to get confirmation from @joepio that this is the right way to go?

joepio (approved):
Seems sensible, LGTM

PR #454, opened 2020-11-06
Timeouts - change default timeout in fetcher  from 30s to 2min

timbl:
This is to avoid spurious timeouts called on things which in the end come though, when the net becomes slow for example, or for users on a slow link.


timbl:
Responds to https://github.com/linkeddata/rdflib.js/issues/455

michielbdejong:
Hm, is this for one single http roundtrip?
When using generators this might be desirable, but when awaiting a result, 2 minutes seems like a lot.
rdflib is used in many places. We can choose any default we want, as long as it's well documented. For instance in the test-suite, I would overwrite that back to 30 seconds, probably.

jeff-zucker:
The user can easily override the timeout setting if needed.  Erring on the long side seems preferable to too short a timeout.  I suppose that the timeout could be overwritten in the global soidFetcher  to reset the timeout everywhere for apps that create multiple fetchers?

michielbdejong:
Hm, but what about my comment in https://github.com/linkeddata/rdflib.js/issues/455#issuecomment-723077526 / https://gitter.im/solid/solidcommunity.net?at=5fa54c68c10273610aeca65a ?

timbl:
"One single http round trip" isn't of course a singe IP round trip, as it can have all the TLS setup time, and CORS prefetches... it also can be delayed by the browser by being put in a pending state for a while.    So when all this happens n a slow link,  it is frustrating when the data actually eventually arrives but the client has given up and ignored it.

timbl:
Yes, agreed that the fetch with the server should be set shorter ... maybe the thing to do is to cover the whole service of one incoming request, not each individual outgoing one.  Would be good to to tell the client in the 500 response what it was doing when it gave up of course.

PR #460, opened 2020-11-22
Add force option for updating resources

ludwigschubi:
I had some problems passing the "editable" check when using the update manager with [solid-auth-cli](https://github.com/jeff-zucker/solid-auth-cli)


jeff-zucker:
I'm pretty sure this issue was solved in https://github.com/linkeddata/rdflib.js/pull/464.  Rather than forcing an update, solid-rest now sends the correct wac-allow and sparql-update headers so that a force isn't necessary for file:// URIs handled by solid-node-client.

ludwigschubi:
@jeff-zucker cool 👍

PR #464, opened 2020-11-30
support non-http resources

jeff-zucker:
This PR makes rdflib's update-manager available to non-http fetchers that supply wac-allow and other related headers, a critical feature for apps like data-kitchen and for the local-storage part of gitter-solid.

With this PR, rdflib will now assesses the editability of a non-http response (file: etc)  the same way it does with http responses - it will check the wac-allow header to see if the resource is writable/appendable, then check the accept-patch and ms-authorVia headers to see if the resource can be edited with a SPARQL patch. 

Rdflib will also now write to local files the same way it does for web resources - with a fetcher.webOperation PUT.  This replaces the firefox-extension specific writing.  If someone wants to write in an extension, they can create a solid-rest plugin or create an alternate fetcher which does its own PUT.  Either way, the actual writing belongs in the fetcher, not in rdflib.

The PR does not put anything in rdflib specific to solid-node-client or solid-rest, but it will now respect the headers they (or other alternate fetchers) send.  Solid rest now checks the native file permissions and creates a wac-allow header that gives wac write/append/control if the current user has write permissions to the resource on the native file system and gives wac read permissions to  both user and public if the user has read permission on the native resource.

If the user invokes a file: handler other than solid-rest/solid-node-client (or defaults to cross-fetch), there may not be any of the relevant  headers.  In that case, we fall back to the old method - check for triple containing "MachineEditableDocument".

Changes in this PR :

  * fetcher now checks for global.solidFetcher, not just window.solidFetcher
    (otherwise we can't set the global fetcher in nodejs)

  * fetcher now saves the metadata in headers from non-http responses

  * update-manager does not fail if MachineEditableDocument is missing,
    it now checks for wac-allow headers and only fails if they are missing too

  * update-manager now checks accept-patch and ms-authorVia headers to
    determine SPARQL editability of non-http resources

  * update-manager now saves local files via webOperation/PUT rather than the
    previous firefox-extension specific method

  * checks for 'file://' are replaced with checks for !http - this supports
    handling of any alternate protocol that supplies appropriate headers
    e.g. solid-rest currently supports app://

### DIFF
--- a/src/fetcher.ts
+++ b/src/fetcher.ts
@@ -116,8 +116,8 @@ declare global {
     panes?: any
     solidFetcher?: any
   }
+  var solidFetcher:Function
 }
-
 declare var $SolidTestEnvironment: {
   localSiteMap?: any
 }
@@ -752,7 +752,10 @@ export default class Fetcher implements CallbackifyInterface {
     this.ns = getNS(this.store.rdfFactory)
     this.timeout = options.timeout || 30000
 
-    this._fetch = options.fetch || (typeof window !== 'undefined' && window.solidFetcher) || crossFetch
+    this._fetch = options.fetch
+               || (typeof global !== 'undefined' && global.solidFetcher)
+               || (typeof window !== 'undefined' && window.solidFetcher)
+               || crossFetch
     if (!this._fetch) {
       throw new Error('No _fetch function available for Fetcher')
     }
@@ -1694,7 +1697,7 @@ export default class Fetcher implements CallbackifyInterface {
     kb.add(responseNode, this.ns.http('statusText'),
     kb.rdfFactory.literal(response.statusText), responseNode)
 
-    if (!options.resource.value.startsWith('http')) {
+    if (!response.headers) {
       return responseNode
     }
 
--- a/src/fetcher.ts
+++ b/src/fetcher.ts
@@ -1697,10 +1697,6 @@ export default class Fetcher implements CallbackifyInterface {
     kb.add(responseNode, this.ns.http('statusText'),
     kb.rdfFactory.literal(response.statusText), responseNode)
 
-    if (!response.headers) {
-      return responseNode
-    }
-
     // Save the response headers
     response.headers.forEach((value, header) => {
       kb.add(responseNode, this.ns.httph(header), this.store.rdfFactory.literal(value), responseNode)
--- a/src/update-manager.ts
+++ b/src/update-manager.ts
@@ -106,7 +106,14 @@ export default class UpdateManager {
     }
     uri = termValue(uri)
 
-    if ((uri as string).slice(0, 8) === 'file:///') {
+    // new logic for requests fron non-http resources
+    //   if not http(s) request
+    //     if has MachineEditableDocument triple -> EDIT AS LOCALFILE
+    //     else if !wac-allow write/append in header -> NOT EDITABLE
+    //     else
+    //       if accept-patch or ms-authorVia -> EDIT AS SPARQL
+    //       else -> EDIT AS LOCALFILE
+    if ((uri as string).slice(0, 4) != 'http') {
       if (kb.holds(
           this.store.rdfFactory.namedNode(uri),
           this.store.rdfFactory.namedNode('http://www.w3.org/1999/02/22-rdf-syntax-ns#type'),
@@ -120,7 +127,8 @@ export default class UpdateManager {
       //   uri + '\n')
       // console.log(sts.map((x) => { return (x as Statement).toNT() }).join('\n'))
 
-      return false
+      // only return false later if no wac-allow write/append perms in header
+      // return false
       // @@ Would be nifty of course to see whether we actually have write access first.
     }
 
@@ -163,6 +171,13 @@ export default class UpdateManager {
               }
             }
           }
+
+          // honor wac-allow headers if present in non-http request
+          if ((uri as string).slice(0, 4) != 'http') {
+            if( !wacAllow ) return false;
+            else return 'LOCALFILE';
+          }		  
+
           var status = kb.each(response, this.ns.http('status'))
           if (status.length) {
             for (let i = 0; i < status.length; i++) {
@@ -957,7 +972,8 @@ export default class UpdateManager {
   updateLocalFile (doc: NamedNode, ds, is, callbackFunction): void {
     const kb = this.store
     // console.log('Writing back to local file\n')
-    // See http://simon-jung.blogspot.com/2007/10/firefox-extension-file-io.html
+    // previously used http://simon-jung.blogspot.com/2007/10/firefox-extension-file-io.html ... now uses PUT
+
     // prepare contents of revised document
     let newSts = kb.statementsMatching(undefined, undefined, undefined, doc).slice() // copy!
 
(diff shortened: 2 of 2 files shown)


src/update-manager.ts
```diff
@@ -106,7 +106,14 @@ export default class UpdateManager {
     }
     uri = termValue(uri)
 
-    if ((uri as string).slice(0, 8) === 'file:///') {
+    // new logic for requests fron non-http resources
```
angelo-v, line 109: **nitpick**: typo `fron`

src/update-manager.ts
```diff
@@ -120,7 +127,8 @@ export default class UpdateManager {
       //   uri + '\n')
       // console.log(sts.map((x) => { return (x as Statement).toNT() }).join('\n'))
 
-      return false
+      // only return false later if no wac-allow write/append perms in header
```
angelo-v, line 130: **suggestion:** Let's just remove all the commented out code here

src/update-manager.ts
```diff
@@ -106,7 +106,14 @@ export default class UpdateManager {
     }
     uri = termValue(uri)
 
-    if ((uri as string).slice(0, 8) === 'file:///') {
+    // new logic for requests fron non-http resources
+    //   if not http(s) request
```
angelo-v, line 110: **suggestion**: Instead of having a high-level comment here, let the the code speak by itself by moving all those parts to functions with a respective name

src/update-manager.ts
```diff
@@ -163,6 +171,13 @@ export default class UpdateManager {
               }
             }
           }
+
+          // honor wac-allow headers if present in non-http request
+          if ((uri as string).slice(0, 4) != 'http') {
```
angelo-v, line 176: **suggestion**: Extracting the condition to a function like `isHttpUri(uri)` would make the code better readable

src/update-manager.ts
```diff
@@ -957,7 +972,8 @@ export default class UpdateManager {
   updateLocalFile (doc: NamedNode, ds, is, callbackFunction): void {
     const kb = this.store
     // console.log('Writing back to local file\n')
-    // See http://simon-jung.blogspot.com/2007/10/firefox-extension-file-io.html
+    // previously used http://simon-jung.blogspot.com/2007/10/firefox-extension-file-io.html ... now uses PUT
```
angelo-v, line 975: **question:** So is this link still relevant or not? If not let's remove it, otherwise it will only lead to confusion

src/update-manager.ts
```diff
@@ -981,44 +997,19 @@ export default class UpdateManager {
...
-    stream.close()
-
-    for (let i = 0; i < ds.length; i++) {
-      kb.remove(ds[ i ])
-    }
-    for (let i = 0; i < is.length; i++) {
-      kb.add(is[ i ].subject, is[ i ].predicate, is[ i ].object, doc)
-    }
-    callbackFunction(doc.value, true, '') // success!
+    kb.fetcher.webOperation('PUT',doc.value,{
```
angelo-v, line 1000: **praise:** So this got a lot more straight-forward. Nice!

src/update-manager.ts
```diff
@@ -981,44 +997,19 @@ export default class UpdateManager {
...
+      "body"      : documentString,
+      contentType : contentType,
+    }).then( (response)=>{
+      if(!response.ok) return callbackFunction(doc.value,false,response.error)
+      for (var _i13 = 0; _i13 < ds.length; _i13++) {
```
angelo-v, line 1005: **suggestion**: Those variable names really confused me on first sight. I propose to change it back to `let i`

angelo-v (changes requested):
Nice work @jeff-zucker ! I left some inline comments.

Can we add some automated tests for the new behaviour? rdflib seems not very well tested until now, but I think we need to change that asap. If you need help with that, let me know

src/fetcher.ts
```diff
@@ -1694,7 +1697,7 @@ export default class Fetcher implements CallbackifyInterface {
     kb.add(responseNode, this.ns.http('statusText'),
     kb.rdfFactory.literal(response.statusText), responseNode)
 
-    if (!options.resource.value.startsWith('http')) {
+    if (!response.headers) {
```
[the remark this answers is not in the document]
jeff-zucker: Yeah, I wondered about that. I don't see any harm in saving headers from "dumb clients" (e.g. that don't send wac-allow), but maybe there was a reason to prevent saving them?

src/update-manager.ts
jeff-zucker: No, if we remove the code for it, it is only historically relevant.

angelo-v: **suggestion:** Let's remove it then. It is still present in the git history, together with the code it refers to

jeff-zucker:
I have implemented all of @angelo-v 's suggestions above except adding tests.  I will do that in the future.

src/update-manager.ts
```diff
@@ -106,30 +112,13 @@ export default class UpdateManager {
...
 
-    // new logic for requests fron non-http resources
-    //   if not http(s) request
-    //     if has MachineEditableDocument triple -> EDIT AS LOCALFILE
-    //     else if !wac-allow write/append in header -> NOT EDITABLE
-    //     else
-    //       if accept-patch or ms-authorVia -> EDIT AS SPARQL
-    //       else -> EDIT AS LOCALFILE
-    if ((uri as string).slice(0, 4) != 'http') {
+    if ( this.isHttpUri(uri as string) ) {
```
angelo-v, line 115: **issue**: You reverted the logic here from `!= http` to `isHttp`, I guess that is not intentional
**suggestion**: Unit Tests could catch such mistakes
**question:** Why do we need to cast `as string`?

src/update-manager.ts
```diff
@@ -172,8 +161,7 @@ export default class UpdateManager {
             }
           }
 
-          // honor wac-allow headers if present in non-http request
-          if ((uri as string).slice(0, 4) != 'http') {
+          if ( this.isHttpUri(uri as string) ) {
```
angelo-v, line 164: **issue:** Same as above

angelo-v (changes requested):

src/update-manager.ts
jeff-zucker: OOPS! :-)

jeff-zucker:
I would see these as the appropriate unit tests.  Am I missing any?
```
non-http resources                                                              
  it can use global.solidFetcher in a nodejs environment                        
  it can update a file if it has a MachineEditableDocument triple               
  it can update a file if it has Write perms in its wac-allow header            
  it refuses to update a file that lacks both MchineEditableDoc and wac-allow   
  it can updata a file via SPARQL patch if its headers support it               
  it can updata a file via text/n3 if its headers support it                    
  it can updata a file (by puting back full store) if it lacks accept-patch     
```

jeff-zucker:
@angelo-v can we merge this and I'll open a separate issue for tests.  Without this merge, it is difficult to document or work on solid-node-client or data-kitchen.

angelo-v (approved):

angelo-v:
Yes that's totally fine for me

timbl (approved):
LGTM

angelo-v:
So, who will merge this? I do not have the required permission. @jeff-zucker can you do it yourself? Or @timbl ?

jeff-zucker:
No, I can't merge, @angelo-v thanks much for following up.

PR #468, opened 2020-12-18
properly parse JSON-LD resources with a @type property

angelo-v:
fixes https://github.com/linkeddata/rdflib.js/issues/408 and https://github.com/linkeddata/rdflib.js/issues/343

### DIFF
--- a/src/jsonldparser.js
+++ b/src/jsonldparser.js
@@ -82,12 +82,13 @@ export default function jsonldParser (str, kb, base, callback) {
           continue
         }
         const value = flatResource[property]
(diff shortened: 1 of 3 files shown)


jeff-zucker (approved):

angelo-v:
load() is working properly, but I just found, there is another issue with nowOrWhenFetched, I will take a look at it and see if it fits into this PR or I will do another one

angelo-v:
ah ok, no parsing with nowOrWhenFetched also works fine, I was just using it wrong (promise instead of callback)

PR #469, opened 2020-12-26
add @type statement function

bourgeoa:
@angelo-v @jeff-zucker 

this allow JSON-LD parse to comply with @type being a node type or a value type
This allow NSS to pass JSON-LD CRUD tests

### DIFF
--- a/src/jsonldparser.js
+++ b/src/jsonldparser.js
@@ -30,6 +30,7 @@ export function jsonldObjectToTerm (kb, obj) {
   }
 
   if (Object.prototype.hasOwnProperty.call(obj, '@type')) {
+    console.log('@type1 ' + Object.keys(obj))
     return kb.rdfFactory.literal(obj['@value'], kb.rdfFactory.namedNode(obj['@type']))
   }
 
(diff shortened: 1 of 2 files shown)


src/jsonldparser.js
```diff
@@ -97,3 +96,17 @@ export default function jsonldParser (str, kb, base, callback) {
...
+
+/**
+ * statement quad depending on @type being a type node or a type value
+ * @param kb
+ * @param subject id
+ * @param property
+ * @param value
+ * @return quad statement
+ */
+function statementFromType(kb, id, property, value) {
```
angelo-v, line 108: **praise**: Good catch! Thanks for fixing.

**thought**: However I find this code confusing, actually only understood what is the matter after looking at the test. Why is it called "FromType"? It is used for all properties, just handles `@type` differently. Perhaps just `createStatement` or `statementFromProperty`? Also, instead of heaving 2 ternaries it might be better readable to just differ between `if property === "@type"` and what happens `else`

bourgeoa: Thanks and agreed. Changes done

angelo-v: Thank you very much!

Issue #215, opened 2018-01-12
RDF+XML serializer never produces typed node elements


akoptelov:
RDF specification states that typed nodes can be represented in XML 
> ... more concisely by replacing the rdf:Description node element name with the namespaced-element corresponding to the IRI of the value of the type relationship"

(see [2.13 Typed Node Elements](https://www.w3.org/TR/rdf-syntax-grammar/#section-Syntax-typed-nodes)).

Still now serializer fails to do so since it has erroneous condition on object node termType.

bourgeoa:
PR #216 fails to pass tests

bourgeoa:
@akoptelov @Fabbi done and committed with PR #471

jeswr:
Closing as part of the maintainer-signed-off triage (plan #824). This was fixed by PR #471 (merged Jan 2021), and I re-verified on v2.4.0 that the RDF/XML serializer now emits typed node elements (e.g. <ex:Person rdf:about="...">) for rdf:type statements. Please reopen with a fresh repro if you still see rdf:Description output where a typed node element is expected.

PR #471, opened 2021-01-22
RDF+XML typed node elements

bourgeoa:
@Fabbi
this is a follow up of #215 with PR #216 failling tests node 6.0


michielbdejong (approved):
Sounds good

PR #474, opened 2021-02-09
changing uncommon prefixes to common

### LINKED ISSUE #472: make rdflib use established prefixes (e.g., `xsd:` instead of `XML:`)
> Would be cool if we could make rdflib use established prefixes as well

_Originally posted by @angelo-v in https://github.com/solid/solid-panes/pull/277#issuecomment-768831562_

The pull request linked above changed --
```
@prefix    xsd: <http://www.w3.org/2001/XMLSchema#> .
``` 
-- in one file, to --
```
@prefix    XML: <http://www.w3.org/2001/XMLSchema#> .
```
-- to match a number of instances of the latter.

I suggested that the fix should have been in the other direction, but apparently the `XML:` prefix is used by rdflib, and inherited from there in many places.

My go-to reference site for prefix common use is prefix.cc, which apparently doesn't support capitalized prefixes at all (which may tell us something).  It does have a lowercase `xml:` prefix, but that doesn't match the above expansion.  `xs:` and `xsd:` both do match the above; in my experience, the latter is more commonly used.

It appears there will be multiple pull requests, in part because this issue crosses multiple repos, needed to change everything relevant from `XML:` to `xsd:`.

TallTed:
e.g., `XML:` -> `xsd: <http://www.w3.org/2001/XMLSchema#>`
e.g., `n:`   -> `org: <http://rdfs.org/sioc/ns#>`
e.g., `n0:`  -> `dc: <http://purl.org/dc/elements/1.1/>`

(partly?) fixes #472

### DIFF
--- a/tests/serialize/prefixgen.ttl
+++ b/tests/serialize/prefixgen.ttl
@@ -1,18 +1,18 @@
 
 @prefix n1: <http://csarven.ca/#>.
-@prefix n0: <http://purl.org/dc/elements/1.1/>.
+@prefix dc: <http://purl.org/dc/elements/1.1/>.
 @prefix c: <https://www.w3.org/People/Berners-Lee/card#>.
 @prefix det: <details.ttl#>.
-@prefix XML: <http://www.w3.org/2001/XMLSchema#>.
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#>.
 @prefix p: <http://www.w3.org/ns/pim/pad#>.
-@prefix n: <http://rdfs.org/sioc/ns#>.
+@prefix org: <http://rdfs.org/sioc/ns#>.
 @prefix pro: <https://deiu.me/profile#>.
 
 
 
-<>  n0:author  n1: .
+<>  dc:author  n1: .
 
-<>  n:content  "foo".
+<>  org:content  "foo".
 
 
 
--- a/tests/serialize/t1.ttl
+++ b/tests/serialize/t1.ttl
@@ -1,13 +1,13 @@
 @prefix det: <details.ttl#>.
-@prefix n0: <http://purl.org/dc/elements/1.1/>.
+@prefix dc: <http://purl.org/dc/elements/1.1/>.
 @prefix c: <https://www.w3.org/People/Berners-Lee/card#>.
-@prefix XML: <http://www.w3.org/2001/XMLSchema#>.
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#>.
 @prefix p: <http://www.w3.org/ns/pim/pad#>.
-@prefix n: <http://rdfs.org/sioc/ns#>.
+@prefix org: <http://rdfs.org/sioc/ns#>.
 
    
 <#id1443100844982>
-    n:content
+    org:content
        "kasdfjsahdkfhkjhdkjsfhjkasdfkhjkajkdsajkhadsfkhjhjkdfajsdsafhjkdfhjksa";
     p:integer
        0;     p:decimal 12.0;  p:float 3.141e0; p:date 2012-12-10; p:dateTime 2012-12-25T23:59;
--- a/tests/serialize/t12-ref.ttl
+++ b/tests/serialize/t12-ref.ttl
@@ -1,11 +1,11 @@
 @prefix : </structures.n3#>.
-@prefix XML: <http://www.w3.org/2001/XMLSchema#>.
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#>.
 
 :DSK :name "Dexter Scott King".
 
 :MLK
-    :born "1929-01-17"^^XML:date;
-    :died "1968-01-15"^^XML:date;
+    :born "1929-01-17"^^xsd:date;
+    :died "1968-01-15"^^xsd:date;
     :friend [];
     :name "Martin Luther King, Jr";
     :spouse _:_g_L4C88;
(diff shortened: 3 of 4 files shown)


tests/serialize/prefixgen.ttl
```diff
@@ -1,18 +1,18 @@
 
 @prefix n1: <http://csarven.ca/#>.
-@prefix n0: <http://purl.org/dc/elements/1.1/>.
+@prefix dc: <http://purl.org/dc/elements/1.1/>.
```
pmcb55, line 3: More a question than a suggestion, but should this be using the DC Terms namespace (http://purl.org/dc/terms/), as opposed to this DC Elements namespace? If you do just want the 15 DC Elements, then is the preferred prefix `dc11`, as suggested here: https://stackoverflow.com/questions/47519315/what-is-the-difference-between-dublin-core-terms-and-dublin-core-elements-vocabu.

That StackOverflow also quotes a recommendation to use the DC Terms namespace, referring to DC Elements as 'legacy', i.e., "**_Over time, however, implementers are encouraged to use the semantically more precise dcterms: properties, as they more fully follow emerging notions of best practice for machine-processable metadata._**".

Secondly (regardless of the first question), but since the Turtle serializations for both DC Terms and DC Elements use the prefix `dcterms` themselves, I'd suggest that that is their preferred prefix, and so would suggest using that prefix here too. (It sure would be nice if everyone actually stipulated a preference by using [vann:preferredNamespacePrefix](https://vocab.org/vann/#preferredNamespacePrefix), but what can you do!):
```suggestion
@prefix dcterms: <http://purl.org/dc/terms/>.
```

TallTed: I don't have a horse in that race.  I went to prefix.cc to get the prefixes listed there for the namespaces in these files, all of which I targeted because of their previous (well, current, until something gets merged) use of `@prefix XML: <http://www.w3.org/2001/XMLSchema#>` which just made me itch.

Maybe your much more detailed contemplation belongs in the issue (#472) rather than this PR, which I've noted only partly addresses the issue?

pmcb55: No, I think the topic in both issues is pretty focused - i.e., the topic in this issue is purely Dublin Core, with the only questions being:
1. whether to reference the DCTerms or DCElements namespace? (My vote would be: `http://purl.org/dc/terms/`).
2. what prefix to use for 'DCStuff'? (My vote would be `dcterms:`).

I totally understand your using `prefix.cc` (I keep meaning to reach out to Richard again!), but it's really pretty much an ad-hoc, for-guidance-only tool. So, in the Dublin Core case at least, I think the references in the StackOverflow [link](https://stackoverflow.com/questions/47519315/what-is-the-difference-between-dublin-core-terms-and-dublin-core-elements-vocabu) I posted above would trump Prefix.cc.

I think the XML prefix issue is very focused too, so better addressed in #472 (although my vote would be `xsd` !).

TallTed: @pmcb55 - This (@474) is a Pull Request.  The other (#472) is an Issue, which is (partially) fixed by this Pull Request.  As I created both #472 and #474, I think I can say whether either is "purely" focused on Dublin Core (neither is) or "very focused" on the XML prefix (that's what first caught my eye and brought me to create Issue #472, but it's not all I focused on in either Issue #472 or Pull Request #474).

pmcb55: Ah, right-you-are @TallTed - sorry for any confusion :( .
Anyway, I've thrown in my 2cents worth - I was only trying to offer you guys my thoughts, specifically as I've asked myself **_exactly_** these very same questions in my own work over the years. Cheers.

angelo-v:
I think the PR is fine. The switch to dcterms should probably be done in a seperate issue/PR. But there is a failing test. Can you fix this @TallTed, then we can merge it, imho

TallTed:
@angelo-v -- The failing tests appear to be comparisons between dynamically generated RDF (which now uses `xsd:`), and some static file(s) (which still use `XML:`) against which that dynamically generated RDF is being checked. Github is not revealing to me what/where the static file(s) is/are, so I cannot change them to match the new dynamically generated data.

angelo-v:
@TallTed There are not so many places left that use `XML:`, so you proably find them here: https://github.com/linkeddata/rdflib.js/search?l=Turtle&q=%22XML:%22

angelo-v:
Ah sorry, no I understood it wrong. So you actually adjusted only static ttl files so far and no turtle generation, yet. So basically you only changed the test expectations, but did not change the way rdflib generates preefixes.

TallTed:
Ah, my bad. I keep expecting Github to search more than it does. And it only alerts that "forked repositories are not currently searchable" if I select "Code" for the search. But it does find a manageable 16 hits in "Commits" on the parent branch. I'll give it another whack.

TallTed:
@angelo-v, others...

I'm at a loss. Github searches brought no joy, so I downloaded my fork, and tried grepping through it locally. I can't find anything that looks like it's associating `XML` with the `XMLSchema` URL, nor doing the other odd prefix settings. So I guess this will languish until someone gets tired of seeing this PR in the list, and goes and finds where the odd prefixes are being set.

angelo-v:
I am afraid rdflib just generates the prefixes based on the URL and does not have any configuration for common prefixes to use. I don't think it is of any help to leave this PR open, since it only adjusts test data. To keep track of the task we still have the issue https://github.com/linkeddata/rdflib.js/issues/472

Issue #485, opened 2021-03-23
parsing a html+RDFa resource containing datatype="rdf:HTML" return 500


bourgeoa:
message: 'Error translating between RDF formats',
  status: 500
```
Sarven Capadisli
@csarven
12:46
@bourgeoa

HTTP/2 500 
date: Tue, 23 Mar 2021 11:52:46 GMT
server: Apache/2.4.29 (Ubuntu)
x-powered-by: solid-server/5.6.6
vary: Accept,Authorization,Origin
access-control-allow-credentials: true
access-control-expose-headers: Authorization, User, Location, Link, Vary, Last-Modified, ETag, Accept-Patch, Accept-Post, Updates-Via, Allow, WAC-Allow, Content-Length, WWW-Authenticate, MS-Author-Via, X-Powered-By
allow: OPTIONS, HEAD, GET, PATCH, POST, PUT, DELETE
link: <linked-research-decentralised-web.acl>; rel="acl", <linked-research-decentralised-web.meta>; rel="describedBy", <http://www.w3.org/ns/ldp#Resource>; rel="type"
wac-allow: user="read",public="read"
ms-author-via: SPARQL
updates-via: wss://csarven.ca
content-type: text/plain; charset=utf-8
content-length: 38
etag: W/"26-QBI7EYtDt6zJzGYxyvtGCQrnW5Q"

Error translating between RDF formats

  solid:get /linked-research-decentralised-web on csarven.ca +12s
  solid:handlers GET -- Reading /var/www/solid/data/linked-research-decentralised-web$.html +12s
  solid:get error translating: /linked-research-decentralised-web text/html -> text/turtle -- 500 Cannot read property 'htmlOptions' of undefined +426ms
  solid:server Error page because of: [HTTPError: Error translating between RDF formats] {
  name: 'HTTPError',
  message: 'Error translating between RDF formats',
  status: 500

Sarven Capadisli
@csarven
14:12
The request was: curl -iH'Accept: text/turtle' https://csarven.ca/linked-research-decentralised-web
Alain Bourgeois
@bourgeoa
15:48

@csarven
Thanks for giving the link url.
I run the html+RDFa tests using a copy of your big file :

    test failed
    run test on part of your file : test succeed until it encounters : datatype="rdf:HTML"
    test succeed when replacing every datatype="rdf:HTML" with "rdf:html"

[...]

PR #487, opened 2021-04-01
resolve rdf:HTML in RDFa

bourgeoa:
includes 2 RDFa tests to turtle in serialize. One from solid/solid-crud-tests without rdf:HTML and the other on an abstract of @csarven thesis containing a rdf:HTML
see issue #485


bourgeoa:
Can you validate I have no access? @angelo-v @jeff-zucker

angelo-v:
LGTM ✅ Would be good to have the `domToString` unit tested as well.

Issue #490, opened 2021-04-23
updater.updateMany() may return a wrong result


bourgeoa:
https://github.com/solid/contacts-pane/issues/40

instead of grouping statements by document, the function fail to de-duplicate st.why. With as aconsequence a duplication of statements. On remove this induces PATCH failure.

`const uniqueDocs = Array.from(new Set(docs))` is incorrect because all objects are different, we must compare the content.
We can use rdflib.js `sameTerm` that make a node key value comparison in something like : 
```
const uniqueDocs = []
docs.map(doc => {
    if (!uniqueDocs.find(item => item.sameTerm(doc))) uniqueDocs.push(doc)
 })
```

PR #491, opened 2021-04-24
fix updateMany issue#490

bourgeoa:
issue #490


bourgeoa:
I also changed a bit the test so that it fails on previous `updateMany()` code where  multiple same PATCH delete occured.

timbl (approved):
Looks reasonable at a read though, and has a test.

PR #503, opened 2021-06-07
Fix fetcher interfaces export

### LINKED ISSUE #500: Cannot import library: "'AutoInitOptions' is not exported from './fetcher'."
**Repro steps:**
1. Create empty react project using CRA.
2. Install rdflib ( npm install rdflib@2.2.3 )
3. Import the library in App.js file (import * as rdf from "rdflib")
3. Start the project (npm start)

**Result:**
Project fails to run. Following error is visible:

`../shared/node_modules/rdflib/esm/index.js 61:0-63:72
Attempted import error: 'AutoInitOptions' is not exported from './fetcher'.`

**Affected versions:**
2.2.2, 2.2.3

**Additional Info**
That was working fine in version 2.1.7

ludwigschubi:
Fixes #500

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,7 +4,7 @@ import BlankNode from './blank-node'
 import Collection from './collection'
 import * as convert from './convert'
 import Empty from './empty'
-import Fetcher, { AutoInitOptions, ExtendedResponse, FetchError } from './fetcher'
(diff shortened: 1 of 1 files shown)


bourgeoa:
Thanks. OK when tested with building mashlib

PR #508, opened 2021-07-16
read json-ld

jeff-zucker:
Fixes https://github.com/solid/solidos/issues/37.


jeff-zucker:
Replaced by better fix in #509

PR #509, opened 2021-07-16
Jsonld

timbl:
Just changed coed to be more logical, hopefully to fix a Store being returned instead of Response.

### DIFF
--- a/src/fetcher.ts
+++ b/src/fetcher.ts
@@ -500,7 +500,9 @@ class JsonLdHandler extends Handler {
     const kb = fetcher.store
     return new Promise((resolve, reject) => {
       try {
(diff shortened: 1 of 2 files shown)


bourgeoa:
Shouldn't it be `this.response`

bourgeoa:
@timbl or @SharonStrats
before publishing the build works but there is a typescript issue
> src/fetcher.ts:505:19 - error TS2345: Argument of type 'Response' is not assignable to parameter of type 'FetchError | PromiseLike<FetchError>'.
  Type 'Response' is missing the following properties from type 'FetchError': name, message
>
> 505           resolve(fetcher.doneFetch(options, response))
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

PR #511, opened 2021-08-10
Fixed typing error

jaxoncreed:


bourgeoa (commented):
Not sure this change is correct.
You cannot parse json or jsonld if data is not a valid json. A minimal string is "{}".

jaxoncreed:
At least as far as the typings are concerned, `options.data` is `string | undefined`, but `jsonString` is `string | undefined | null`. All this does is handle the case where `jsonString` is null.

Issue #441, opened 2020-09-27
Cached WAC-allow headers and 403 and 401 responses must be ignored after login


timbl:
Currently the quad store in the Fetcher keeps copies of the HTTP headers of each response, and uses that to test whether something is forbidden (403), and is if modifiable by the current user (WAC-Alllow).  The problem is that if the app accesses a resource, gets those headers, and then the user logs in, the cached values will be wrong: the operation should be retried to get a value appropriate to the now-logged-in user.

So these functions need some way of testing the date-time of the user's last login status change.

An example problem causes by this is:  https://github.com/solid/chat-pane/issues/48

timbl:
I guess they should be ignored after logout too.

timbl:
401 errors also should be ignored if the predate a login status change.

timbl:
Suggest: Add a date to the fetcher when the login state changes. Like` fetcher.loginTime = new Date() `or something, and then extract the datetime of the responses being considered in the cache, and ignore cachshed 403 and 402 responses which predate the loginTime.

timbl:
This issue may be the underlying issue to https://github.com/solid/solidos/issues/29

timbl:
From gitter: @jeff-zucker You made a good point in the meeting that, with the Wac-Allow header, we have to be wary of even responses like 200 and 404 which are not to do with authentication errors, as the WAC-Allow status my have changed for 200 (or 404) responses. Its not just 403 and 401 which need to be disregarded.

angelo-v:
Repopen, since the PR was only about the editable check

bourgeoa:
https://github.com/SolidOS/solid-ui/pull/536#pullrequestreview-1353362763

angelo-v:
With #512 merged the editable check now is working fine. But the fetcher still caches the 401 and 403 responses and will not re-fetch a resource after login unless it is forced. I guess the fetcher should consider the new outOfDate flag when deciding wether to fetch a resource or take it from cache.

PR #512, opened 2021-08-16
Flush metadata

timbl:
This is designed to address https://github.com/linkeddata/rdflib.js/issues/441 along with other code to call updater.flagAuthorizationMetadata when the login state changes.


src/update-manager.ts
```diff
@@ -123,15 +161,22 @@ export default class UpdateManager {
...
-        var response = kb.any(request, this.ns.link('response')) as Quad_Subject
-        if (request !== undefined) {
-          var wacAllow = kb.anyValue(response, this.ns.httph('wac-allow'))
+        const response = kb.any(request, this.ns.link('response'), null, meta) as Quad_Subject
+        if (response !== undefined) { // ts
+
+          const outOfDate = kb.anyJS(response, this.ns.link('outOfDate'), null, meta)
+          if (outOfDate) continue
+
+          var wacAllow = kb.anyValue(response, this.ns.httph('wac-allow'), null, meta)
```
SharonStrats, line 179: One thing I was having trouble grasping.. when it sees that is is outOfDate is there something it needs to do to "refresh" and become up to date? It's probably not the case but I feel like maybe we are assuming that if it's out of date that it is okay to continue.  I think I'm missing some key thing that happens when we login.

tests/unit/update-manager-test.js
```diff
@@ -198,4 +197,53 @@ describe('UpdateManager', () => {
...
+
+    it('Async version should detect a document is editable from metadata', () => {
+      loadMeta(updater.store)
+      updater.checkEditable(doc1).then(
+        res => expect(result).to.equal('SPARQL')
+      )
+      expect(updater.editable(doc1)).to.equal('SPARQL')
+    })
+
+    it('Async version should not detect a document is editable from metadata after flush', () => {
```
theRealImy, line 235: I initially thought the logic is the other way around: if the triples are flagged as outdated it means the doc is editable.

src/update-manager.ts
timbl: Yes and no.  At that level, the editable function is synch so it can't replace the out of date data as it can't access the web.  But any ftehc which has been done since logic will noot be out of date.  The fomrs for example tend to fetch the data when they display a form.  But alos in future those who can use anyc functions can use checkEditable which first tries using existing data and if it find nothing which is not out of date, does a fetcher.load() of the thing, which will add new not out of data metadata, so the second call to editable() should work.

tests/unit/update-manager-test.js
timbl: No,  many docs out there on many web servers are not editable.  So the default has to be not editable.  We have to find evidence from interaction with the document that it is in fact editable by the user in their current state.

tests/unit/update-manager-test.js
```diff
@@ -198,4 +197,53 @@ describe('UpdateManager', () => {
...
+      loadMeta(updater.store)
+      updater.flagAuthorizationMetadata()
+      expect(updater.editable(doc1)).to.equal(undefined)
+    })
+
+
+    it('Async version should detect a document is editable from metadata', () => {
+      loadMeta(updater.store)
+      updater.checkEditable(doc1).then(
+        res => expect(result).to.equal('SPARQL')
```
angelo-v, line 230: The test will not wait for the promise to resolve and this expect will not be executed. You either need to `await` or pass in the `done` method to the test function and call it when all expects are done. (see)https://jestjs.io/docs/asynchronous

src/fetcher.ts
```diff
@@ -761,8 +761,9 @@ export default class Fetcher implements CallbackifyInterface {
     if (!this._fetch) {
       throw new Error('No _fetch function available for Fetcher')
     }
-
-    this.appNode = this.store.rdfFactory.blankNode()
+    // This is the name of the graphh we store all the HTTP metadata in
+    this.appNode = this.store.sym('chrome://TheCurrentSession')
```
angelo-v, line 765: why chrome://?

src/update-manager.ts
```diff
@@ -91,14 +91,52 @@ export default class UpdateManager {
     return( uri.slice(0,4) === 'http' )
   }
 
+/** Remove from the store HTTP authorization metadata
+* The editble function below relies on copies we have in the store
+* of the results of previous HTTP transactions. Howver, when
+* the user logs in, then that data misrepresents what would happen
+* if the user tried again.
+*/
+async flagAuthorizationMetadata () {
```
angelo-v, line 100: Does this have to be called manually to trigger an invalidation where needed or is it called somewhere from rdflib itself?

bourgeoa:
@timbl can you review Angelo comments?

src/update-manager.ts
timbl: It has to be called by the app when the user logs in.  Or, I guess from the login code.  But rdflib itself does not have a spot in the login flow.

bourgeoa:
After merging main, build succeed, but new tests fail : 
```
update callback uri = undefined, ok = false, text = <<<Exception in update: Error: Update: Loaded <https://example.com/test/foo>but stil can't figure out what editing protcol it supports.
Error: Update: Loaded <https://example.com/test/foo>but stil can't figure out what editing protcol it supports.
  undefined
    Error: Update: Loaded <https://example.com/test/foo>but stil can't figure out what editing protcol it supports.  undefined
        at UpdateManager.update (/mnt/d/github/source/workspaces/rdflib/src/update-manager.ts:832:17)  undefined
        at /mnt/d/github/source/workspaces/rdflib/src/update-manager.ts:836:16  undefined
        at processTicksAndRejections (node:internal/process/task_queues:96:5)>>>
```

angelo-v:
The failing tests are

 - live-store-test.js
 - update-manager-sparql-insert-test.ts

src/update-manager.ts
```diff
@@ -113,15 +151,22 @@ export default class UpdateManager {
...
-    var requests = kb.each(undefined, this.ns.link('requestedURI'), docpart(uri))
+    var requests = kb.each(undefined, this.ns.link('requestedURI'), docpart(uri), meta)
     var method: string
     for (var r = 0; r < requests.length; r++) {
       request = requests[r]
       if (request !== undefined) {
-        var response = kb.any(request, this.ns.link('response')) as Quad_Subject
-        if (request !== undefined) {
-          var wacAllow = kb.anyValue(response, this.ns.httph('wac-allow'))
+        const response = kb.any(request, this.ns.link('response'), null, meta) as Quad_Subject
```
angelo-v, line 163: @timbl now that we are looking up the response meta data in the `meta` graph we need to make sure that the fetcher stores it there as well. Currently the fetcher `saveResponseMetadata` uses a blank node and I think it needs to use `this.appNode` instead. This is the reason why the tests in `update-manager-sparql-insert-test.ts` are currently failing

tests/unit/update-manager-test.js
```diff
@@ -198,4 +197,53 @@ describe('UpdateManager', () => {
...
+    const self = {err: ''}
+    let updater, docuri, rterm, options, userCallback, loadStub
+    var loadStatus = 200
+
+    beforeEach(() => {
+      options = {}
+      userCallback = () => {}
+
+      updater = new UpdateManager()
+      // updater.store.fetcher.webOperation = sinon.stub().resolves({ ok: true, status: 200, statusText: "Dummy stub 1"})
```
angelo-v, line 210: This needs to be included for the last test to work, since the last test will fetch data

tests/unit/update-manager-test.js
angelo-v: Let's replace it with async/await as discussed

tests/unit/update-manager-test.js
```diff
@@ -198,4 +197,53 @@ describe('UpdateManager', () => {
...
+        res => expect(result).to.equal('SPARQL')
+      )
+      expect(updater.editable(doc1)).to.equal('SPARQL')
+    })
+
+    it('Async version should not detect a document is editable from metadata after flush', () => {
+      loadMeta(updater.store)
+      updater.flagAuthorizationMetadata()
+      updater.checkEditable(doc1).then(
+        res => expect(result).to.equal('SPARQL')
```
angelo-v, line 239: I think this for this to work we need to 1) include an accept-patch header in the (currently commented-out) mocks in the `beforeEach` and 2) make sure the fetcher stores the meta in the correct graph (as mentioned also in my other comment at https://github.com/linkeddata/rdflib.js/pull/512/files#r1137660733)

tests/unit/update-manager-test.js
angelo-v:
```
    it('Async version should detect a document is editable from metadata', async () => {
      loadMeta(updater.store)
      expect(updater.editable(doc1)).to.equal('SPARQL')
      const res = await updater.checkEditable(doc1)
      expect(res).to.equal('SPARQL')
    })

    it('Async version should not detect a document is editable from metadata after flush', async () => {
      loadMeta(updater.store)
      await updater.flagAuthorizationMetadata()
      expect(updater.editable(doc1)).to.equal(undefined)

      const result = await updater.checkEditable(doc1)
      expect(result).to.equal('SPARQL')
    })
```

tests/unit/update-manager-sparql-insert-test.ts
```diff
@@ -1,4 +1,4 @@
-import * as chai from 'chai'
+npimport * as chai from 'chai'
```
angelo-v: typo

src/update-manager.ts
angelo-v: @bourgeoa why did you resolve this? If my assumption is wrong, I would like to understand why.

tests/unit/update-manager-test.js
```diff
@@ -198,4 +197,55 @@ describe('UpdateManager', () => {
...
+    it('Async version should not detect a document is editable from metadata after flush', async () => {
+      loadMeta(updater.store)
+
+      updater.flagAuthorizationMetadata()
+
+      expect(updater.editable(doc1)).to.equal(undefined)
+
+      const result = await updater.checkEditable(doc1)
+
+      expect(result).to.equal(undefined)
```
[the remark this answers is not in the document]
bourgeoa: What is the purpose of `outOfDate`. It appears to be either `undefined` or `true`

src/update-manager.ts
angelo-v: ok. But this still puzzles me. The line will find `any` response. There might be an `outOfDate` one and a current one in the store. It might just pick the `outOfDate` one and `editable` will return undefined then, despite having a more up to date response that says it is editable via SPARQL. I guess we need to find any response that is not out of date.

tests/unit/update-manager-test.js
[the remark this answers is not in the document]
angelo-v: The `outOfDate` is used to flag a response as outdated, so that the editable function does not return out of date information about whether a response is editable or not. E.g. if I fetched a resource before log in it was not editable. When I log in and fetch the resource again or call the async checkEditable() function it should respect the current information about editability.

Therefore I would like to see a proof (aka a test) that this is working, like that:

- given preloaded meta data says a resource is editable
- when I call editable() it says `SPARQL` (because of the metadata in the store)
- when I flagAuthorizationMetadata and call editable it says `undefined` (because of the `outOfDate` flag)
- when I call checkEditable it says (`SPARQL`) again (because current data is fetched and respected)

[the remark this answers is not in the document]
bourgeoa: > when I call checkEditable it says (SPARQL) again (because current data is fetched and respected)

This is possible by removing the flag in `checkEditable().`

```
   const meta = kb.fetcher.appNode
   const requests = kb.statementsMatching(undefined, this.ns.link('requestedURI'), kb.literal(uri), meta).map(st => st.subject)
   for (const request of requests) {
    // remove outOfDate metadata flag
    const response = kb.any(request, this.ns.link('response'), null, meta) as Quad_Subject
    if (response !== undefined) {
      const quad = kb.statementsMatching(response, this.ns.link('outOfDate'), true as any, meta)
      kb.remove(quad)
      
    }

```
Is it the expected behaviour without removing/reloading the document and the metadata ?
If `checkEditable(uri)` removes/reloads the document and the metadata => the test will need a fake `kb.fetcher.load(uri)`

I'm wondering what are the purpose of 
- `flagAuthorizationMetadata()`
- `checkEditable(uri)`

What is needed at login/logout ? is `flagAuthorizationMetaData()` enough ? shall a new store be reinitiated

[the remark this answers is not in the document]
angelo-v: We could at least delete all the request/response meta data, not necessarily the whole store. I do not know why @timbl decided to go with a flag instead. We can discuss it in the call

src/fetcher.ts
angelo-v: From the call: Comes not from the browser name, but used internally in browsers (predates the Browser Chrome)

src/update-manager.ts
```diff
@@ -81,14 +81,52 @@ export default class UpdateManager {
...
+ *
+ * @returns The method string SPARQL or DAV or
+ *   LOCALFILE or false if known, undefined if not known.
+ */
+ async checkEditable (uri: string | NamedNode, kb?: IndexedFormula): Promise<string | boolean | undefined> {
+   const initial = this.editable(uri, kb)
+   if (initial !==  undefined) {
+     return initial
+   }
+   await this.store.fetcher.load(uri)
```
angelo-v, line 118: From the call: Should be forced by options `{ force: true}`

src/update-manager.ts
```diff
@@ -113,14 +151,21 @@ export default class UpdateManager {
 
     var request
     var definitive = false
+    const meta = this.store.fetcher.appNode
+    // const kb = s
+
      // @ts-ignore passes a string to kb.each, which expects a term. Should this work?
-    var requests = kb.each(undefined, this.ns.link('requestedURI'), docpart(uri))
+    var requests = kb.each(undefined, this.ns.link('requestedURI'), docpart(uri), meta)
```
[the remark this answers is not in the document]
angelo-v: From the call: yes, it should be there, has been a bug to leave it out

src/update-manager.ts
angelo-v: From the call: all good, we are looping through the requests!

tests/unit/update-manager-test.js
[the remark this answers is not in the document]
angelo-v: From the call: The out of date flag is used, to keep the data in case we need it (e.g. debugging)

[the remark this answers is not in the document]
angelo-v: From the call: Since we mocked load, nobody adds the meta data, so the expecation is fine as it is

angelo-v (approved):

PR #513, opened 2021-09-14
Produce the complete set of .d.ts files

pduchesne:
The current TS types are incomplete, as they do not contain types for the js files. This causes TS projects with "noImplicitAny=true" to fail.

This PR does 2 things : 
 - have the build:types script include the js files
 - do the minimal set of fixes in those files to have the type checker succeed in producing type files

Ideally the whole set of js files should be migrated to ts; this PR does not go that way and is a minimal effort to have complete types.


bourgeoa:
@Vinnl @ludwigschubi Are there any reason not to merge this PR to finally have rdflib with TS support.

Vinnl:
I don't know, I'm not an rdflib maintainer so I wouldn't block it :)

bourgeoa:
@angelo-v I suppose the npm-publish-build failed is not an issue here.

angelo-v:
@bourgeoa it's an unrelated problem. Can you check the token @timbl ?

angelo-v (approved):

ludwigschubi:
Just tested it locally, works 👍

timbl (approved):
To the extent that a non-TS expert can review this isseu, this seems like a good idea to me.

Issue #518, opened 2021-10-16
bug in statementsMatching function in formula.ts


bblfish:
There is a bug here in rdflib.js [statementsMatching](https://github.com/linkeddata/rdflib.js/blob/062cc104da7b6abeac928be68b453e54b2d0c61a/src/formula.ts#L250) implementation in formula.ts. The last line below compares the graph to the subject. 
```scala
statementsMatching<JustOne extends boolean = false>(
    s?: Quad_Subject | null,
    p?: Quad_Predicate | null,
    o?: Quad_Object | null,
    g?: Quad_Graph | null,
    justOne?: boolean
  ): Statement[] {
    const sts = this.statements.filter(st =>
      (!s || s.equals(st.subject)) &&
      (!p || p.equals(st.predicate)) &&
      (!o || o.equals(st.object)) &&
      (!g || g.equals(st.subject))
    )
```

jeswr:
As part of the maintainer-signed-off triage pass (plan #824; agent-assisted for @jeswr): the bug reported here — `Formula.statementsMatching` in formula.ts comparing the graph term against `st.subject` instead of `st.graph` — was real, and it was fixed the day after this issue was filed by PR #519 (commit c1bd4441, "bug in statementsMatching()", which references this issue; @bourgeoa's comment about adding a test was referring to that PR). I verified both the v2.4.0 release source and current master: the filter now reads `(!g || g.equals(st.graph))`. Closing as fixed; please reopen if you can show a graph-filtered `statementsMatching` mismatch on 2.4.x.

PR #519, opened 2021-10-17
bug in statementsMatching()

bourgeoa:
issue #518


src/formula.ts
```diff
@@ -258,7 +258,7 @@ export default class Formula extends Node {
       (!s || s.equals(st.subject)) &&
       (!p || p.equals(st.predicate)) &&
       (!o || o.equals(st.object)) &&
-      (!g || g.equals(st.subject))
+      (!g || g.equals(st.graph))
```
angelo-v, line 261: would be a great opportunity to add a unit test :)

angelo-v (approved):

PR #521, opened 2021-10-17
Create ci.yml

bourgeoa:


.github/workflow/ci.yml
```diff
@@ -0,0 +1,88 @@
...
+          - 15.x
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: npm ci
+      - run: npm test
+      - run: npm run doc
```
angelo-v, line 28: This command does not work locally for me as well. We can leave it out for now and fix it seperately.

.github/workflow/ci.yml
```diff
@@ -0,0 +1,88 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version:
+          - 12.x
+          - 14.x
+          - 15.x
```
angelo-v, line 19: let's use the current version 16 instead of 15

.github/workflow/ci.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          node-version: 14.x
+      - uses: rlespinasse/github-slug-action@v3.x
+      - name: Append commit hash to package version
+        run: 'sed -i -E "s/(\"version\": *\"[^\"]+)/\1-${GITHUB_SHA_SHORT}/" package.json'
+      - name: Disable pre- and post-publish actions
+        run: 'sed -i -E "s/\"((pre|post)publish)/\"ignore:\1/" package.json'
+      - uses: JS-DevTools/npm-publish@v1
+        with:
+          token: ${{ secrets.NPM_TOKEN }}
```
angelo-v, line 70: NPM_TOKEN has to be placed in the repository settings, if not done yet

angelo-v (commented):
Thanks @bourgeoa for working on that. I find it really useful especially since I wondered who will when publish a new version after my last bugfix :)

angelo-v:
> Documentation are not the same link in solid/ui and rdflib.js but looking at travis I suppose it is working

In Travis we only published `dist`, `doc` and `Documentation` if I see it correctly, now we will publish the whole repo root to gh-pages. It's not optimal, but ok, and the links should be the same in the and.

> 
> Do you think automatic publish is correct, or needed

looks find. It is definetly a big win in my opinion.

.github/workflow/ci.yml
angelo-v: ok, if we leave it out we will loose content on http://linkeddata.github.io/rdflib.js/doc/. I will try to see what is wrong with the command

angelo-v: fixed in https://github.com/linkeddata/rdflib.js/pull/522

.github/workflow/ci.yml
bourgeoa: To add the NPM_TOKEN I need access to settings in rdflib.js @timbl can you do something.

PR #527, opened 2021-10-29
This fixes #526 by recursively parsing all nested `@graph` arrays.

### LINKED ISSUE #526: rdflib.js does not support JSON-LD with named graphs
I have JSON-LD files with (nested) named graphs.

Looking at the JSON-LD parser here:

https://github.com/linkeddata/rdflib.js/blob/23386aeb2d3421cd40e39067972b0f8890517a96/src/jsonldparser.js#L68-L98

It appears that it only asserts quads in the "base" graph, and ignores any triples in a nested `@graph` key for the triple.

A flattened structure for a triple-in-a-graph will look like this:

```
{
      "@graph": [
        {
          "@id": "https://example.com/someSubject",
          "https://example.com/somePredicate": {
            "@language": "en",
            "@value": "some literal value"
          }
        }
      ],
      "@id": "https://example.com/someGraphIdentifier"
}
```

This will attempt to create a new property in the store from the "uri" `@graph`, with a value that is the array containing the one triple... and this fails silently.

RinkeHoekstra:
Added a `processResource` function that determines whether the resource is a triple statement, or a graph. 

If it's a graph, it will use the graph's identifier as the new base for any statements contained in the graph's array.

These statements are processed by recursively calling the `processResource` function again.

### DIFF
--- a/src/jsonldparser.js
+++ b/src/jsonldparser.js
@@ -73,28 +73,48 @@ export default function jsonldParser (str, kb, base, callback) {
   return jsonld
     .flatten(JSON.parse(str), null, { base: baseString })
     .then((flattened) => flattened.reduce((store, flatResource) => {
-      const id = flatResource['@id']
-        ? kb.rdfFactory.namedNode(flatResource['@id'])
-        : kb.rdfFactory.blankNode()
 
-      for (const property of Object.keys(flatResource)) {
-        if (property === '@id') {
-          continue
-        }
-        const value = flatResource[property]
-        if (Array.isArray(value)) {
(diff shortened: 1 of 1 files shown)


RinkeHoekstra:
Thanks @bourgeoa for your quick reaction! I have only touched the parser code, not the serialization. How is this related to `test:serialize:9`?

bourgeoa:
Sorry not being an expert.
Seeing the use cases it seems to be the way to compare data output with input. So it is an integrated test to check that parser and serialisation are working together correctly. Actually we can see there are no regression, but not that all inputs from your code can be serialised.

If you have better/other way please do.

PR #530, opened 2021-11-22
user fetch renamed to solidFetch

bourgeoa:


jeff-zucker (approved):
Looks great, thank you!

Issue #531, opened 2021-11-27
How-to add a user-defined fetch()


bourgeoa:
There is no documentation on how-to replace the default fetch() by a user defined one. This a frequent developer question.

jeff-zucker:
Excellent point.

PR #532, opened 2021-11-27
Create alternate-fetches.md

jeff-zucker:
Documentation for authenticated and alternate fetch methods. This should resolve https://github.com/linkeddata/rdflib.js/issues/531.

### DIFF
--- a/Documentation/alternate-fetches.md
+++ b/Documentation/alternate-fetches.md
@@ -29,7 +29,7 @@ async function test() {
 
 ### A Browser Example
 
-This is a fully functinal script, just change the IDP and privateResource addresses.
+This is a fully functional script; just change the `IdP` and `privateResource` addresses.
 
 ```html
 <!DOCTYPE html><html><head><meta charset="UTF-8" />                             
--- a/Documentation/alternate-fetches.md
+++ b/Documentation/alternate-fetches.md
@@ -73,7 +73,7 @@ r@1.11.2/dist/solid-client-authn.bundle.js"></script>
 
 ##  Custom fetchers
 
-Another way to specify an authenticated or alternate fetch is to do it when you create the fetcher.  This method follows the same steps as the first method with two exceptions : omit the line *global.solidFetch = auth.fetch* and instead of the line *const fetcher = $rdf.fetcher(kb)* use :
+Another way to specify an authenticated or alternate fetch is to do it when you create the fetcher.  This method follows the same steps as the first method, with two exceptions: omit the line *`global.solidFetch = auth.fetch`*; and instead of the line *`const fetcher = $rdf.fetcher(kb)`* use :
 ```javascript
 const fetcher = $rdf.fetcher(kb,{fetch:auth.fetch.bind(auth)});
 ```
--- a/Documentation/alternate-fetches.md
+++ b/Documentation/alternate-fetches.md
@@ -4,7 +4,7 @@ By default, `rdflib`'s fetcher uses `cross-fetch.fetch()` to preform reads and w
 
 ## The solidFetch variable                                                
 
-To to use an authenticated or alternate fetch with `rdflib`
+To use an authenticated or alternate fetch with `rdflib`
 
 1. Load the authentication library.
 2. Use it to log in to your identity provider.
--- a/Documentation/alternate-fetches.md
+++ b/Documentation/alternate-fetches.md
@@ -4,7 +4,7 @@ By default, `rdflib`'s fetcher uses `cross-fetch.fetch()` to preform reads and w
 
 ## The solidFetch variable                                                
 
-To to use an authenticated or alternate fetch with rdflib
+To to use an authenticated or alternate fetch with `rdflib`
 
 1. Load the authentication library.
 2. Use it to log in to your identity provider.
--- a/Documentation/alternate-fetches.md
+++ b/Documentation/alternate-fetches.md
@@ -1,6 +1,6 @@
 # Using authenticated and alternate fetches in rdflib
 
-By default, rdflib's fetcher uses cross-fetch.fetch() to preform reads and writes.  This is a plain fetch similar to wget or curl that does not carry any authentication information.  In order to use rdflib with private Solid data that requires authentication, the user or app needs to pass a different fetch method to rdflib.  This approach is also useful if you have created your own fetch, for example against Dropbox or a database.                                                                       
+By default, `rdflib`'s fetcher uses `cross-fetch.fetch()` to preform reads and writes.  This is a plain `fetch`, similar to `wget` or `curl`, that does not carry any authentication information.  To use `rdflib` with private Solid data that requires authentication, the user or app needs to pass a different fetch method to `rdflib`.  This approach is also useful if you have created your own fetch, such as against Dropbox or a database.                                                                       
 
 ## The solidFetch variable                                                
 
--- a/Documentation/alternate-fetches.md
+++ b/Documentation/alternate-fetches.md
@@ -11,7 +11,7 @@ To to use an authenticated or alternate fetch with rdflib
 3. Set the `global.solidFetch` (or `window.solidFetch` in a browser) equal to the auth library's fetch method.
 4. Load and use `rdflib`.  All fetches will be authenticated.
 
-**Important Note :** prior to rdflib version 2.2.9 this variable was named solidFetcher so use that with older rdflibs.  Going forward, please use solidFetch.
+_**Important Note:** Prior to `rdflib` version `2.2.9`, this variable was named `solidFetcher`, so use that with older `rdflib`s.  Going forward, please use `solidFetch`._
 
 ### A CLI (out-of-browser) Example
 
--- a/Documentation/alternate-fetches.md
+++ b/Documentation/alternate-fetches.md
@@ -81,5 +81,5 @@ const fetcher = $rdf.fetcher(kb,{fetch:auth.fetch.bind(auth)});
 ## Currently available libraries with authenticated fetch methods
 
 * [Inrupt's Solid-Client-Authn-Browser](https://github.com/inrupt/solid-client-js) for use in browsers
-* [Inrupt's Solid-Client-Authn-Node](https://github.com/inrupt/solid-client-js) for use in node when you don't need full access to the local file system
-* [Solid-Node-Client](https://github.com/solid/solid-node-client) for use in node when you need full access to the local file system
+* [Inrupt's Solid-Client-Authn-Node](https://github.com/inrupt/solid-client-js) for use in `node`, when you don't need full access to the local file system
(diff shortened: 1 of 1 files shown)


Documentation/alternate-fetches.md
```diff
@@ -0,0 +1,85 @@
+# Using authenticated and alternate fetches in rdflib
+
+By default, rdflib's fetcher uses cross-fetch.fetch() to preform reads and writes.  This is a plain fetch similar to wget or curl that does not carry any authentication information.  In order to use rdflib with private Solid data that requires authentication, the user or app needs to pass a different fetch method to rdflib.  This approach is also useful if you have created your own fetch, for example against Dropbox or a database.
```
TallTed, line 3:
```suggestion
By default, `rdflib`'s fetcher uses `cross-fetch.fetch()` to preform reads and writes.  This is a plain `fetch`, similar to `wget` or `curl`, that does not carry any authentication information.  To use `rdflib` with private Solid data that requires authentication, the user or app needs to pass a different fetch method to `rdflib`.  This approach is also useful if you have created your own fetch, such as against Dropbox or a database.                                                                       
```

Documentation/alternate-fetches.md
```diff
@@ -0,0 +1,85 @@
+# Using authenticated and alternate fetches in rdflib
+
+By default, rdflib's fetcher uses cross-fetch.fetch() to preform reads and writes.  This is a plain fetch similar to wget or curl that does not carry any authentication information.  In order to use rdflib with private Solid data that requires authentication, the user or app needs to pass a different fetch method to rdflib.  This approach is also useful if you have created your own fetch, for example against Dropbox or a database.                                                                       
+
+## The solidFetch variable                                                
+
+To to use an authenticated or alternate fetch with rdflib
```
TallTed, line 7:
```suggestion
To to use an authenticated or alternate fetch with `rdflib`
```

Documentation/alternate-fetches.md
```diff
@@ -0,0 +1,85 @@
...
+
+1. load the authentication library
+2. use it to login to your identity provider
+3. set the global.solidFetch (or window.solidFetch in a browser) equal to the auth library's fetch method
+4. load and use rdflib, all fetches will be authenticated
```
TallTed, line 12:
```suggestion
2. Use it to log in to your identity provider.
3. Set the `global.solidFetch` (or `window.solidFetch` in a browser) equal to the auth library's fetch method.
4. Load and use `rdflib`.  All fetches will be authenticated.
```

Documentation/alternate-fetches.md
```diff
@@ -0,0 +1,85 @@
+# Using authenticated and alternate fetches in rdflib
+
+By default, rdflib's fetcher uses cross-fetch.fetch() to preform reads and writes.  This is a plain fetch similar to wget or curl that does not carry any authentication information.  In order to use rdflib with private Solid data that requires authentication, the user or app needs to pass a different fetch method to rdflib.  This approach is also useful if you have created your own fetch, for example against Dropbox or a database.                                                                       
+
+## The solidFetch variable                                                
+
+To to use an authenticated or alternate fetch with rdflib
+
+1. load the authentication library
```
TallTed, line 9:
```suggestion
1. Load the authentication library.
```

Documentation/alternate-fetches.md
```diff
@@ -0,0 +1,85 @@
...
+2. use it to login to your identity provider
+3. set the global.solidFetch (or window.solidFetch in a browser) equal to the auth library's fetch method
+4. load and use rdflib, all fetches will be authenticated
+
+**Important Note :** prior to rdflib version 2.2.9 this variable was named solidFetcher so use that with older rdflibs.  Going forward, please use solidFetch.
```
TallTed, line 14:
```suggestion
_**Important Note:** Prior to `rdflib` version `2.2.9`, this variable was named `solidFetcher`, so use that with older `rdflib`s.  Going forward, please use `solidFetch`._
```

Documentation/alternate-fetches.md
```diff
@@ -0,0 +1,85 @@
...
+    const $rdf = global.$rdf = require('rdflib');                     
+    const kb = $rdf.graph(); 
+    const fetcher = $rdf.fetcher(kb);
+    await fetcher.load( some-private-url );  
+}
+```               
+
+### A Browser Example
+
+This is a fully functinal script, just change the IDP and privateResource addresses.
```
TallTed, line 32:
```suggestion
This is a fully functional script; just change the `IdP` and `privateResource` addresses.
```

Documentation/alternate-fetches.md
```diff
@@ -0,0 +1,85 @@
...
+        if (session.info.isLoggedIn)  main(session);                            
+    }                                                                           
+    handleRedirectAfterLogin();                                                 
+</script></html>                                                                
+```
+
+
+##  Custom fetchers
+
+Another way to specify an authenticated or alternate fetch is to do it when you create the fetcher.  This method follows the same steps as the first method with two exceptions : omit the line *global.solidFetch = auth.fetch* and instead of the line *const fetcher = $rdf.fetcher(kb)* use :
```
TallTed, line 76:
```suggestion
Another way to specify an authenticated or alternate fetch is to do it when you create the fetcher.  This method follows the same steps as the first method, with two exceptions: omit the line *`global.solidFetch = auth.fetch`*; and instead of the line *`const fetcher = $rdf.fetcher(kb)`* use :
```

Documentation/alternate-fetches.md
```diff
@@ -0,0 +1,85 @@
...
+```javascript
+const fetcher = $rdf.fetcher(kb,{fetch:auth.fetch.bind(auth)});
+```
+
+## Currently available libraries with authenticated fetch methods
+
+* [Inrupt's Solid-Client-Authn-Browser](https://github.com/inrupt/solid-client-js) for use in browsers
+* [Inrupt's Solid-Client-Authn-Node](https://github.com/inrupt/solid-client-js) for use in node when you don't need full access to the local file system
+* [Solid-Node-Client](https://github.com/solid/solid-node-client) for use in node when you need full access to the local file system
```
TallTed, line 85:
```suggestion
* [Inrupt's Solid-Client-Authn-Node](https://github.com/inrupt/solid-client-js) for use in `node`, when you don't need full access to the local file system
* [Solid-Node-Client](https://github.com/solid/solid-node-client) for use in `node`, when you need full access to the local file system
```

TallTed (changes requested):
Markup, grammar, etc., for clarity

Documentation/alternate-fetches.md
```diff
@@ -0,0 +1,85 @@
+# Using authenticated and alternate fetches in rdflib
+
+By default, `rdflib`'s fetcher uses `cross-fetch.fetch()` to preform reads and writes.  This is a plain `fetch`, similar to `wget` or `curl`, that does not carry any authentication information.  To use `rdflib` with private Solid data that requires authentication, the user or app needs to pass a different fetch method to `rdflib`.  This approach is also useful if you have created your own fetch, such as against Dropbox or a database.                                                                       
+
+## The solidFetch variable                                                
+
+To to use an authenticated or alternate fetch with `rdflib`
```
bourgeoa, line 7:
```suggestion
To use an authenticated or alternate fetch with `rdflib`
```

bourgeoa:
I was with the idea that best practice was to avoid using global in javascript.
Why not indicate directly in 2. that you could either use global/window or $rdf.fetcher ?

jeff-zucker:
> I was with the idea that best practice was to avoid using global in javascript. Why not indicate directly in 3. that you could either use global/window or $rdf.fetcher ?

Yes, that's probably a good idea.  I don't know how to make that change without losing the changes you and Ted have made. If I edit the original, it doesn't have any of your changes.

bourgeoa:
@jeff-zucker 
- On your branch `jeff-zucker-alternate-fetches` you should `git pull` locally. 
- or something like `git pull origin jeff-zucker-alternate-fetches`

Then make the new changes

Documentation/alternate-fetches.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+
+##  Custom fetchers
+
+Another way to specify an authenticated or alternate fetch is to do it when you create the fetcher object.  This method follows the same steps as the first method, with two exceptions: omit the line *`global.solidFetch = auth.fetch`*; and instead of the line *`const fetcher = $rdf.fetcher(kb)`* use :
+```javascript
+const fetcher = $rdf.fetcher(kb,{fetch:auth.fetch.bind(auth)});
+```
+Doing things this way avoids using global variables, but also means that the changes to fetch need to be added every time you create a new fetcher object.
```
bourgeoa, line 81: Is this true ?
My understanding was that the fetch was installed as $rdf.fetch so was kept if you do not create a new rdf object. It is the way it is used in SolidOS. You can create new store $rdf.graph() or $rdf.fetcher() and keep the same $rdf.fetch.

jeff-zucker: @bourgeoa - I believe you are wrong.  I tested and could not make it work.  If you create a fetcher with fetcher = $rdf.fetcher(kb) without options it will be an unauthenticted fetch even if you previously created a fetcher that was authenticated.  If you have a test case where the authentication is preserved across creation of fetchers, please post it.

jeff-zucker: This succeeds on the first fetch and fails 401 on the second : 
```
const auth = new (require("solid-node-client").SolidNodeClient)();              
const $rdf = global.$rdf = require('rdflib');                                   
const privateResource = "https://jeff-zucker.solidcommunity.net/private/";      
async function test() {                                                         
    await auth.login();                                                         
    const kb = $rdf.graph();                                                    
    const fetcher1 = $rdf.fetcher(kb,{fetch:auth.fetch.bind(auth)});            
    try {                                                                       
       await fetcher1.load(privateResource);                                    
       console.log("Private resource successfully loaded");                     
     }                                                                          
    catch(e) { console.log(e) }                                                 
    const fetcher2 = $rdf.fetcher(kb);                                          
    try {                                                                       
       await fetcher2.load(privateResource);                                    
       console.log("Private resource successfully loaded");                     
     }                                                                          
    catch(e) { console.log(e) }                                                 
}                                                                               
test();    
```

jeff-zucker: If I am correct in this, it is really important that within solidOS, if we need to create a new fetcher, we grab the _fetch from the existing fetch method in case the user has over-ridden it.

bourgeoa: Yes you are right. I was fully wrong. Fetcher do not default to current fetch.

jeff-zucker: Worth checking, though, thanks for your careful eye.

bourgeoa (approved):
LGTM

PR #534, opened 2021-12-13
Serializer set namespace prefix

pduchesne:
This PR adds the possibility to (forcefully) set namespaces prefixes on a Serializer instance, as opposed to the current `suggestNamespaces` and `suggestPrefix` methods that propose prefixes but cannot override defaults prefixes.

What it does is

- fix `IndexedFormula.setPrefixForURI` that did allow multiple namespaces to be mapped on the same prefix
- add `setNamespaces` and `setPrefix` methods on `Serializer` that can override default prefixes
- add a `options.namespaces`  to the `serialize` function, that defines namespaces that should be forcefully set
- override the `IndexedFormula.serialize` method so that the current graph namespaces are kept as mandatory prefixes

As a result ,
```
const graph: IndexedFormula = ...;

graph.serialize(base, contentType, provenance);
// is equivalent to
serialize(provenance, graph, base, contentType, undefined, { namespaces: graph.namespaces });
```
and keeps the graph namespace prefixes, while 
```
serialize(provenance, graph, base, contentType, undefined);
```
behaves as before, i.e. giving priority to default prefixes

### DIFF
--- a/src/formula.ts
+++ b/src/formula.ts
@@ -5,7 +5,7 @@ import log from './log'
 import RDFlibNamedNode from './named-node'
 import Namespace from './namespace'
 import Node from './node-internal'
-import Serializer from './serialize'
+import serialize from './serialize'
 import Statement from './statement'
 import {
   Bindings,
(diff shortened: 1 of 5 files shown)


timbl (approved):
Looks reasonable at a read-through

Issue #538, opened 2021-12-16
Natively Support In-Browser RDF


jeff-zucker:
Rdflib can not currently be used directly on RDF data stored in a browser's local storage because 1) default fetches all barf on any URL that isn't http(s) or file. 2) rdflib needs to see a content-type header to know what to do with documents it fetches 3) rdflib needs a way to know if/how a document is editable before it can write to it.                                                  
                                                                             
Simple usage can get around some these problems by using a 'MachineEditableDocument' triple, or by using `getItem + parse` instead of `load` and `serialize + setItem` instead of `putBack`.  @NoeldeMartin's apps use this approach (get+parse/serialze+put) though not with rdflib.                            
                                                                             
But even with these work-arounds, update can not work - it calls load and putBack with no possibility to replace them.  Similarly the forms system can not work - it calls update internally with no possibility to replace it.      
                                                                             
I have created a simple fetch that provides what rdflib needs to see in order to work directly on the local storage (it sends a full Response object with wac-allow and content-type headers).  This means that rdflib methods including `load`,`putback`,`update`,`webOperation` and the forms system can be used against documents in the local storage.                                   
                                                                             
A user or app would need to allow this storage by setting inBrowserStorage to true when creating a fetcher.  Once they had done that, their http and file fetches will work as previously but fetches in the form "ls://path" will read or write path in the browser's Local Storage.                             
```                                                                          
const fetcher = $rdf.fetcher(kb,{inBrowserStorage:true});                           
await fetcher.load('https://example.com/foo.ttl'); // loads remote file      
await fetcher.load('ls://foo.ttl');                             // loads  in-browser file 
```

josephguillaume:
I like it! It's an elegant solution for rdflib to work with documents in local storage without having to use solid-rest and browserfs.

In terms of use cases, it would be nice if this could play well with a local-first strategy.
With the proposed solution, an app would need to maintain two different sets of URIs. Presumably it could use a convention to replace ls by http/https to remember where it wanted to save the triples.

An alternative would be to maintain two separate stores, with one exclusively using this localStorage fetch. The advantage would be that the original why URIs could be kept.

Rather than committing to the use of ls:// now, perhaps we can first work out what the next step would look like.

[...]

PR #539, opened 2021-12-16
support local storage

jeff-zucker:
See  issue https://github.com/linkeddata/rdflib.js/issues/538


jeff-zucker:
Argh, forgot the typescript stuff ... fixing ...

src/fetcher.ts
```diff
@@ -193,6 +194,8 @@ export interface AutoInitOptions extends RequestInit{
   body?: string
   headers: HeadersInit
   credentials?: 'include' | 'omit'
+  // for local storage support
+  inBrowserStorage?: boolean
```
NoelDeMartin, line 198: Why do users need to add this? If we default to true, both http(s) and localStorage would work, so from a developer experience point of view that's better. Maybe the reason is that using this without actually using localStorage will always make an extra call with that if that checks for the `lp://` prefix, but I think the performance impact is negligible.

src/fetcher.ts
```diff
@@ -756,11 +759,39 @@ export default class Fetcher implements CallbackifyInterface {
...
+      }
+      else if(options.method==='GET'){
+        let content = window.localStorage.getItem(uri);
+        if(content) return new Response(content,success);
+        else {
+          failure.status = 404;
+          return(failure);
+        }
+      }
+      else return(failure);
```
NoelDeMartin, line 791: Would it be too difficult to support PATCH? And what about DELETE?

I have to say that this may work most of the time, but it may create some unexpected behaviour (I guess that's ok if we document this as "experimental").

The way I do it in my apps, I'm not really working with Solid when I use localStorage, in a different layer I am serializing my data as JsonLD, and then I am writing or modifying plain javascript objects into localStorage. So my code dealing with localStorage is not aware that it's dealing with semantic data.

example/localStorage.html
```diff
@@ -0,0 +1,37 @@
+<body>
```
NoelDeMartin, line 1: Besides this example, is it worth mentioning in the README or the documentation?

NoelDeMartin (commented):
It looks good overall, although I have some comments I've added inline.

Other than that, I have a couple more questions:
- If we're implementing this, why not IndexedDB as well? For apps with a lot of data, it should be a better approach.
- What happens with containers? If I'm creating a new file, do I also need to manually create the containment triples, and even the containers metadata? That would create some unexpected behaviour.

src/fetcher.ts
josephguillaume: I suppose one reason to make this opt-in would be that many applications would not need this functionality. I've only recently heard about tree shaking, so I'm not sure what it would involve here, but in principle we might want a developer to only import this fetcher if they need it?

src/fetcher.ts
josephguillaume: I suppose this solution actually doesn't know it's semantic data either - and that's the problem with supporting PATCH, which would require parsing of e.g. a sparql update query and then modification of the document. That said, rdflib is probably pretty close to having the parsing functionality already, so it might be possible. Either way, it probably makes sense for PATCH to be implemented in a separate PR.
DELETE seems more manageable.
I don't fully understand the limitations on content or keys in local storage but I assume that's what you mean by potential for unexpected behaviour?

josephguillaume:
> * What happens with containers? If I'm creating a new file, do I also need to manually create the containment triples, and even the containers metadata? That would create some unexpected behaviour.

I think this is solely about storing single documents in local storage. In a sense it's a simplified version of solid-rest, which provides a more complete solid interface.

src/fetcher.ts
josephguillaume: Ha. Took me a while to work it out but we don't need to implement PATCH because update with a WAC and non http Uri calls updateLocalFile, which only uses a PUT
https://github.com/linkeddata/rdflib.js/blob/e3037a1d6816591900608784856ecde6afd5eba1/src/update-manager.ts#L993

josephguillaume: Perhaps the failure could provide a message, e.g. localStorage only supports GET and PUT. Use updater.update to patch the document.

src/fetcher.ts
jeff-zucker: Yes, we could change to always-on.  It would add a single if statement to every fetch.  Not sure if that matters.

src/fetcher.ts
jeff-zucker: @josephguillaume  - the fact that update defaults to a PUT and that non-http URIs can be recognized as editable via wac-allow is due to my previous work to make rdflib compatible for file:// URLs.

@NoelDeMartin - Solid-rest does implement patch and container support and options and more.  At one point I had it working with indexedDb, Native File API, and Dropbox.  If someone needs a full-featured browser storage handler, they can dust off solid-rest and they should be good to go.  My thought was that all of that should be an outside library one could import and declare with the solidFetch global.  The purpose of this PR is to support rdflib's use of the storage - load, putBack, and update and minimal was my goal.

In Solid-Rest, I supported indexedDB and Native File Api through an external package called BrowserFS.  That means I could probably add support for them here with a line or two that would only be called if the user had BrowserFS imported.

jeff-zucker: Although this PR does an end-run around PATCH, Solid-Rest does implement actual PATCH.  I'm not sure there is much difference between PUT and PATCH when it comes to local storage.

src/fetcher.ts
NoelDeMartin: I think it's better to keep a simple API, so the less features that are configurable, the better. That is, if there is no reason to make it configurable of course.

What @josephguillaume mentions about tree-shaking would be a decent reason to make it opt-in, although at this point we wouldn't benefit a lot from tree-shaking the localStorage functionality because it's very small (but it could grow). But if we want to enable tree-shaking, it cannot be configured with an option. It should be implemented with some hooks or pluginable interface. Basically, applying the [inversion of control](https://en.wikipedia.org/wiki/Inversion_of_control) principle. Which in essence, can already be done with the `fetch` argument :). So maybe a better API, if we want to enable that, would be:

```js
$rdf.fetcher(kb, { fetch: localStorageFetch() });
```

src/fetcher.ts
NoelDeMartin: > I don't fully understand the limitations on content or keys in local storage but I assume that's what you mean by potential for unexpected behaviour?

@josephguillaume Well what I mean with that is that someone may presume that using `lp://` will work the same way than using `http://`, but I'm sure there are many things that will not work. And if what works/doesn't work is not clear, than can make this library frustrating to work with.

I haven't worked a lot with the, so maybe it's not necessary to handle PATCH, I was just raising the point to clarify.

src/fetcher.ts
jeff-zucker: We definitely do not want to use the fetch param.  Say I have a Dropbox fetch or a file:// fetch using solid-rest or even an inrupt-client fetch ... the app can define those with the fetch flag. 
 The current PR leaves those in place and adds the localStorage part on top of them.

jeff-zucker: Also, the way the PR does it supports the user over-riding the browser fetch.  They can set fetch in the call to fetcher to their own inBrowserStorage and leave inBrowserStorage false,

src/fetcher.ts
jeff-zucker: > someone may presume that using `ls://` will work the same way than using `http://`, but I'm sure there are many things that will not work. And if what works/doesn't work is not clear, than can make this library frustrating to work with.

My intention is to have this clear distinction : ls:// (or more likely browser://localStorage/) can be used in rdflib and libraries that import it wtih the load, putBack, and update methods.  As far as those methods go, the only difference between browser: and http: would be in loading  containers AFAIK.  Do you know of something else?

jeff-zucker:
> > * What happens with containers? 

The code for returning turtle for a container is only about 40 lines in solid-rest.  I could use that here, and that would solve reading a container.  We can already create a container with PUT either directly or indirectly by creating  file in a non-existent container.

src/fetcher.ts
josephguillaume: Given that the code doesn't to seem to import much that's new, I agree that it's probably a minimal increase to the bundle size and it doesn't matter in this case. I just think this is something we should be thinking about. I've recently been bothered by the size of jsonld when all I wanted was two functions. (I haven't yet worked out of that's fixable)

Following Noel's suggestion, if bundle size were important, then I would suggest a function e.g. createInBrowserFetch that took the user fetch as an argument and returned an augmented fetch for use in fetcher.
Alternatively, a wrapper around this constructor specifically for in-browser capabilities. 
A developer then has the choice on whether this code is in the bundle.

bourgeoa:
I'm a bit anxious to see solid protocol implemented in rdflib.
And container representation is not even stable.

josephguillaume:
I'm in favour of keeping this code minimal too and leaving out solid-specific implementation. Solid-rest seems like the right place for that.

josephguillaume:
Just noting that regarding the API choice (use of a dedicated Uri + constructor options parameter), I'd also added a comment on the linked issue regarding how I would want to use this functionality in a local first context.

https://github.com/linkeddata/rdflib.js/issues/538#issuecomment-995584629

jeff-zucker:
> I'm a bit anxious to see solid protocol implemented in rdflib. And container representation is not even stable.

My initial impulse was to go with minimal so I am fine with delegating container reads to solid-rest.

jeff-zucker:
This PR was obviously premature.  I have updated [Solid Rest Browser](https://github.com/solid/solid-rest/tree/main/browser) with a local-first example and a possible way to handle it.  Let's continue the discussion there and possibly brings some of it back into rdflib if that makes sense.

Issue #528, opened 2021-11-19
Move LiveStore and ConnectedStore to rdflib


timbl:
These are rdflib's most powerful classes, being versions of Store (aka IndexedFormula) which a Fetcher and UpdateManager attached as  helper objects.  For some reason their definition is not in the rdflib itself, but in solidUI

It is only logical to define them in rdflib.js. 
(And then move the use of them to rdflib)

timbl:
See also https://gitter.im/solid/solidos?at=619772c0abdd6644e3931f75 "the types creation in solid-logic is a hack made because rdflib do not publish the declaration files d.ts."

jeff-zucker:
Something to keep in mind when moving them to rdflib: in solid-ui the fetcher assumes a browser environment and should not in rdflib.

jeff-zucker:
See : https://github.com/solid/solidos/issues/72

timbl:
OnlineStore is just a Store with a Fetcher as a helper object.    LiveStore is just a OnlineStore with also a UpdateManager as an extra helper object.  They  only depend on things in rdflib -- there is nothing from Solid-UI they need.

timbl:
((note that the Fetcher is an rdflib.js class, nothing  to do with fetch() function.  There is a fetch() in rdflib,  that's where the platform-dependent and URI-scheme-dependent stuff can happen.    Maybe we want to make very easy to explicitly override by the app.  But that is a separate issue))

jeff-zucker:
@timbl - I am not sure which fetch() you are referring to or how it would be useful in other URI-schemes.  When we define a Fetcher object, it has an _fetch method which is what is used by load, putBack, updater, etc.  We can override the Fetcher._fetch method either by passing in a custom fetch on Fetcher creation or with a global or window solidFetch.   This has the correct effect on everything done with that fetcher obect.  However it does not impact the fetcher used by SolidLogicSingleton which is created on import of solid-logic and, as so far I have not been able to override it.

timbl:
The solid singleton object a single unique LiveStore ...  that was IIRC the reason to create the solid singleton idea at all.

I see that as you say, the Solid Singleton has its badly named `.fetcher`  at https://github.com/solid/solid-logic/blob/main/src/index.ts#L42 . It is really bad to have that  called `fetcher` instead of `fetch`.
I wonder if anyone uses it. 

Maybe we could remove this and suggest people use `kb.fetcher._fetch` 
[...]

PR #541, opened 2021-12-28
paste in LiveStore from paneRegistry.ts

timbl:
with adaptation to its new home
and removing a bug that LiveStore inherited from Store instead of from ConnectedStore


theRealImy (approved):

theRealImy:
connected issue: https://github.com/linkeddata/rdflib.js/issues/528 (for reference)

theRealImy (approved):

PR #543, opened 2022-01-13
turtle serialization : add whitespace between colon and dot

### LINKED ISSUE #542: Turtle serializer : no whitespace between colon and dot when a prefixed symbol has no local name
If a prefixed symbol is a statement's object and has no local name, it results in a colon immediately followed by a dot, e.g. `example:subject example:pred example:.`
In particular, the automatic prefix production of the turtle serializer produces such outputs if a symbol has a trailing slash.

While this is syntactically valid as per the turtle grammar, some other tools do not handle this well (e.g. RDFLib/rdflib , cf RDFLib/rdflib#1664) and require a whitespace between the colon and the dot.
In the sake of cross-ecosystems interoperability, fixing this would be an easy win.

pduchesne:
Fixes #542

### DIFF
--- a/src/serializer.js
+++ b/src/serializer.js
@@ -340,7 +340,13 @@ export class Serializer {
         if (typeof branch === 'string') {
           if (branch.length === 1 && str.slice(-1) === '\n') {
             if (',.;'.indexOf(branch) >= 0) {
-              str = str.slice(0, -1) + branch + '\n' //  slip punct'n on end
+              str = str.slice(0, -1)
+              // be conservative and ensure a whitespace between some chars and a final dot, as in treeToLine above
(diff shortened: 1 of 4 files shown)


bourgeoa:
If the code is actually correct from spec perspective why not change the other libraries that refuse it ?
Manually made turtle with dot without space will still be refused by other  libraries.

TallTed:
@bourgeoa -- Postel's Law dictates that this is the best choice -- i.e., be liberal in what you accept, and conservative in what you generate.  At some point in the future, the tools that choke on `prefix:.` may be fixed, but until that point, it's trivial (as demonstrated by this PR) to make `rdflib.js` emit `prefix: .` (and still to accept `prefix:.`).

PR #551, opened 2022-03-07
Insist that ConnectedStore has a fetcher and LiveStore has an update…

timbl:
to save TS worrying about it in the calling modules.

Avoids TS errors like "liveStore.fetcher may be undefined"

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -120,6 +120,9 @@ export {
 export { termValue } from './utils/termValue'
 
 export class ConnectedStore extends Store {
+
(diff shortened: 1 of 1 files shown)


bourgeoa:
LGTM

timbl:
the change from DOMtree to HTMLDocument was to fix an error which only showed in solid-ui -- the way TSC complains about remote bits of code can be a bit annoying.   I didn't know that it would absorb the type info in old JS comments.

PR #552, opened 2022-03-24
local dev updated to node 16, CI is same

theRealImy:
See: https://github.com/solid/solidos/issues/125


theRealImy:
Just to clarify -> this PR only changes the local dev environment to node 16.14.0 (see .nvmrc)
CI will build on 12, 14 and 16 and npm contains node 14 published versions.

PR #556, opened 2022-05-12
Simplify example code showing setting solidFetcher

ThisIsMissEm:
In the `@inrupt/solid-client-authn-browser` SDK, we always expose a `fetch` method, so you don't need to set `solidFetcher` only after login, you can do this before and it'll switch based on whether you're logged in or not — by default the `fetch` method uses global fetch until the session is established, at which point it switches to fetch with authorization and dpop headers.


jeff-zucker:
LGTM, thanks for catching those.

Issue #405, opened 2020-04-16
Merging graphs with blank nodes


th1j5:
I'm using rdflib in a project to update an existing graph with new data. This new data contains a bit overhead, because it is possible a certain 'device1' already exists, but when merging with the old graph they have the same IRI's and thus mean the same. We are using blank nodes to represent different measurements.
Old graph:
```
@prefix ns0: <https://florsanders.inrupt.net/public/ontologies/omalwm2m.ttl#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

<http://basisLeshan.com/device1/3303/0>
  a ns0:ObjectInstance;
  ns0:consistsOf [
    a ns0:5700, ns0:ResourceInstance ;
    ns0:hasTimeStamp "2020-04-16T08:24:03.755Z"^^xsd:dateTime ;
    ns0:hasValue "-2.5"^^xsd:float ;
    ns0:organizedInto <http://basisLeshan.com/device1/3303/0>
  ] ; 
  ns0:containedBy <http://basisLeshan.com/device1> .

<http://basisLeshan.com/device1>
  a ns0:Device ;
  ns0:contains <http://basisLeshan.com/device1/3303/0> .
```
new Graph:
```
@prefix ns0: <https://florsanders.inrupt.net/public/ontologies/omalwm2m.ttl#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

<http://basisLeshan.com/device1/3303/0>
  a ns0:ObjectInstance;
  ns0:consistsOf [
    a ns0:5700, ns0:ResourceInstance ;
    ns0:hasTimeStamp "2020-04-16T08:26:32.988Z"^^xsd:dateTime ;
    ns0:hasValue "-5.1"^^xsd:float ;
    ns0:organizedInto <http://basisLeshan.com/device1/3303/0>
  ] ; 
  ns0:containedBy <http://basisLeshan.com/device1> .

<http://basisLeshan.com/device1>
  a ns0:Device ;
  ns0:contains <http://basisLeshan.com/device1/3303/0> .
```
(The only difference is a different value & timestamp)

When I would merge them, I would expect these 2 blank nodes to kept seperate, because there is nothing that could suggest otherwise. This is also the behavior of [rdflib in python](https://rdflib.readthedocs.io/en/stable/).

Thus:
```
#!/usr/bin/env python3                                                          
from rdflib import Graph                                                        
[...]

PR #558, opened 2022-07-19
Fix for `BlankNode` as `NamedNode` when parsing JSON-LD #555

### LINKED ISSUE #555: Turtle serializer shouldn't write blank nodes as <...>
I'm converting JSON-LD to Turtle using rdflib.js.

Example input:

```json
{
    "@context": {
        "ex": "http://example.com#",
    },
    "@id": "ex:myid",
    "ex:prop1": {
        "ex:prop2": {
            "ex:prop3": "value",
        },
    },
}
```

Example current output out of rdflib.js:

```ttl
@prefix ex: <http://example.com#>.

<_:b0> ex:prop2 <_:b1>.
<_:b1> ex:prop3 "value".
ex:myid ex:prop1 <_:b0>.
```

[Turtle spec](https://www.w3.org/TR/turtle/#BNodes) states following:

> [RDF blank nodes](http://www.w3.org/TR/rdf11-concepts/#dfn-blank-node) in Turtle are expressed as _: followed by a blank node label which is a series of name characters.

So, I think blank nodes should be expressed without <...>, because this makes them absolute or relative IRIs and not blank nodes.

As an additional feature, it would be nice to be able to control the blank node output to have them nested or not nested.

Questions:
1. Is this a known issue? I saw some non-conformances in #329, but couldn't find this exact case there.
2. Could this be affected by arguments? In case I'm calling the functions wrong? I'm including below my code snippet.

```ts
/**
 * Convert JSON-LD to Turtle
 * @param input JSON string
 * @param base Base IRI for the content
 * @param namespaces The namespace map for use in ttl
 * @returns TTL string
 */
async function convertJsonLdToTtl(
    input: string,
    base: string,
    namespaces: Record<string, string> = {},
): Promise<string> {
    return new Promise<string>((res, rej) => {
        const store = rdflib.graph()
        rdflib.parse(input, store, base, "application/ld+json", (err, kb) => {
            if (err) {
                rej(err)
            } else {
                if (!kb) {
                    rej("KB empty: " + kb)
                } else {
                    console.log("KB # statements: " + kb.statements.length)
                    rdflib.serialize(
                        null,
                        kb,
                        undefined,
                        "text/turtle",
                        (err, output) => {
                            if (err) {
                                rej(err)
                            } else {
                                if (!output) {
                                    rej("Empty output: " + output)
                                } else {
                                    res(output)
                                }
                            }
                        },
                        {
                            namespaces,
                        },
                    )
                }
            }
        })
    })
}

```

Many thanks.

RinkeHoekstra:
The JSON-LD parser currently uses the JSON-LD flattening algorithm of `jsonld.js`. This algorithm assigns `@id` attributes to blank nodes, but the JSON-LD parser only creates a `BlankNode` when no `@id` is present. 

This means that currently, the blank nodes in JSON-LD are interpreted as `NamedNode`, which is incorrect.

This pull request does the following:

* Add a check to see whether the `@id` value starts with `_:`. 
* If it does, it is a `BlankNode`, otherwise it is a `NamedNode`.
* If it is a blank node, strip it from the `_:` as this will be added by the instantiation of the `BlankNode` class.
* Make sure to perform this check whenever a new object node is created (there were three separate checks in the code)

This fixes #555 

Also raised an issue in `rdf-canonize` (<https://github.com/digitalbazaar/rdf-canonize/issues/45>) to flag the problem that blank node identifiers generated by `jsonld.js` are not unique across loads (this is the cause of issue #405 )

### DIFF
--- a/src/jsonldparser.js
+++ b/src/jsonldparser.js
@@ -43,10 +43,7 @@ export function jsonldObjectToTerm (kb, obj) {
     return kb.rdfFactory.literal(obj['@value'])
   }
 
-  // I'm not sure we should return a literal if we don't have any of the above.
-  // I think this should be a blank node by default
-  return kb.rdfFactory.BlankNode();
-  // return kb.rdfFactory.literal(obj)
+  return kb.rdfFactory.literal();
 }
 
 /**
--- a/src/jsonldparser.js
+++ b/src/jsonldparser.js
@@ -22,7 +22,13 @@ export function jsonldObjectToTerm (kb, obj) {
   }
 
   if (Object.prototype.hasOwnProperty.call(obj, '@id')) {
-    return kb.rdfFactory.namedNode(obj['@id'])
+    if (obj['@id'].startsWith('_:')) {
+      // This object is a Blank Node
+      return kb.rdfFactory.blankNode(obj['@id']);
+    } else {
+      // This object is a Named Node
+      return kb.rdfFactory.namedNode(obj['@id']);
+    }    
   }
 
   if (Object.prototype.hasOwnProperty.call(obj, '@language')) {
--- a/src/jsonldparser.js
+++ b/src/jsonldparser.js
@@ -23,8 +23,8 @@ export function jsonldObjectToTerm (kb, obj) {
 
   if (Object.prototype.hasOwnProperty.call(obj, '@id')) {
     if (obj['@id'].startsWith('_:')) {
-      // This object is a Blank Node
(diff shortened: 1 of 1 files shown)


bourgeoa:
CI fails

bourgeoa:
resolved with https://github.com/linkeddata/rdflib.js/pull/589/commits/17052b62209a92112d21a01dd8d2cd2f5fde16a3

PR #560, opened 2022-07-19
Added `prepare` hook for building directly from source (GitHub)

### LINKED ISSUE #559: Installing from GitHub source will not include `lib`, `dist` and `esm` folders
When installing rdflib from latest source, e.g. through GitHub, the `npm install` script does not build the package properly and does not create the `lib`, `dist` and `esm` folders.

RinkeHoekstra:
A standard `npm install` from source (especially GitHub) did not build all of the packages in `lib`, `dist` and `esm`.

This adds a `prepare` hook in `package.json` that ensures that everything is built properly before the `rdflib` packages is added to `node_modules`.

This fixes #559

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,7 @@
     "test": "npm run test:unit && npm run test:serialize && npm run test:types",
     "test:clean": "rimraf tests/serialize/,*",
     "test:serialize": "npm run build && npm run test:serialize:all && npm run test:clean",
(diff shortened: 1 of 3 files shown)


angelo-v:
I am not sure if I get what you are trying to achieve. Are you taking about something like `npm install git+ssh://...`?

I am against using the prepare hook to build things, as it slows down every `npm install`.  If you want to build, then use the build scripts. If you need to install dev snapshots that match the github code, then we should publish them via the pipeline. You can also checkout and build the github code and then install the local module in your project.

PR #566, opened 2022-07-24
Use accept headers that are supplied instead of always overwriting

ludwigschubi:
I got aware of this because of the new changes made to ess (enterprise solid server). Because i tried fetching the new webId form of https://id.inrupt.com/<user> but it would not load the data (only html) even if I passed accept headers to `.load()`. These changes would allow supplied accept headers to actually work.

### DIFF
--- a/src/fetcher.ts
+++ b/src/fetcher.ts
@@ -1054,9 +1054,11 @@ export default class Fetcher implements CallbackifyInterface {
       options.cache = 'no-cache'
     }
 
-    let acceptString = this.acceptString()
-    // @ts-ignore
-    options.headers['accept'] = acceptString
+    if(!options.headers["accept"]){
+      let acceptString = this.acceptString()
+      // @ts-ignore
+      options.headers['accept'] = acceptString
+    }
 
     let requestedURI = Fetcher.offlineOverride(uri)
     options.requestedURI = requestedURI
--- a/src/fetcher.ts
+++ b/src/fetcher.ts
@@ -1054,7 +1054,7 @@ export default class Fetcher implements CallbackifyInterface {
       options.cache = 'no-cache'
     }
 
-    if(!options.headers["accept"]){
+    if(!options.headers["accept"] || (!(options.headers as Headers).get("accept"))){
       let acceptString = this.acceptString()
       // @ts-ignore
       options.headers['accept'] = acceptString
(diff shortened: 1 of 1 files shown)


ludwigschubi:
I needed to make the condition more elaborate because:
```javascript
const headers = new Headers({ accept: "text/turtle" })
!!headers.get("accept") // true
!!headers["accept"] // false
```

jeff-zucker:
I can not reproduce your problem.  This correctly prints out both the ESS and NSS webids as retrieved from the WebID Profile Documents with rdflib.load

```javascript
const store = $rdf.graph();                                                                                  
const fetcher = $rdf.fetcher(store);                                                                         
                                                                                                             
const isa = $rdf.sym("http://www.w3.org/1999/02/22-rdf-syntax-ns#type");                                     
const agent = $rdf.sym("http://xmlns.com/foaf/0.1/Agent");                                                   
const person = $rdf.sym("http://schema.org/Person");                                                         
const ess = $rdf.sym('https://id.inrupt.com/jeff-zucker');                                                   
const nss = $rdf.sym('https://jeff-zucker.solidcommunity.net/profile/card#me');                              
                                                                                                             
(async ()=>{                                                                                                 
  await fetcher.load(nss,{headers:{Accept:'text/turtle'}})                                                   
  console.log( store.any( null, isa, person ).value);                                                        
  await fetcher.load(ess,{headers:{Accept:'text/turtle'}})                                                   
  console.log( store.any( null, isa, agent ).value);                                                         
})(); 
```

jeff-zucker:
Apparently, rdflib is able to handle this and correctly supplies RDF from the RDFa source.  NSS uses rdflib to send turtle in get requests with accept headers, so a curl or wget with accept headers returns turtle from RDFa.  BUT a wget or curl against a ESS WebID  will return HTML even with the accept header.  This is because the ESS WebID document is not on a Solid server - it is kept separate from the storage server by design.  Still, it should supply RDF for the WebID document.

ludwigschubi:
Hey jeff,
Thank you for responding to me so quickly, I really appreciate that! I will try to reproduce the issue

ludwigschubi:
I just tested it you are right it works already as it should. My problem is that I am using rdflib@2.2.12 because of incompatibility issues with webpack polyfills that i need to make [shex-methods](https://github.com/ludwigschubi/shex-methods) work in the browser. Dependency hell but not the fault of rdflib. In this older version it doesn't work, so i think i have to find another way of using shex so that i can use the latest rdflib version. Sorry for being so quick with creating an issue but it wasn't so easy spotting the origin of the misbehavior :v:

PR #569, opened 2022-08-24
Do not anonymize triples in INSERT DATA query

smessie:
In INSERT DATA and DELETE DATA, the generated names should never start with a '?'.

I, however, had this problem where the generated SPARQL query contained `?something` names for the triples while it was a `INSERT DATA` and `DELETE DATA` query.
To prevent this from happening, in this PR, I changed it so that the `anonymizeNT` is only used in `DELETE`, `INSERT` and `WHERE` clauses, but no longer in the `DELETE DATA` and `INSERT DATA`.


bourgeoa:
Can you give some more information on the reasons. SPAQL specifications, working failures .....

smessie:
> Variables in QuadDatas are disallowed in INSERT DATA requests (see Notes 8 in the [grammar](http://www.w3.org/TR/sparql11-query/#sparqlGrammar)). That is, the INSERT DATA statement only allows to insert ground triples. Blank nodes in QuadDatas are assumed to be disjoint from the blank nodes in the Graph Store, i.e., will be inserted with "fresh" blank nodes.

From https://www.w3.org/TR/sparql11-update/#insertData, however, now `anonymizeNT` is used, which results in the chance that https://github.com/linkeddata/rdflib.js/blob/a78fc0f3cada4358e941c8feedb7eb61149b1459/src/update-manager.ts#L198 is used to transform the object into a string, which results in a variable, which is disallowed.

smessie:
When debugging another case, I saw that I missed following part in the spec
> QuadData denotes triples to be removed and is as described in [INSERT DATA](https://www.w3.org/TR/sparql11-update/#insertData), with the difference that in a DELETE DATA operation neither variables nor blank nodes are allowed (see Notes 8+9 in the [grammar](http://www.w3.org/TR/sparql11-query/#sparqlGrammar)).

So I've altered my commit so it only changes `INSERT DATA` and no longer `DELETE DATA`, so it is conform the spec again.

angelo-v:
Please add a unit test that covers your change.

smessie:
@angelo-v I've added a unit test, and while doing so I figured out that this specific test file was not executed because it was not suffixed with `-test`, and https://github.com/linkeddata/rdflib.js/blob/a78fc0f3cada4358e941c8feedb7eb61149b1459/package.json#L120 requires this. So I've fixed this as well.

PR #572, opened 2022-09-02
Update implementation of removeMatches to use removeMany

Jan-PieterBaert:
Since `removeMany` and `removeMatches` both do the same thing (removing statements that match the arguments) it's best to let one refer to the other, this avoids bugs (e.g. the one mentioned in https://github.com/linkeddata/rdflib.js/blob/main/src/store.ts#L900 ) happening in one of the implementations.


angelo-v:
`removeMatches` seems not to be covered by unit test. I prefer to add tests first, to ensure the refactoring actually does not change the behaviour unintendetly

angelo-v (approved):

PR #574, opened 2022-09-06
Update applyPatch function and add tests for it

Jan-PieterBaert:
The `applyPatch` function had a few small bugs when I tried it, so I fixed those and added tests for the function

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -102,9 +102,9 @@ export default class Collection<
    * @param bindings - The bindings to substitute
    */
   substitute (bindings: Bindings) {
-    const elementsCopy = this.elements.map((ea) => ea.substitute(bindings))
-
-    return new Collection(elementsCopy) as Collection<Node | Collection<any> | Literal | Variable>
+    let collection = new Collection();
+    this.elements.forEach((ea) => {collection.append(ea.substitute(bindings))})
+    return collection as Collection<Node | Collection<any> | Literal | Variable>;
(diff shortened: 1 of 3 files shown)


angelo-v:
Hi and thanks for your contribution. Can you please describe what bugs you experienced and how your changes fix those?

Jan-PieterBaert:
The bugs I found:
- the `substitute` function of a collection didn't seem to work with the current constructor, creating a new collection and adding the elements did work
- in the `applyPatch` function `patch['insert']` was always used, but in the function declaration it said `'patch'`
- the `ds = ds.statements` line meant that if there was a where clause this statement would be executed twice instead of once which meant there would suddenly need to be a `ds.statements.statements`
- since the other functions seem to assume `ds` is a collection and the elements of a collection are accessed via `elements` and not `statements` I edited that as well
- the `targetKB.query` function failed because the `initBindings` was always undefined, the code seemed to assume it's a list so I made it an empty list

timbl (approved):
On the face of it this all seems to make sense.

timbl:
Looking back at this now it is merged (and broken for solid-ui.)  ` ds` was designed to be an Array, not a Collection - which code expects it to be a Collection?

timbl:
So the test is wrong too.  The parameters are expected to be  IndexedFormulas, not collections.  Let's roll this back.   Sorry I don't read it carefully in review.

PR #577, opened 2022-09-29
n3parser now fixes lists to use Collection objects if necessary. 

timbl:
Also a few  more tests and a bit of test clean-up. (Make them less dependent on arbitrary things like bnode IDs.)


tests/unit/parse-test.js
```diff
@@ -52,8 +80,71 @@ describe('Parse', () => {
...
+`ent:Q328
+      rdfs:label
+              [
+                  rdf:first 1;
+                  rdf:rest
+                  [ rdf:first 2; rdf:rest [ rdf:first 3; rdf:rest rdf:nil ] ]
+              ].
+
+`))
+/*
```
angelo-v, line 128: Why is this commented out? Is it an open TODO or should it be removed?

PR #579, opened 2022-10-19
Create update-manager.md

jeff-zucker:
Documentation on the update manager

### DIFF
--- a/Documentation/update-manager.md
+++ b/Documentation/update-manager.md
@@ -56,6 +56,6 @@ Connection: keep-alive
 Keep-Alive: timeout=5
 ```
 
-Note : currently rdflib does not support N3-PATCH although it is, AFAIK, the only method of update mentioned in the specs.  To fix this would require changes to the editable function to accept an N3-PATCH header and changes to the fire function to be able to submit an N3-PATCH.  See https://github.com/SolidOS/solidos/issues/185.
+Note: Currently, `rdflib` does not support `N3-PATCH` although it is, AFAIK, the only method of update mentioned in the specs.  To fix this would require changes to the `editable` function to accept an `N3-PATCH` header and changes to the `fire` function to be able to submit an `N3-PATCH`.  See <https://github.com/SolidOS/solidos/issues/185>.
 
 Also Note : When submitting a form, PATCH is used *except* when re-ordering elements in an ordered list, in which case PUT is used.
--- a/Documentation/update-manager.md
+++ b/Documentation/update-manager.md
@@ -35,7 +35,7 @@ A different mechanism for updates using web sockets uses the `updates-via` heade
 
 ### About Headers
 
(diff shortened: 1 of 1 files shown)


Documentation/update-manager.md
```diff
@@ -29,6 +29,33 @@
 
   * is updated using PUT if the ms-author-via header contains the word "DAV"
 
+### Web Sockets
+
+A different mechanism for updates using web sockets uses the "updates-via" header to send and receive using sockets. See src/updates-via
```
TallTed, line 34:
```suggestion
A different mechanism for updates using web sockets uses the `updates-via` header to send and receive using sockets. See <src/updates-via>
```

Documentation/update-manager.md
```diff
@@ -29,6 +29,33 @@
...
+
+### About Headers
+
+Headers include information from the server about how to interact with it.  You can see the headers for any response with e.g. `curl --head URI` which, on the Solid Community server produces something like this (note the wac-allow,ms-author-via,and updates-via headers):
```
TallTed, line 38:
```suggestion
Headers include information from the server about how to interact with it.  You can see the headers for any response with a command like `curl --head URI` which, on the Solid Community server produces something like this (note the `wac-allow,ms-author-via`, and `updates-via` headers):
```

Documentation/update-manager.md
```diff
@@ -29,6 +29,33 @@
...
+Updates-Via: wss://solidcommunity.net
+Content-Type: text/html; charset=utf-8
+Content-Length: 2
+ETag: W/"2-nOO9QiTIwXgNtWtBJezz8kv3SLc"
+Date: Wed, 19 Oct 2022 18:43:08 GMT
+Connection: keep-alive
+Keep-Alive: timeout=5
+```
+
 Note : currently rdflib does not support N3-PATCH although it is, AFAIK, the only method of update mentioned in the specs.  To fix this would require changes to the editable function to accept an N3-PATCH header and changes to the fire function to be able to submit an N3-PATCH.  See https://github.com/SolidOS/solidos/issues/185.
```
TallTed, line 59:
```suggestion
Note: Currently, `rdflib` does not support `N3-PATCH` although it is, AFAIK, the only method of update mentioned in the specs.  To fix this would require changes to the `editable` function to accept an `N3-PATCH` header and changes to the `fire` function to be able to submit an `N3-PATCH`.  See <https://github.com/SolidOS/solidos/issues/185>.
```

Documentation/update-manager.md
```diff
@@ -29,6 +29,33 @@
...
 
 Also Note : When submitting a form, PATCH is used *except* when re-ordering elements in an ordered list, in which case PUT is used.
```
TallTed, line 61:
```suggestion
Also Note: When submitting a form, `PATCH` is used *except* when re-ordering elements in an ordered list, in which case `PUT` is used.
```

TallTed (changes requested):

PR #587, opened 2022-11-27
updates for latest dependencies

bourgeoa:
Some changes where needed for updating to latest typescript@4.9.3 and node-fetch@3.3.0


bourgeoa:
I updated to latest dependencies that allowed to resolve a few issues in code and typescript
A special issue with jsonld that do not support node 12 ans only node 14 latest.
Thus change in CI.

I kept publish with node 14. Shall I use node 16 ?

angelo-v:
> I kept publish with node 14. Shall I use node 16 ?

Yes, please use 16, node 14 will not get security updates in a few month

angelo-v (approved):

PR #590, opened 2022-12-04
Serialize jsonld

bourgeoa:


bourgeoa:
@timbl @angelo-v 
The issue I have is that my patch returns a `Promise<string> and not a `String` 
To serialize `application/ld+json` in tests I need to make the the test `async` 'await serialize(...)

Have you any idea :
- how to resolve the Promise inside /rdflib/serialize.ts code
- if not is this kind of use acceptable  :  
```
async function serializeStore (store, contentType) {
const result = contentType === application/ld+json ? 
   await serialize(null, store, null, contentType) : 
   serialize(null, store, null, contentType)
}
```

tests/serialize/data.js
```diff
@@ -122,7 +121,7 @@ var doNext = function (remaining) {
           })
           return
 
-          case '-dump':
+          /* case '-dump':
```
angelo-v, line 124: let's remove unused code

src/serialize.ts
```diff
@@ -114,4 +110,11 @@ export default function serialize (
       return result as string
     }
   }
+
+  function toJsonld (item ) {
+    try {
+      return jsonld.fromRDF(item, {format: 'application/n-quads'}).then( docJsonld => { return JSON.stringify(docJsonld) })
+      // return JSON.stringify(await jsonld.fromRDF(item, {format: 'application/n-quads'}))
```
angelo-v, line 117: let's remove unused code

tests/unit/parse-test.js
```diff
@@ -231,7 +235,70 @@ describe('Parse', () => {
       })
     })
 
-    describe('with a @type node', () => {
+    describe('jsonld with blanknodes', () => {
```
angelo-v, line 238: Thanks for adding tests :+1:

angelo-v (approved):

PR #595, opened 2023-01-23
Ts tweaks

timbl:
When compiling profile-pane, tsc objects unless these changes are made to rdflib.
Clearly flagging things as ReadOnly can be a bit of a trap as you have to then propagate it all over your code.
So here we just tweak one file and stop it spreading to the store.


timbl:
.... but the CI tests say
```
npm ERR! `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
```

and many errors like
```
npm ERR! Missing: @babel/cli@7.20.7 from lock file
```

PR #621, opened 2023-05-05
redirect to container

bourgeoa:


src/fetcher.ts
```diff
@@ -601,7 +601,11 @@ class N3Handler extends Handler {
   ): ExtendedResponse | Promise<FetchError> {
     // Parse the text of this N3 file
     let kb = fetcher.store
-    let p = N3Parser(kb, kb, options.original.value, options.original.value,
+    let baseUrl = options.original.value
+    const isContainer = kb.any(options.original, null, ns.ldp('Container'))
+    console.log('@@ isContainer ' + isContainer)
+    if (isContainer && !baseUrl.endsWith('/')) baseUrl = baseUrl + '/'
```
NoelDeMartin, line 607: I don't think this is the proper way to solve this problem, what if the redirect goes to a different path altogether (for whatever reason)? We should be reading from the `response.url` instead. And if we don't consider redirecting to a completely different resource a valid situation, throw an error rather than keep going with potentially wrong data.

src/fetcher.ts
```diff
@@ -601,7 +601,11 @@ class N3Handler extends Handler {
...
+    let p = N3Parser(kb, kb, options.original.value, baseUrl,
```
bourgeoa, line 608:
```suggestion
    let p = N3Parser(kb, kb, baseUrl, baseUrl,
```

Issue #631, opened 2024-01-15
deleteDocument() : Error : Statement to be removed is not on store


bourgeoa:
For example trying to edit with the source-pane `https://bourgeoa.solidcommunity.net:8443/public/`
```
Statement to be removed is not on store: _:n185 <http://www.w3.org/2007/ont/link#status> _:186 .
```
`_:186` object is the term.value of `Collection`

The issue occurs only at the `st.object` being an `(object) Collection` in the `metadata status part`.
It looks like if `st.object` do not get the Collection elements.

`removeMatches()` calls `removeMany()` that calls `remove()` that fails on that object Collection.

bourgeoa:
This must be better explained : 
This is the serialization of the `metadata` for `https://bob.localhost:8443/profile/card`
```
    @prefix : <#>.
    @prefix http: <http://www.w3.org/2007/ont/http#>.
    @prefix httph: <http://www.w3.org/2007/ont/httph#>.
    @prefix tabont: <http://www.w3.org/2007/ont/link#>.
    @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
    @prefix pro: <https://bob.localhost:8443/profile/>.
    @prefix tur: <http://www.w3.org/ns/iana/media-types/text/turtle#>.
    
    pro:card a tabont:Document, tabont:RDFDocument, tur:Resource.
    [
      rdfs:label "[0:16:34]  Request for https://bob.localhost:8443/profile/card";
      tabont:requestedURI "https://bob.localhost:8443/profile/card";
      tabont:response
              [
                  http:status 200;
                  http:statusText "OK";
                  httph:accept-patch
                      "text/n3, application/sparql-update, application/sparql-update-single-match";
                  httph:accept-post "*/*";
                  httph:accept-put "*/*";
                  httph:access-control-allow-credentials "true";
                  httph:access-control-expose-headers
                      "Authorization, User, Location, Link, Vary, Last-Modified, ETag, Accept-Patch, Accept-Post, Accept-Put, Updates-Via, Allow, WAC-Allow, Content-Length, WWW-Authenticate, MS-Author-Via, X-Powered-By";
                  httph:allow "OPTIONS, HEAD, GET, PATCH, POST, PUT, DELETE";
                  httph:connection "keep-alive";
                  httph:content-type "text/turtle";
                  httph:date "Thu, 08 Feb 2024 23:16:35 GMT";
                  httph:keep-alive "timeout=5";
                  httph:link
[...]

PR #633, opened 2024-02-04
removeMetadata

bourgeoa:
`store.remove()` fails on triple containing a `Collection`

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -871,25 +871,46 @@ export default class IndexedFormula extends Formula { // IN future - allow pass 
   }
 
   /**
+   * Removes all metadata
+   * @param doc - The document / graph
+   */
+  removeMetadata(doc: Quad_Graph): IndexedFormula {
+    const meta = this.fetcher?.appNode // this.sym('chrome://TheCurrentSession')
+    const linkNamespaceURI = 'http://www.w3.org/2007/ont/link#'
+    const kb = this
+    // removeMatches() --> removeMany() --> remove() fails on Collection
+    function removeBySubject (subject) {
+      const sts = kb.statementsMatching(subject, null, null, meta)
+      // console.log(sts)
+      for (var i = 0; i < sts.length; i++) {
+        kb.removeStatement(sts[i])
+      }
+    }
+    const requests = this.statementsMatching(null, this.sym(`${linkNamespaceURI}requestedURI`), this.rdfFactory.literal(doc.value), meta).map(st => st.subject)
+    for (var r = 0; r < requests.length; r++) {
+      const request = requests[r]
+      if (request != null) { // null or undefined
+        const response = this.any(request, this.sym(`${linkNamespaceURI}response`), null, meta) as Quad_Subject
+        // console.log('REQUEST ' + request.value)
+        removeBySubject(request)
+        if (response != null) { // null or undefined
+          // console.log('RESPONSE ' + response.value)
+          removeBySubject(response)
+        }
+      }
+    }
+    // console.log('DOCTYPE ' + doc.value)
+    removeBySubject(doc)
+    return this
+  }
+
+  /**
    * Removes all statements in a doc, along with the related metadata including request/response
    * @param doc - The document / graph
    */
   removeDocument(doc: Quad_Graph): IndexedFormula {
-    const meta = this.sym('chrome://TheCurrentSession') // or this.rdfFactory.namedNode('chrome://TheCurrentSession')
-    const linkNamespaceURI = 'http://www.w3.org/2007/ont/link#' // alain
     // remove request/response and metadata
-    const requests = this.statementsMatching(undefined, this.sym(`${linkNamespaceURI}requestedURI`), this.rdfFactory.literal(doc.value), meta).map(st => st.subject)
-    for (var r = 0; r < requests.length; r++) {
-      const request = requests[r]
-      if (request !== undefined) {
-        this.removeMatches(request, null, null, meta)
-        const response = this.any(request, this.sym(`${linkNamespaceURI}response`), null, meta) as Quad_Subject
-        if (response !== undefined) { // ts
-          this.removeMatches(response, null, null, meta)
-        }
-      }
-    }
-    this.removeMatches(this.sym(doc.value), null, null, meta) // content-type
+    this.removeMetadata(doc)
 
     // remove document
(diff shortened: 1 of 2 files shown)


src/store.ts
```diff
@@ -871,25 +871,46 @@ export default class IndexedFormula extends Formula { // IN future - allow pass
...
-  removeDocument(doc: Quad_Graph): IndexedFormula {
-    const meta = this.sym('chrome://TheCurrentSession') // or this.rdfFactory.namedNode('chrome://TheCurrentSession')
-    const linkNamespaceURI = 'http://www.w3.org/2007/ont/link#' // alain
-    // remove request/response and metadata
-    const requests = this.statementsMatching(undefined, this.sym(`${linkNamespaceURI}requestedURI`), this.rdfFactory.literal(doc.value), meta).map(st => st.subject)
+  removeMetadata(doc: Quad_Graph): IndexedFormula {
+    const meta = this.fetcher?.appNode // this.sym('chrome://TheCurrentSession')
+    const linkNamespaceURI = 'http://www.w3.org/2007/ont/link#'
+    const kb = this
+    // removeMatches() --> removeMany() --> remove() fails on Collection
```
angelo-v, line 881: Wouldn't it be better to fix the remove() function to work with Collection instead of applying a workarround here?

src/store.ts
```diff
@@ -871,25 +871,46 @@ export default class IndexedFormula extends Formula { // IN future - allow pass
...
+      for (var i = 0; i < sts.length; i++) {
+        kb.removeStatement(sts[i])
+      }
+    }
+    const requests = this.statementsMatching(null, this.sym(`${linkNamespaceURI}requestedURI`), this.rdfFactory.literal(doc.value), meta).map(st => st.subject)
     for (var r = 0; r < requests.length; r++) {
       const request = requests[r]
-      if (request !== undefined) {
-        this.removeMatches(request, null, null, meta)
+      if (request != null) { // null or undefined
```
angelo-v, line 892: The comment says "or undefined" but the code onyl checks for null. Why?

tests/unit/update-manager-test.js
```diff
@@ -223,6 +223,17 @@ describe('UpdateManager', () => {
       expect(updater.editable(doc1)).to.equal(undefined)
     })
 
+    it('Should not detect a document is editable from metadata after removeMetadata', () => {
```
angelo-v, line 226: How are these tests related to the changes? I would have expected a test that assures that Collections are now working.

src/store.ts
bourgeoa: Loose inequality does it
It is used in other places in rdflib

tests/unit/update-manager-test.js
bourgeoa: They are not. It only test the new function `removeMetadata`

src/store.ts
bourgeoa: Yes it would. I did not find howto.

timbl: What goes wrong with when one f the nodes being removed is a collection?

bourgeoa: @timbl
This is I hope a detailed enough issue https://github.com/linkeddata/rdflib.js/issues/631

tests/unit/update-manager-test.js
bourgeoa: the remove date includes the tricky Collection.
see issue https://github.com/linkeddata/rdflib.js/issues/631

src/store.ts
jeff-zucker: **** I am not sure I understand this - if I create a file containing only `if(foo != null) console.log(1); ` I get `ReferenceError: foo is not **defined**`.  So I don't see how this can work.  I so much miss perl in this regard where you could just say if(foo) and that would cover null, undefined, and empty.

jeff-zucker:
AH! Thank you.  It works if the variable is declared but not defined but
does not work if the variable is not declared, makes sense.

jeff-zucker:
I ran the attached script on solidcommunity.net and it errored with `statement  to be removed is not in store`.  I got this error even though there was no collection so I don't think the collection is the problem.

The same script on the test server did not produce an error.  But it also did not remove all metadata about the document.  Even after the new removeDocument(), there were three statements left over.  The predicates related to the document that were not removed were acl, describedBy, and type - things from the link relations.  

I am not sure if it was intentional to leave these, but if not, it should be easy to fix.

```turtle

<script src="https://solidcommunity.net:8443/mashlib.js"></script>                                  
<script type="module">                                                                              
                                                                                                    
let doc = UI.rdf.sym("https://localhost:8443/public/s/test/tmpr/test.ttl");                         
                                                                                                    
async function test(){                                                                              
  await UI.store.fetcher.load(doc);                                                                 
  show('load,store');                                                                               
  await UI.store.removeDocument(doc);                                                                     
  show('load,store,removeDocument');;                                                               
}                                                                                                   
test();                                                                                             
                                                                                                    
function show(msg){                                                                                 
  let stmts  = UI.store.match(null,null,null,doc);                                                  
  console.log(`${stmts.length} statements with graph specified after ` + msg);                      
  stmts  = UI.store.match();                                                                        
  console.log(`${stmts.length} statements with graph not specified after ` + msg);                  
}                                                                                                   
                                                                                                    
</script>                                                                                           
```
The document in question looks like this:
```text
<#A> <#isa> <#B>.                                                                                   
<#C> <#isa> <#D>.                                                                                   
```
The script output was this (notice the 3 in the last one) : 
```text
2 statements with graph specified after load,store
30 statements with graph not specified after load,store
0 statements with graph specified after load,store,removeDocument
3 statements with graph not specified after load,store,removeDocument
```
```
```

jeff-zucker:
Also note that removeDocument did remove everything that had the document as the graph, but the three which were not  removed had the document as subject and a blank node as the graph.

bourgeoa:
> I ran the attached script on solidcommunity.net and it errored with `statement to be removed is not in store`. I got this error even though there was no collection so I don't think the collection is the problem.

The collection is in the metadata graph that is were the collection is an issue, not on the document graph

> Also note that `removeDocument` did remove everything that had the document as the graph, but the three which were not removed had the document as subject and a blank node as the graph.

Then these 3 statements are not related to either the graph document or the document part of the metadata graph
I shall look where these are created. If they may be removed. They make no harm for the time being

jeff-zucker (approved):
I'll raise a separate issue on the un-removed link relations.  Other than that, this looks good to go.

PR #652, opened 2024-07-10
Pass this.flags to `stringToN3`, fixes #650.

### LINKED ISSUE #650: `stringToN3` uses invalid delimiter for literals that contain quotes or newlines.
The serializer to NQuads (and turtle and others) uses the `stringToN3` function to convert the value of a string literal to a safe string in the output syntax.

However, the code uses the triple quote `"""` as delimiter for strings that contain a newline or a quote `"` character:

https://github.com/linkeddata/rdflib.js/blob/cfc0f8c437d7872d76dc384d686c28d5d6100e35/src/serializer.js#L584-L589

This is not a valid serialization for NQuads, see the literal production grammar:

https://www.w3.org/TR/n-quads/#grammar-production-literal

The `stringToN3` function should use the normal escaping using a backslash, i.e. `\n` and `\"`.

RinkeHoekstra:
Passing this.flags to `stringToN3` to ensure that format flags are properly used. Fixes #650

The `stringToN3` function resets to default flag 'e` if `flags` has no value. This fix ensures that any flags that were set prior to calling `stringToN3` are properly taken into account.

### DIFF
--- a/src/serializer.js
+++ b/src/serializer.js
@@ -553,7 +553,7 @@ export class Serializer {
               return expr.value === '1' ? 'true' : 'false'
           }
         }
-        var str = this.stringToN3(expr.value)
+        var str = this.stringToN3(expr.value, this.flags)
         if (expr.language) {
           str += '@' + expr.language
         } else if (!expr.datatype.equals(this.xsd.string)) {


RinkeHoekstra:
The `CI / npm-publish-build (pull_request)` failure has nothing to do with the change in the pull request. I think there's something off with the build workflow.

PR #655, opened 2024-08-16
Fix quotation marks inside code

Kibubu:
The code examples contained a mix of slanted, opening and closing quotations marks and could not be copy pasted into your playground.

Example: $rdf.sym(‘’https://example.com/alice/card”); 

Fixed by replacing all marks with either " or ' and fixed unmatched pairs.


angelo-v:
Thank you

PR #693, opened 2025-06-29
Skip rdfParse if script content is empty

### LINKED ISSUE #692: Fetching data islands in RDFa too zealous
Hi,

I have RDFa markup that has `<script>` tags that have a `type` that is `Parsable` but they have `src` attributes so there's nothing in the bodies. The following code thus blows up:

https://github.com/linkeddata/rdflib.js/blob/85a72c83072f37ce2b31a2545d3d70c51aa18d89/reference/fetcher-classes.js#L117

Would it be possible to add an option to `Fetcher.load` to turn this behaviour off? Or otherwise skip over it if there's an `src` attribute, or otherwise have it no-op or wrap in a `try` block or something?

csarven:
May be sufficient to resolve https://github.com/linkeddata/rdflib.js/issues/692 . There should be better checks before committing to parse<del>, e.g., type=text/html and application/xhtml+xml in a script block is not particularly common</del>.

### DIFF
--- a/reference/fetcher-classes.js
+++ b/reference/fetcher-classes.js
@@ -113,7 +113,7 @@ class XHTMLHandler {
       var scripts = this.dom.getElementsByTagName('script')
       for (var i = 0; i < scripts.length; i++) {
         var contentType = scripts[i].getAttribute('type')
-        if (Parsable[contentType]) {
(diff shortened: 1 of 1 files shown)


doriantaylor:
I didn't originally but I could I suppose; is there a guideline for adding tests somewhere?

bourgeoa (approved):

PR #694, opened 2025-07-11
JSON-LD Parser to TS

### LINKED ISSUE #602: jsonld parse do not return parsing errors
https://github.com/linkeddata/rdflib.js/blob/f94953b5e307f808fa3f437cb372b95d31dc638b/src/jsonldparser.js#L82

multimeric:
* Convert `jsonldparser` to TS
* Use async
* Add type annnotations
* Handle parsing errors (closes #602)


multimeric:
The `@list` processing also needs to be sorted out. Adding types makes `listToStatements` fail, because an object can't have both `@id` and `@list`.

Issue #709, opened 2025-09-20
JSON-LD parsing failes silently in browser context


angelo-v:
The parser fails with `TypeError: jsonld.flatten is not a function` when used via browser / esm (seems to work in nodejs as far as I can tell)

The error is hidden, due to issue #435, but still results in none of the RDF statements from the JSON-LD document being added to the store.

Issue #688, opened 2025-04-28
Support reading PeerTube application/activity+json


jg10-mastodon-social:
I managed to get use rdflib.js to read the JSON-LD in https://spectra.video/videos/watch/3a0159d9-4f12-4dd7-b025-afbb663cd19c

Here's what I ended up doing (from a SolidOS instance):
```js
let kb=panes.UI.store
panes.UI.store.fetcher.mediatypes["application/activity+json"]={}
panes.UI.store.fetcher.handlers[6].pattern=new RegExp("application/activity\\+json")
panes.UI.store.fetcher.load(kb.sym("https://spectra.video/videos/watch/3a0159d9-4f12-4dd7-b025-afbb663cd19c"),{force:true,headers:{},credentials: "omit"}).then(
 ()=>console.log(kb.match(kb.sym("https://spectra.video/videos/watch/3a0159d9-4f12-4dd7-b025-afbb663cd19c")))
)
```

Specifying the media type is required to adjust the accept header, otherwise PeerTube returns html.
The new accept header is `image/*;q=0.9, */*;q=0.1, application/rdf+xml;q=0.9, application/xhtml+xml, text/xml;q=0.5, application/xml;q=0.5, text/html;q=0.9, text/plain;q=0.5, text/n3;q=1.0, text/turtle;q=1, application/ld+json;q=0.9, application/activity+json`

Specifying the handler pattern is required so that load recognises the media type as parseable by the jsonldparser. To still support jsonld, this should probably be something like `pattern=new RegExp("application/(ld|activity)\\+json")`

I don't know why the options to load are needed but they seem to be.

jeswr:
I think the better push here is to get ActivityPub and PeerTube to respond with `application/activity+ld+json` content types.

nightpool:
Peertube should already respond to the LD JSON header IMO, based on my reading of the spec. There's no harm in including the activity+json header as well though

nightpool:
Ah, I see the issue. you're listing text/html and application/ld+json at the same priority level. So Peertube is going to respond with HTML to break the tie. Instead, you should put html at a lower priority level than ld+json

jg10-mastodon-social:
> Instead, you should put html at a lower priority level than ld+json

Can confirm that works:
```js
let kb=panes.UI.store
panes.UI.store.fetcher.mediatypes["text/html"]={q:0.8}
panes.UI.store.fetcher.handlers[6].pattern=new RegExp("application/activity\\+json")
[...]

Issue #435, opened 2020-08-09
jsonldParser fails to report or handle parse errors


josephguillaume:
On error, `callback` is called (https://github.com/linkeddata/rdflib.js/blob/master/src/jsonldparser.js#L97) but this is actually `executeCallback` (https://github.com/linkeddata/rdflib.js/blob/master/src/parse.ts#L55), which does not do any error handling (https://github.com/linkeddata/rdflib.js/blob/master/src/parse.ts#L81)

It appears the fix is to simple delete `.catch(callback)` to allow the error to be caught by `executeErrorCallback` (https://github.com/linkeddata/rdflib.js/blob/master/src/parse.ts#L66)

angelo-v:
I came to the same conclusion

Issue #426, opened 2020-06-27
headers['accept'] does not work on Headers object?


josephguillaume:
The problem I ran into was that fetch was always sending `Accept=*/*`, which meant I was receiving html instead of ttl.

I traced this to the statement 
https://github.com/linkeddata/rdflib.js/blob/14c93f5cd18fd4658561f243c2f4b3f33a9216df/src/fetcher.ts#L1045

Apparently for a Headers object, we have to use `headers.append`, or alternatively just use a plain object?

This seems to be the case in Chrome 83.0.4103.116 (Official Build) (64-bit) (cohort: Stable) and Brave | 1.4.95 Chromium: 80.0.3987.122 (Official Build) (64-bit)

A workaround when using solid is to specify a headers object, e.g.
`fetcher.load(folder, {headers: { accept: "text/turtle" }});`
This then means that a `Headers` object is not used: 
https://github.com/linkeddata/rdflib.js/blob/14c93f5cd18fd4658561f243c2f4b3f33a9216df/src/fetcher.ts#L1034

PR #710, opened 2025-09-20
Fix/jsonld parser

angelo-v:
fixes several JSON-LD related issues:

- [x] https://github.com/linkeddata/rdflib.js/issues/709
- [x] https://github.com/linkeddata/rdflib.js/issues/435
- [x] https://github.com/linkeddata/rdflib.js/issues/688
- [x] https://github.com/linkeddata/rdflib.js/issues/426
- [x] https://github.com/linkeddata/rdflib.js/issues/711

### DIFF
--- a/src/parse.ts
+++ b/src/parse.ts
@@ -54,7 +54,7 @@ export default function parse (
     } else if (contentType === JSONLDContentType) {
       // since we do not await the promise here, rejections will not be covered by the surrounding try catch
       // we do not use await, because parse() should stay sync
-      // so, to not lose the async error, we need to catch the rejection and call the error callback here additionally
+      // so, to not lose the async error, we need to catch the rejection and call the error callback here too
       jsonldParser(str, kb, base)
           .then(executeCallback)
           .catch(executeErrorCallback)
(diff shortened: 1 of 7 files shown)


src/parse.ts
```diff
@@ -17,7 +17,7 @@ type CallbackFunc = (error: any, kb: Formula | null) => void
  * Parse a string and put the result into the graph kb.
  * Normal method is sync.
  * Unfortunately jsdonld is currently written to need to be called async.
- * Hence the mess below with executeCallback.
+ * If you are parsing JSON-LD and want to no when and whether it succeeded, you need to use the callback param.
```
TallTed, line 20:
```suggestion
 * If you are parsing JSON-LD and want to know when and whether it succeeded, you need to use the callback param.
```

tests/unit/parse-test.js
```diff
@@ -126,7 +126,7 @@ describe('Parse', () => {
               ].
 
 `))
-/*      // replaced by above expect due to inpredictable order in store.statements 
+/*      // replaced by above expect due to inpredictable order in store.statements
```
TallTed, line 129:
```suggestion
/*      // replaced by above expect, due to unpredictable order in store.statements
```

jeff-zucker (approved):
I tested on Linux Mint with Chrome  140.0.7339.207 and can confirm that using rdflib main silently fails to parse JSON-LD in the browser but correctly parses in node.  After applying this PR, parsing JSON-LD succeeded in both contexts.

Issue #228, opened 2018-04-17
XML qname parses correctly but is not serialized correctly.


jamsden:
Error: Invalid character "." cannot be in XML qname for URI: http://jazz.net/xmlns/prod/jazz/rtc/ext/1.0/com.ibm.team.apt.attribute.acceptance
at Serializer.qnameMethod (/Users/jamsden/GitHub/rdflib.js/lib/serializer.js:913:17)
at Serializer.subjectXMLTreeMethod (/Users/jamsden/GitHub/rdflib.js/lib/serializer.js:819:13)
at Serializer.statementListToXMLTreeMethod (/Users/jamsden/GitHub/rdflib.js/lib/serializer.js:736:22)
at Serializer.statementsToXML (/Users/jamsden/GitHub/rdflib.js/lib/serializer.js:931:16)
at Object.serialize (/Users/jamsden/GitHub/rdflib.js/lib/serialize.js:27:29)

According to https://www.w3.org/TR/REC-xml/#NT-Name, NCName can contain ".". So the serializer should handle these QNames.

jamsden:
For the syntax of a QName, https://www.w3.org/TR/REC-xml/ says that a local name for a qname is a Name less colon. A Name is defined here: https://www.w3.org/TR/REC-xml/#NT-Name. There's almost nothing illegal in that set of characters except colon, control characters, and unassigned code points (and as far as I can see the latter are merely discouraged, not illegal).

jamsden:
I see this is still open.  https://www.w3.org/TR/xml-names11/#NT-NCName explicitly states that "." is included in NCName which is part of a Prefix. 

The fix is simple: In rdflib/src/serializer.js:196:
Change:
  __Serializer.prototype._notQNameChars = '\t\r\n !"#$%&\'()*.,+/;<=>?@[\\]^`{|}~';
to:
  __Serializer.prototype._notQNameChars = '\t\r\n !"#$%&\'()*,+/;<=>?@[\\]^`{|}~';
(i.e., remove . from the _notQNameChars)

Any chance this could get addressed?

bourgeoa:
Thanks for reopening.
But this other consequences that make some tests fail on serialize text/turtle
see https://github.com/linkeddata/rdflib.js/pull/686

jamsden:
Understood. But given that, is this something that could be addressed in a future release or do I need to fork?

bourgeoa:
@TallTed thanks I shall try to remember.

PR #718, opened 2025-11-04
parse and serialize with dot in term name

### LINKED ISSUE #601: turtle parsing not correctly handling `.` in suffix
As discussed in [this thread](https://forum.solidproject.org/t/mashlib-databrowser-hl7-fhir/6118/5); terms like `ex:a.b` appear to be valid according to the [definition of `PN_LOCAL` in the Turtle Grammar](https://www.w3.org/TR/turtle/#grammar-production-PN_LOCAL). However it appears that rdflib is interpreting this as `ex:a . b` rather than as a single term.

The terms are correctly parsed by `N3.js` but not by the custom parser in this library.

---

It might be worth considering migrating to use [rdf-parse](https://www.npmjs.com/package/rdf-parse) & [rdf-serialize](https://www.npmjs.com/package/rdf-serialize) instead of the custom parsers; given that those packages use parser/serializers that are 100% passing the spec tests, and most are also currently having RDF-star & RDF 1.2 support added to them.

bourgeoa:
# Fix Turtle parser/serializer handling of dots in terms

## Summary
Fixes #601 #228 - Parser and serializer now correctly handle dots within local names per Turtle 1.1 specification.

## Problem
The Turtle parser incorrectly treated dots (`.`) inside local names as path operators or statement terminators, breaking valid Turtle like:
```turtle
@prefix ex: <http://example.com/> .
ex:subject.example ex:pred ex:obj .
[] :loves [] .
```

The serializer also failed to produce spec-compliant abbreviated forms, outputting `<http://example.com/subject.example>` instead of the valid `ex:subject.example`. It also occasionally created a spurious `loc:` prefix for base‑relative IRIs like `</results.ttl>`.

## Solution

**Parser (`src/n3parser.js`)**
- Allow dots in names unless followed by whitespace/comment/EOF
- Added `dotTerminatesName(str, i)` helper to centralize logic across 3 call sites
- Shared `wsOrHash` regex at module level for performance
- Clarified why `node()` leaves `.` for unified `checkDot()` handling

**Serializer (`src/serializer.js`)**
- Implemented `isValidPNLocal(local)` validator per Turtle 1.1 spec
- Produces abbreviated qnames with dots: `ex:subject.example` ✅
- Rejects trailing dots: `ex:subject.` → `<http://example.com/subject.>` ✅
- Allows empty local names: `ex:` for URIs ending in `/` or `#` ✅
- Added minimum namespace length check to prevent splitting at protocol
- Avoid abbreviating when namespace equals the document's base directory (prevents accidental `loc:` with `</>`) ✅
- New optional flag `o`: do not abbreviate to a prefixed name when the local part contains a dot (opt‑out of dotted locals)
  - Usage (Turtle): `serialize(doc, kb, doc.value, 'text/turtle', undefined, { flags: 'o' })`
  - Notes: user flags are merged with defaults for Turtle/JSON‑LD; `'p'` still disables all QName abbreviation

**Tests (`tests/unit/dot-in-term-test.ts`)**
- Parse dots in local names: `ex:subject.example`
- Parse blank node terminators: `[] :loves [] .`
- Serialize with dots: `ex:subject.example` (not `<URI>`)
- Reject trailing dots in serialization
- Empty local names for URIs ending in `/`
- Honors 'o' flag: dotted locals fall back to `<IRI>` instead of `prefix:local`

## Results
- Spec‑compliant default output: abbreviates dotted locals when valid.
- Opt‑out available via 'o' flag for conservative output.
- Serialize fixtures generate expected outputs with the harness using 'o'.

## Behavior Change
**Before:** Conservative approach — URIs with dots serialized as `<IRI>`.
**After (default):** Spec-compliant — valid qnames with dots abbreviated (e.g., `ex:subject.example`).
**Opt-out:** Pass `'o'` in `flags` to keep dotted locals as `<IRI>`.

This produces more compact, spec‑correct Turtle by default while offering a simple opt‑out for legacy expectations.

## Files Changed
- `src/n3parser.js` — Parser fix, helper, shared regex
- `src/serializer.js` — PN_LOCAL validator; improved splitting logic; base‑dir guard; `'o'` flag support
- `src/serialize.ts` — Merge user flags with defaults so `'o'` is honored in Turtle/JSON‑LD
- `tests/unit/dot-in-term-test.ts` — Added tests including `'o'` flag case
- `tests/serialize/data.js` — Pass `'o'` for Turtle fixture generation; preserve RDF/XML behavior
- `README.md` — Document serializer flags and `'o'` usage
- `changes.txt` — Changelog entry
- `lib/*` — Transpiled updates kept in sync (for repo consumers)

## References
- Issue: #601
- Spec: [W3C Turtle 1.1 - PN_LOCAL](https://www.w3.org/TR/turtle/#grammar-production-PN_LOCAL)

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ npm install --save rdflib
 
 ## Serializer flags
 
-The Turtle/N3/JSON‑LD serializers accept an optional flags string to tweak output formatting and abbreviation behavior.
+The Turtle/N3/JSON‑LD serializers accept an optional `flags` string to tweak output formatting and abbreviation behavior.
 
 - Pass flags via the options argument to `serialize(...)`:
 


README.md
```diff
@@ -63,6 +63,36 @@ installed.
 npm install --save rdflib
 ```
 
+## Serializer flags
+
+The Turtle/N3/JSON‑LD serializers accept an optional flags string to tweak output formatting and abbreviation behavior.
```
TallTed:
```suggestion
The Turtle/N3/JSON‑LD serializers accept an optional `flags` string to tweak output formatting and abbreviation behavior.
```

src/serialize.ts
```diff
@@ -72,15 +72,17 @@ export default function serialize (
         return executeCallback(null, documentString)
       case TurtleContentType:
       case TurtleLegacyContentType:
-        sz.setFlags('si') // Suppress = for sameAs and => for implies
+        // Suppress = for sameAs and => for implies; preserve any user-specified flags (e.g., 'o')
+        sz.setFlags('si' + (opts.flags ? (' ' + opts.flags) : ''))
         documentString = sz.statementsToN3(newSts)
         return executeCallback(null, documentString)
       case NTriplesContentType:
         sz.setFlags('deinprstux') // Suppress nice parts of N3 to make ntriples
```
TallTed: Should this be `deinprstux` or `d e i n p r s t u x`?

src/serializer.js
```diff
@@ -255,6 +261,29 @@ export class Serializer {
   _notNameChars =
     (this._notQNameChars + ':')
 
+  // Validate if a string is a valid PN_LOCAL per Turtle 1.1 spec
+  // Allows dots inside the local name but not as trailing character
+  // Also allows empty local names (for URIs ending in / or #)
+  isValidPNLocal(local) {
+    // Empty local name is valid (e.g., ex: for http://example.com/)
```
TallTed: Note that resolution of this "empty" local name is determined server-side. It often defaults to `index.html`, but not always. This is configurable in Apache and some (most? all?) other HTTP servers, so it cannot be relied upon without out-of-band communications. Configured local name can be things like `index.php`.

TallTed:
> @linkeddata deleted a comment from @TallTed [7 hours ago](https://github.com/linkeddata/rdflib.js/pull/718#event-21476471760)

What was that comment? And why was it deleted, rather than responded to?

PR #758, opened 2026-01-15
flatten for 

bourgeoa:
Handle both ESM (browser) and CommonJS (Node.js) module formats
  Browser ESM: jsonld.default.flatten
  Node.js CommonJS: jsonld.flatten

### DIFF
--- a/src/jsonldparser.js
+++ b/src/jsonldparser.js
@@ -69,9 +69,11 @@ export default async function jsonldParser(str, kb, base) {
     : base
 
   const jsonld = await import('jsonld')
-  // ⚠ Unit tests also work without accessing `jsonld.default` explicitly, but real browser usage will fail with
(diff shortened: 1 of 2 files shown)


src/jsonldparser.js
```diff
@@ -69,9 +69,11 @@ export default async function jsonldParser(str, kb, base) {
     : base
 
   const jsonld = await import('jsonld')
-  // ⚠ Unit tests also work without accessing `jsonld.default` explicitly, but real browser usage will fail with
-  // just calling `jsonld.flatten`, so please do not remove `default`
```
angelo-v, line 73: I prefer to keep that warning, even if now both is handled, no test is stopping someone from removing the default variant.

bourgeoa: done thanks

PR #775, opened 2026-03-14
Npm publish only on semver (stable and prerelease)

bourgeoa:
- publish to npm on stable semver like actually
- remove npm publish with hash
- add npm publish on semver `new` prerelease version : Check is made from package.json version and npm
  - Pre-release can be created with :
    - `npm version prerelease`
    - or `npm version prerelease --preid <any>`

  - prerelease is tagged with `dev`. Last published prerelease : `rdflib@dev`


timea-solid (approved):
Looks good to me and a great improvement towards better testing options

Issue #741, opened 2025-12-26
Expose link rel statements also as http://www.iana.org/assignments/relation/


jg10-mastodon-social:
As mentioned in #solid_solidos:gitter.im https://matrix.to/#/!wAfwwonbRnRLYejFOR%3Agitter.im/%24iTplbPyDpPTSZanbOv4MHf8ZPrN8_OdlmJmSpjTzxD8 link header rel predicates are defined in http://www.iana.org/assignments/relation/

alternate, seeAlso, meta, describedby are currently grouped as `rdfs:seeAlso`
https://github.com/linkeddata/rdflib.js/blob/f6d21ba405c7db773697aeb7c10885bed49b8b83/src/fetcher.ts#L1343-L1357

It would be useful to include their original predicate too, i.e. remove the if else structure.

PR #809, opened 2026-06-09
iana link relations

angelo-v:
Implements #741 

Disclosure: The PR has been created with support from LLM and Coding Agents. Details about Agent vs. Human contributions are disclosed in commit messages. I have used these technologies with close oversight and steering


melvincarvalho (approved):
LGTM. Additive, it emits the specific IANA link-relation predicate alongside the existing rdfs:seeAlso, so existing consumers are unaffected while the precise relation becomes available.

Issue #823, opened 2026-07-03
bulk maintenance plan


jeswr:
Hi hi (this is Jesse with his opensourcer / in his free time hat on) - I have been making use of Fable now that it has been online to do bulk triage and bug fixing on several of my libraries. For those that I maintain with others I have been following a practise of:
 - Getting it to draft a triage plan such as https://gist.github.com/jeswr/f48f7f8337f32b125a512bf122fee5fa. Signing off on it with any comments I might have; having it then action closing issues etc.
 - Having it open draft PRs for me to review. When I am happy with them I will mark them / would be willing to submit them as if they were my own. Then I mark them as ready to review and tag the rest of the maintainers.

I was planning to do this on rdflib.js. Please flag if you have questions or concerns. My aim is not to create work for people, especially near the weekend. Please note if I am tagging you for a review there is no expectation of a fast response (or one at all for that matter).

PR #825, opened 2026-07-03
fix: correct `this.keyords` typo breaking the 'k' serializer flag (#742)

### LINKED ISSUE #742: Serializer contains misspelling of 'this.keyords' --> 'this.keywords'
In the serializer `symbolToN3` has a miss spelling of `keywords`, breaking the 'k' flag. 
https://github.com/linkeddata/rdflib.js/blob/1b386c4780f56aaf3939ef937d9c2b1d0b775de6/src/serializer.js#L682-L682

I have fixed this in my refactor of serializer see https://github.com/linkeddata/rdflib.js/issues/740

I do have question about the 'k' flag, what is it used for since it is not allowed in the n3/ttl syntax

jeswr:
> **Note:** This PR is agent-generated as part of the agent-assisted triage described in #823, and is awaiting review by @jeswr. Please don't merge until he has signed off — it will stay a draft until then.

Fixes #742

## The bug

`Serializer#symbolToN3` referenced `this.keyords` (typo) instead of `this.keywords` (defined in the constructor at `src/serializer.js:39`). This is the code path for the `'k'` (keywords) serializer flag, so any serialization using that flag together with a default namespace crashed.

## Reproduction

```js
const { graph, sym, Serializer } = require('rdflib')

const serializer = Serializer(graph())
serializer.setFlags('k')
serializer.defaultNamespace = 'http://example.org/ns#'
serializer.symbolToN3(sym('http://example.org/ns#foo'))
// TypeError: Cannot read properties of undefined (reading 'indexOf')
```

## The fix

One word in `src/serializer.js`: `this.keyords` → `this.keywords`. With the fix, the snippet above returns `"foo"` (bare word, as the `'k'` flag intends), and keywords such as `a` keep their colon (`:a`) so they can't be confused with the `a` keyword.

## Tests

Added two regression tests to `tests/unit/serializer-test.ts` covering the `'k'` flag path (bare word for non-keywords, `:a` for keywords). Verified they fail with the exact `TypeError` on the unfixed code and pass with the fix.

Local run: `npm run lint` (0 errors; 5 pre-existing warnings unchanged), `npm run build`, and the full `npm test` (306 unit tests + 18 serialize round-trips + type checks) all pass.

---

@jeswr — over to you when you have a moment; no rush.

### DIFF
--- a/src/serializer.js
+++ b/src/serializer.js
@@ -679,7 +679,7 @@ export class Serializer {
         if (this.defaultNamespace && this.defaultNamespace === namesp &&
             this.flags.indexOf('d') < 0) { // d -> suppress default
           if (this.flags.indexOf('k') >= 0 &&
-            this.keyords.indexOf(localid) < 0) {
+            this.keywords.indexOf(localid) < 0) {
             return localid
           }
           return ':' + localid
--- a/tests/unit/serializer-test.ts
+++ b/tests/unit/serializer-test.ts
@@ -1,5 +1,5 @@
 import { expect } from "chai";
-import { graph, Serializer } from "../../src/index";
+import { graph, Serializer, sym } from "../../src/index";
 
 describe("Serializer", () => {
 	describe("can make up prefixes", () => {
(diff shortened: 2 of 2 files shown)


jeswr:
LGTM

Issue #824, opened 2026-07-03
Triage plan (agent-assisted) — for @jeswr sign-off


jeswr:
> **Note: this issue was generated by an agent** (Jesse's Fable assistant), acting per the process described in #823. It is a **proposal only** — nothing gets closed, merged, or changed until @jeswr (and any co-maintainer who wants to weigh in) signs off. If you're reading this near a weekend: there is absolutely no expectation of a response, fast or otherwise.

Hi! Following up on #823, I went through **all 212 open issues and 34 open PRs** and this is the consolidated triage plan for sign-off. Tick a box to approve that line, strike it out or comment to veto — partial sign-off is fine, and anything unticked simply stays as it is.

**How items were assessed:** items marked *runtime-verified* were reproduced (or confirmed fixed) against rdflib v2.4.0 with small scripts; the rest were assessed by reading the thread plus the current source. If any proposed closure looks wrong, that's exactly what this review step is for.

**Summary:** propose closing **110** (21 already fixed, 77 stale/obsolete/answered, 12 duplicates) · **5** need info from the reporter · **11** quick fixes · **49** actionable bugs · **32** enhancements · **5** docs · **34** PRs dispositioned · plus a modernization section at the end.

---

### 1a. Already fixed — close with a short "fixed in current release" comment (21)

- [ ] #20 — RDFa as a node module? — *RDFa parser has long been bundled in the npm package*
- [ ] #41 — node compatibility error — *parser now uses @xmldom/xmldom in node; the jsdom-era error is gone*
- [ ] #42 — is there a link for documentation on how to use this? — *README now has a Documentation section linking API docs and three tutorials*
- [ ] #49 — Create Releases with minified versions for greater adoption — *regular GitHub releases + npm dist with browser bundle now exist (v2.4.0)*
- [ ] #56 — parse function invokes unknown function (see pull request) — *verified in src/parse.ts: sparqlUpdateParser typo is gone*
- [ ] #67 — jsdom 4.x onward only works on io.js — *jsdom is no longer a dependency; @xmldom/xmldom is used*
- [ ] #173 — termType values case confusion — *termType values are now RDF/JS-style initial caps; the migration completed*
- [ ] #183 — Serializing more than 2 namespaces in rdf+xml — *runtime-verified at v2.4.0: makeUpPrefix rewritten, all suggested prefixes preserved*
- [ ] #215 — RDF+XML serializer never produces typed node elements — *per thread, fixed by PR #471 (typed node elements)*
[...]

Issue #147, opened 2016-10-18
Boolean literals parsed as "0", "1"


dan-f:
Boolean literals are getting parsed as correctly having the boolean `datatype`, but (I argue) incorrectly having the value `"0"` or `"1"`, which I find to be a leaky abstraction.

I'd fix this myself, but I cannot understand how the n3 parser is implemented. I assume it's running some sort of DFA, but I'm finding the code difficult to understand.

One curiosity is that [the `BOOLEAN_DATATYPE` constant is not being used by the parser](https://github.com/linkeddata/rdflib.js/blob/master/src/n3parser.js#L202).

See attached screenshot for what I'm referring to.

![rdflib_boolean_literal](https://cloud.githubusercontent.com/assets/2146312/19491449/ffceffaa-9540-11e6-8067-a3e73aed7532.png)

timbl:
The parser is a standard top-down recursive descent parser.   (It was written i python and transcoded using pyjs aka  pyjamas)

 Results are pushed onto a result stack [res] in the style of python parser history.   
Functions like commaSeparatedList = function(str, j, res, ofUris) parse the production in the grammar with their name.  Here str is the whole string to be parsed, j is the pointer, res an array to push result values onto, and ofUris just a flag to allow this code to cover two cases.
It returns a new value of the pointer or  < 0 if the string does not match the grammar.

The function nodeOrLiteral parses a node or or literal value syntax.
The first line 
```
    var j = this.node(str, i, res);
    if ((j >= 0)) {
        return j;
    }
```
tried named node syntax ( <fooo#bar>    ex:bar etc)  and if that fails t assumes literal.
In fact node() covers the true and false cases:

```
    var j = this.tok("true", str, i);
    if ((j >= 0)) {
        res.push(true);
        return j;
    }
    var j = this.tok("false", str, i);
    if ((j >= 0)) {
        res.push(false);
        return j;
    }
```
at line 847.  
Those raw boolean values are what get passed to makeStatement and store.add()
IndexedFormula.add() calls Node.fromValue() on each part of the quad whcih calls Literal.fromValue which ends up doing this in literal.js

```
  /**
   * @method fromBoolean
   * @static
   * @param value {Boolean}
   * @return {Literal}
   */
  static fromBoolean (value) {
    let strValue = value ? '1' : '0'
    return new Literal(strValue, null, XSD.boolean)
  }
```
at line 64.

[...]

PR #827, opened 2026-07-03
fix: Turtle serializer abbreviates typed literals without validating lexical forms (#772)

### LINKED ISSUE #772: Turtle serializer mis‑normalizes xsd:boolean literal "true" as "false"
When using rdflib.js to create xsd:boolean literals and serialise them to Turtle, boolean literals with lexical form "true" are incorrectly serialised as "false". This appears to be a bug in the boolean normalisation logic used by the Turtle serializer.

According to the XSD and RDF specs for xsd:boolean, the lexical space is:

"true" → value true

"false" → value false

"1" → value true

"0" → value false

In the current rdflib.js behaviour, "true" is being normalised to the lexical form "false" on serialisation, which is not spec‑compliant.

## Reproducible example

```js
const $rdf = require('rdflib')

const XSD = $rdf.Namespace('http://www.w3.org/2001/XMLSchema#')

const store = $rdf.graph()
const subject = $rdf.sym('http://example.org/s')
const predicate = $rdf.sym('http://example.org/p')

// Four literals with xsd:boolean datatype and different lexical forms
const litTrue  = $rdf.literal('true',  null, XSD('boolean'))
const litFalse = $rdf.literal('false', null, XSD('boolean'))
const litOne   = $rdf.literal('1',     null, XSD('boolean'))
const litZero  = $rdf.literal('0',     null, XSD('boolean'))

store.add(subject, predicate, litTrue)
store.add(subject, predicate, litFalse)
store.add(subject, predicate, litOne)
store.add(subject, predicate, litZero)

$rdf.serialize(
  undefined,           // base
  store,
  'http://example.org/',
  'text/turtle',
  (err, ttl) => {
    if (err) {
      console.error(err)
      return
    }
    console.log(ttl)
  }
)

```

On my system, the relevant part of the Turtle output looks like:
```
<http://example.org/s> <http://example.org/p> false, false, true, false .
```

jeswr:
> **🚧 DRAFT — Note: this is an agent-generated draft PR** (part of the agent-assisted triage announced in #823). It is for @jeswr to review — please do not merge or mark ready-for-review until he has signed off.

Fixes #772. Part of the triage programme tracked in #824.

## Problem

`atomicTermToN3` (`src/serializer.js`) abbreviates `xsd:integer` / `xsd:decimal` / `xsd:double` / `xsd:boolean` literals to native Turtle tokens based only on the datatype URI, without checking that the lexical form is valid for that datatype. Originally reported for booleans in #772, generalized to the whole class after review:

| Input literal | Before | After |
|---|---|---|
| `"true"^^xsd:boolean` | `false` — **value flipped** (the #772 report) | `true` |
| `"false"` / `"1"` / `"0"` `^^xsd:boolean` | `false` / `true` / `false` | unchanged |
| `"yes"^^xsd:boolean` (invalid) | `false` — value lost | `"yes"^^xsd:boolean` |
| `"abc"^^xsd:integer` (invalid) | bare `abc` — invalid Turtle, datatype lost | `"abc"^^xsd:integer` |
| `"1.2.3"^^xsd:decimal` (invalid) | `1.2.3` — invalid Turtle | `"1.2.3"^^xsd:decimal` |
| `"2."^^xsd:decimal` (valid XSD; Turtle needs a digit after the dot) | `2.` — invalid Turtle | `2.0` |
| `"NaN"` / `"INF"` / `"-INF"` `^^xsd:double` (valid XSD; no Turtle DOUBLE form) | `NaN.0e0` etc. — mangled | `"NaN"^^xsd:double` etc. |
| `"abc"^^xsd:double` (invalid) | `abc.0e0` — mangled | `"abc"^^xsd:double` |

## Fix

Each of the four abbreviation cases now abbreviates only when the lexical form is valid for the datatype **and** expressible as a Turtle token:

- `xsd:integer` — `/^[+-]?[0-9]+$/`
- `xsd:decimal` — `/^[+-]?(?:[0-9]+\.?[0-9]*|\.[0-9]+)$/`, then normalized so the token has a digit after the dot (`5` → `5.0`, `2.` → `2.0`)
- `xsd:double` — `/^[+-]?(?:[0-9]+\.?[0-9]*|\.[0-9]+)(?:[eE][+-]?[0-9]+)?$/` (excludes `INF`/`-INF`/`NaN`), then the existing `.`/`e` normalization
- `xsd:boolean` — `true`/`1` → `true`, `false`/`0` → `false`

Anything else falls through to the existing verbose `"value"^^datatype` path, so serialization is lossless and always emits valid Turtle. Diff: `src/serializer.js` +17/−1; everything else is tests.

## Ecosystem backward-compatibility assessment

- **Serializer-only** — parsing is untouched (read lenient, write strict).
- **Valid, correctly-serialized forms are byte-identical.** No `tests/serialize/*-ref.*` golden fixture changes; the full fixture corpus serializes identically (CI `test:serialize` green on this head).
- **Output changes only where the previous output was wrong** — the table rows above: a flipped value, lost data, or non-conformant Turtle. Every changed output is valid Turtle and round-trips with value and datatype preserved.
- **Who the #772 flip actually hit:** booleans stored with the standard lexical form `true` — e.g. minted by applications via `lit('true', undefined, XSD.boolean)`, or imported from RDF/JS-ecosystem libraries (N3.js et al.) that keep the canonical lexical form. Booleans from rdflib's own Turtle parser or `Literal.fromBoolean` are stored as `'1'`/`'0'` and always serialized correctly, so SolidOS/NSS round-trips through rdflib's own parser are unaffected.
- Related: #147 (the parser normalizes `true`/`false` to `'1'`/`'0'` on parse) is untouched here — but both lexical conventions now serialize to correct tokens.
- **Semver:** no API change; suggest `release-patch` (pure bug fix), though `release-minor` is defensible if any serializer-output delta is treated as behavioural.

## Tests

Extends the existing `tests/unit/serialize-test.js` (no new file):

- booleans: valid forms (`true`/`false`/`1`/`0`) unchanged; invalid forms (`yes`, `TRUE`, `2`, `""`, `01`) serialize verbosely
- new integer/decimal/double blocks: valid forms abbreviate exactly as before; invalid forms serialize verbosely
- serialize→parse round-trip tests asserting value + datatype are preserved for every invalid lexical form (no data loss)

CI is green on this head (Node 22.x / 24.x: unit + serialize golden fixtures + type tests + typedoc).

## Relationship to #830

#830 (term-level serialization rework) supersedes this PR and subsumes this fix at its term layer. This PR stands as the minimal, self-contained alternative if the larger rework is not taken — whichever lands, the other should be closed or rebased accordingly.

---
@jeswr — over to you for review; leaving as draft.

### DIFF
--- a/src/serializer.js
+++ b/src/serializer.js
@@ -577,7 +577,7 @@ export class Serializer {
             }
 
             case 'http://www.w3.org/2001/XMLSchema#boolean':
-              return expr.value === '1' ? 'true' : 'false'
+              return expr.value === '1' || expr.value === 'true' ? 'true' : 'false'
           }
         }
         var str = this.stringToN3(expr.value, this.flags)
--- a/src/serializer.js
+++ b/src/serializer.js
@@ -546,6 +546,11 @@ export class Serializer {
   }
   // //////////////////////////////////////////// Atomic Terms
 
+  // Valid lexical forms that can be abbreviated to native Turtle tokens
+  validInteger = new RegExp(/^[+-]?[0-9]+$/)
+  validDecimal = new RegExp(/^[+-]?(?:[0-9]+\.?[0-9]*|\.[0-9]+)$/)
+  validDouble = new RegExp(/^[+-]?(?:[0-9]+\.?[0-9]*|\.[0-9]+)(?:[eE][+-]?[0-9]+)?$/)
+
   //  Deal with term level things and nesting with no bnode structure
   atomicTermToN3 (expr, stats) {
     switch (expr.termType) {
@@ -558,17 +563,25 @@ export class Serializer {
           throw new TypeError('Value of RDF literal node must be a string')
         }
         // var val = expr.value.toString() // should be a string already
+        // Only abbreviate to native Turtle syntax when the lexical form is valid
+        // for the datatype AND expressible as a Turtle token; otherwise fall
+        // through to the verbose "value"^^datatype form so no data is lost.
         if (expr.datatype && this.flags.indexOf('x') < 0) { // Supress native numbers
           switch (expr.datatype.uri) {
 
             case 'http://www.w3.org/2001/XMLSchema#integer':
+              if (!this.validInteger.test(val)) break // Invalid lexical form: serialize verbosely
               return val
 
             case 'http://www.w3.org/2001/XMLSchema#decimal': // In Turtle, must have dot
+              if (!this.validDecimal.test(val)) break // Invalid lexical form: serialize verbosely
               if (val.indexOf('.') < 0) val += '.0'
+              else if (val.charAt(val.length - 1) === '.') val += '0' // Turtle needs a digit after the dot
               return val
 
             case 'http://www.w3.org/2001/XMLSchema#double': {
+              // INF, -INF and NaN are valid xsd:double but have no native Turtle form
+              if (!this.validDouble.test(val)) break // Invalid lexical form: serialize verbosely
               // Must force use of 'e'
               const eNotation = val.toLowerCase().indexOf('e') > 0
               if (val.indexOf('.') < 0 && !eNotation) val += '.0'
@@ -577,7 +590,9 @@ export class Serializer {
             }
 
             case 'http://www.w3.org/2001/XMLSchema#boolean':
-              return expr.value === '1' || expr.value === 'true' ? 'true' : 'false'
+              if (val === 'true' || val === '1') return 'true'
+              if (val === 'false' || val === '0') return 'false'
+              break // Invalid lexical form: serialize verbosely
           }
         }
         var str = this.stringToN3(expr.value, this.flags)
--- a/tests/unit/serialize-test.js
+++ b/tests/unit/serialize-test.js
@@ -240,6 +240,42 @@ example:subject schema2:predicate 123e-2 .
 
     })
 
+    describe('booleans', () => {
+        const serializeBoolean = (lexicalForm) => {
+            const doc = sym("https://example.net/doc")
+            const statement = st(
+                sym('https://subject.example'),
+                sym('https://predicate.example'),
+                lit(lexicalForm, undefined, sym("http://www.w3.org/2001/XMLSchema#boolean")),
+                doc
+            )
+            const kb = graph()
+            kb.add(statement)
+            return serialize(doc, kb, null, 'text/turtle')
+        }
+        const expected = (token) => `@prefix : </doc#>.
+
+<https://subject.example> <https://predicate.example> ${token}.
+
+`
+
+        it('lexical form "true" serializes to true', () => {
+            expect(serializeBoolean('true')).to.equal(expected('true'))
+        })
+
+        it('lexical form "false" serializes to false', () => {
+            expect(serializeBoolean('false')).to.equal(expected('false'))
+        })
+
+        it('lexical form "1" serializes to true', () => {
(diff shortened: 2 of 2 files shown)


src/serializer.js
```diff
@@ -577,7 +577,7 @@ export class Serializer {
             }
 
             case 'http://www.w3.org/2001/XMLSchema#boolean':
-              return expr.value === '1' ? 'true' : 'false'
+              return expr.value === '1' || expr.value === 'true' ? 'true' : 'false'
```
jeswr: This is also buggy. There may be invalid booleans which are not '0' or 'false' in which case the verbose serialisation is needed. It looks like most datatype serialisation has similar bugs.

jeswr: Fixed in c13c99c, and broadened to the whole class you pointed at (agent-generated change, for @jeswr to review).

`atomicTermToN3` now abbreviates integer/decimal/double/boolean literals to native Turtle tokens only when the lexical form is valid for the datatype **and** expressible as a Turtle token; everything else `break`s out of the switch and falls through to the existing verbose `"value"^^datatype` path, so no value is silently lost and no invalid Turtle is emitted. Examples now fixed:

- `"yes"^^xsd:boolean` → `"yes"^^xsd:boolean` (was `false` — value lost)
- `"abc"^^xsd:integer` → `"abc"^^xsd:integer` (was bare `abc` — invalid Turtle)
- `"1.2.3"^^xsd:decimal` → verbose (was `1.2.3` — invalid Turtle)
- `"NaN"`/`"INF"`/`"-INF"^^xsd:double` → verbose (valid xsd:double, but not a Turtle DOUBLE token; was `NaN.0e0`)
- `"2."^^xsd:decimal` (valid XSD, invalid Turtle) now normalizes to `2.0` instead of emitting `2.`

Added per-datatype native/verbose tests plus serialize→parse round-trip tests asserting value+datatype are preserved for all the invalid forms.

jeswr:
> **Note: agent-generated update, for @jeswr to review.**

Addressed the review comment by broadening the fix from booleans to the whole class of datatype-abbreviation bugs: in `atomicTermToN3`, native Turtle abbreviation of `xsd:integer` / `xsd:decimal` / `xsd:double` / `xsd:boolean` literals is now guarded by lexical-form validation. Invalid (or Turtle-inexpressible) lexical forms fall through to the verbose `"value"^^datatype` serialization — no data loss, no invalid Turtle.

Concretely fixed:
- `"yes"^^xsd:boolean` no longer flips to `false`; `"TRUE"`, `"2"`, `""`, `"01"` also go verbose
- `"abc"`/`"1.5"`/`"0x10"^^xsd:integer` no longer emit bare invalid tokens
- `"1.2.3"`/`"."`/`"1e5"^^xsd:decimal` go verbose; valid-XSD-but-not-Turtle `"2."` normalizes to `2.0`
- `"NaN"`/`"INF"`/`"-INF"^^xsd:double` (valid xsd:double, no Turtle DOUBLE form) and `"abc"` no longer emit `NaN.0e0`-style garbage

Valid lexical forms serialize exactly as before. Tests: per-datatype native/verbose assertions plus serialize→parse round-trips asserting value+datatype preservation for every invalid form (30 new tests; unit suite 338 passing, full `npm test` + `npm run lint` clean).

jeswr (commented):
LGTM.

This should be flagged on the linkeddata/rdflib, solidOs and discourse channels before merging in case this affects consumers of the library in unexpected ways.

May need to reserve for an mver.

Issue #567, opened 2022-08-09
Can't process formula in subject because isSubject not compatible with N3


RubenVerborgh:
The document at https://drive.verborgh.org/tmp/2022/cha-58-direct.n3 causes problems when read/written in a Solid Pod via Mashlib, with the following error:

```
Outline.expand: Unable to fetch <https://drive.verborgh.org/tmp/2022/cha-58-direct.n3>: Failed to load  <https://drive.verborgh.org/tmp/2022/cha-58-direct.n3> Fetcher: Error trying to parse <https://drive.verborgh.org/tmp/2022/cha-58-direct.n3> as Notation3:
Error: Subject is not a subject type status: 200
```

The `Subject is not a subject` is thrown by `src/store.ts`, which uses an `isSubject` call that checks for `NamedNodeTermType`/`VariableTermType`/`BlankNodeTermType`.

So either rdflib needs to be adjusted to output RDF/JS-compatible Notation3 subjects (which is what N3.js does), or the `isSubject` check needs to be adjusted when working with Notation3.

bourgeoa:
> So either rdflib needs to be adjusted to output RDF/JS-compatible Notation3 subjects (which is what N3.js does), or the `isSubject` check needs to be adjusted when working with Notation3.

@timbl Can you help with this I don't understand what is missing in isSubject call

Issue #449, opened 2020-10-23
webpack 5 compatibility - reference to buffer


devkral:
Webpack  5 removed all nodejs specific polyfills (you have to include them manually). This causes rdflib.js to fail when not provided a polyfill explicitly (because of a reference to buffer)

It would be nice if this reference could be removed, so you can use rdflib.js without nodejs polyfills.

devkral:
stack trace (shortened):
 @ ./node_modules/readable-stream/lib/_stream_readable.js 72:17-58    <= (here are the problems)
 @ ./node_modules/readable-stream/readable-browser.js 1:10-63
 @ ./node_modules/n3/src/N3StreamWriter.js 2:0-44 6:44-53
 @ ./node_modules/n3/src/index.js 6:0-44 25:0-47:2
 @ ./node_modules/rdflib/esm/parse.js 4:0-42 56:25-35
 @ ./node_modules/rdflib/esm/index.js 15:0-28 61:0-63:72

As far I see it is a problem of a dependency. I will report a bugreport separately

srl295:
@devkral I also hit this. Also, `build:browser` does not pass `-t web`, which hits this err: https://github.com/webpack/webpack/issues/11660

To work around this, so far, _having no idea what I'm doing_ I did this:

- `npm i --save webpack@4`
- `npx webpack --progress -t web`

_Warning: YMMVAFAIKBBQ_

Vinnl:
@srl295 That downgrades you to Webpack 4, which works because it still ships the polyfills by itself. You can still add them manually in Webpack 5 by installing the required polyfills as instructed by the error message, so in the case of `buffer` that would be `npm install buffer`, in the case of `stream` IIRC it's `npm install readable-stream`. Webpack should then pick that up automatically.

Unfortunately the dependency comes from the package n3, which AFAIK has no plans to move away from it :(

Windvis:
I ran into a similar issue due to the `@frogcat/ttl2jsonld` package. It [conditionally requires the `url` module](https://github.com/frogcat/ttl2jsonld/blob/686ae54dd13c5769750a0dd879c5551c8e1ca7ad/ttl2jsonld.js#L4617) but that isn't polyfilled anymore in Webpack 5 so it fails the build.

<details>
<summary>Error message</summary>

```
WARNING in configuration
The 'mode' option has not been set, webpack will fallback to 'production' for this value.
Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
[...]

Issue #275, opened 2018-11-11
Return line number and character number in file in parser SyntaxError objects


timbl:
This would be handy for things like the source pane to set the cursor

PR #831, opened 2026-07-04
feat: replace the hand-rolled N3 parser with N3.js across the Turtle/N3 family

### LINKED ISSUE #110: Mime-Type for N-Triples

### LINKED ISSUE #205: Why "python Assertion failed" ?
see the following:
https://github.com/linkeddata/rdflib.js/blob/34a3742041d49e67d686351e865597e2fe908a1c/src/n3parser.js#L118

### LINKED ISSUE #214: multi-line literals in turtle only supported with double-quotes
Starting a multiline-literals with 3 single-quote will lead to 

    Bad syntax: objectList expected

Using 3 double quotes works as it should

### LINKED ISSUE #329: Failing RDF Turtle 1.1 conformance cases
I have set up [a repository](https://github.com/RubenVerborgh/rdflib-evaluation) to track rdflib.js conformance with regard to the specifications. We can use this in order to improve spec compliance.

Currently, [193 out of 298 tests are passing](https://travis-ci.org/RubenVerborgh/rdflib-evaluation/builds/566404908) for Turtle.

The 105 currently failing cases are:

- [ ] IRI_with_four_digit_numeric_escape (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#IRI_with_four_digit_numeric_escape)
- [ ] IRI_with_eight_digit_numeric_escape (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#IRI_with_eight_digit_numeric_escape)
- [ ] SPARQL_style_prefix (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#SPARQL_style_prefix)
- [ ] prefix_with_non_leading_extras (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#prefix_with_non_leading_extras)
- [ ] reserved_escaped_localName (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#reserved_escaped_localName)
- [ ] percent_escaped_localName (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#percent_escaped_localName)
- [ ] localname_with_COLON (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#localname_with_COLON)
- [ ] localName_with_non_leading_extras (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#localName_with_non_leading_extras)
- [ ] SPARQL_style_base (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#SPARQL_style_base)
- [ ] labeled_blank_node_with_non_leading_extras (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#labeled_blank_node_with_non_leading_extras)
- [ ] LITERAL1 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#LITERAL1)
- [ ] LITERAL1_ascii_boundaries (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#LITERAL1_ascii_boundaries)
- [ ] LITERAL1_with_UTF8_boundaries (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#LITERAL1_with_UTF8_boundaries)
- [ ] LITERAL1_all_controls (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#LITERAL1_all_controls)
- [ ] LITERAL1_all_punctuation (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#LITERAL1_all_punctuation)
- [ ] LITERAL_LONG1 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#LITERAL_LONG1)
- [ ] LITERAL_LONG1_ascii_boundaries (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#LITERAL_LONG1_ascii_boundaries)
- [ ] LITERAL_LONG1_with_UTF8_boundaries (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#LITERAL_LONG1_with_UTF8_boundaries)
- [ ] LITERAL_LONG1_with_1_squote (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#LITERAL_LONG1_with_1_squote)
- [ ] LITERAL_LONG1_with_2_squotes (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#LITERAL_LONG1_with_2_squotes)
- [ ] literal_with_CHARACTER_TABULATION (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#literal_with_CHARACTER_TABULATION)
- [ ] literal_with_BACKSPACE (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#literal_with_BACKSPACE)
- [ ] literal_with_LINE_FEED (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#literal_with_LINE_FEED)
- [ ] literal_with_CARRIAGE_RETURN (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#literal_with_CARRIAGE_RETURN)
- [ ] literal_with_FORM_FEED (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#literal_with_FORM_FEED)
- [ ] bareword_decimal (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#bareword_decimal)
- [ ] bareword_double (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#bareword_double)
- [ ] double_lower_case_e (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#double_lower_case_e)
- [ ] positive_numeric (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#positive_numeric)
- [ ] numeric_with_leading_0 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#numeric_with_leading_0)
- [ ] literal_true (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#literal_true)
- [ ] literal_false (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#literal_false)
- [ ] repeated_semis_at_end (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#repeated_semis_at_end)
- [ ] repeated_semis_not_at_end (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#repeated_semis_not_at_end)
- [ ] number_sign_following_localName (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#number_sign_following_localName)
- [ ] number_sign_following_PNAME_NS (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#number_sign_following_PNAME_NS)
- [ ] turtle-syntax-bad-num-05 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-LITERAL2_with_langtag_and_datatype)
- [ ] langtagged_LONG_with_subtag (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#langtagged_LONG_with_subtag)
- [ ] turtle-syntax-base-02 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-base-02)
- [ ] turtle-syntax-base-04 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-base-04)
- [ ] turtle-syntax-prefix-02 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-prefix-02)
- [ ] turtle-syntax-prefix-03 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-prefix-03)
- [ ] turtle-syntax-prefix-04 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-prefix-04)
- [ ] turtle-syntax-prefix-06 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-prefix-06)
- [ ] turtle-syntax-prefix-09 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-prefix-09)
- [ ] turtle-syntax-string-04 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-string-04)
- [ ] turtle-syntax-string-05 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-string-05)
- [ ] turtle-syntax-string-06 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-string-06)
- [ ] turtle-syntax-string-09 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-string-09)
- [ ] turtle-syntax-string-11 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-string-11)
- [ ] turtle-syntax-pname-esc-01 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-pname-esc-01)
- [ ] turtle-syntax-pname-esc-02 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-pname-esc-02)
- [ ] turtle-syntax-pname-esc-03 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-pname-esc-03)
- [ ] turtle-syntax-number-05 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-number-05)
- [ ] turtle-syntax-number-11 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-number-11)
- [ ] turtle-syntax-struct-04 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-struct-04)
- [ ] turtle-syntax-struct-05 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-struct-05)
- [ ] turtle-syntax-bad-uri-01 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-uri-01)
- [ ] turtle-syntax-bad-uri-02 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-uri-02)
- [ ] turtle-syntax-bad-uri-03 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-uri-03)
- [ ] turtle-syntax-bad-uri-04 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-uri-04)
- [ ] turtle-syntax-bad-uri-05 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-uri-05)
- [ ] turtle-syntax-bad-prefix-01 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-prefix-01)
- [ ] turtle-syntax-bad-prefix-02 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-prefix-02)
- [ ] turtle-syntax-bad-struct-02 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-struct-02)
- [ ] turtle-syntax-bad-struct-04 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-struct-04)
- [ ] turtle-syntax-bad-struct-05 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-struct-05)
- [ ] turtle-syntax-bad-struct-06 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-struct-06)
- [ ] turtle-syntax-bad-struct-07 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-struct-07)
- [ ] turtle-syntax-bad-kw-04 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-kw-04)
- [ ] turtle-syntax-bad-kw-05 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-kw-05)
- [ ] turtle-syntax-bad-n3-extras-02 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-n3-extras-02)
- [ ] turtle-syntax-bad-n3-extras-03 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-n3-extras-03)
- [ ] turtle-syntax-bad-n3-extras-04 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-n3-extras-04)
- [ ] turtle-syntax-bad-n3-extras-05 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-n3-extras-05)
- [ ] turtle-syntax-bad-n3-extras-06 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-n3-extras-06)
- [ ] turtle-syntax-bad-n3-extras-07 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-n3-extras-07)
- [ ] turtle-syntax-bad-n3-extras-08 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-n3-extras-08)
- [ ] turtle-syntax-bad-n3-extras-09 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-n3-extras-09)
- [ ] turtle-syntax-bad-n3-extras-10 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-n3-extras-10)
- [ ] turtle-syntax-bad-n3-extras-13 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-n3-extras-13)
- [ ] turtle-syntax-bad-struct-14 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-struct-14)
- [ ] turtle-syntax-bad-struct-15 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-struct-15)
- [ ] turtle-syntax-bad-struct-16 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-struct-16)
- [ ] turtle-syntax-bad-struct-17 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-struct-17)
- [ ] turtle-syntax-bad-lang-01 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-lang-01)
- [ ] turtle-subm-11 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-subm-11)
- [ ] turtle-subm-16 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-subm-16)
- [ ] turtle-subm-17 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-subm-17)
- [ ] turtle-subm-19 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-subm-19)
- [ ] turtle-subm-20 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-subm-20)
- [ ] turtle-subm-22 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-subm-22)
- [ ] turtle-eval-bad-01 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-eval-bad-01)
- [ ] turtle-eval-bad-02 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-eval-bad-02)
- [ ] turtle-eval-bad-03 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-eval-bad-03)
- [ ] turtle-eval-bad-04 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-eval-bad-04)
- [ ] turtle-syntax-bad-ln-dash-start (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-bad-ln-dash-start)
- [ ] turtle-syntax-blank-label (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-blank-label)
- [ ] turtle-syntax-ln-colons (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-ln-colons)
- [ ] turtle-syntax-ln-dots (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-ln-dots)
- [ ] turtle-syntax-ns-dots (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#turtle-syntax-ns-dots)
- [ ] IRI-resolution-01 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#IRI-resolution-01)
- [ ] IRI-resolution-02 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#IRI-resolution-02)
- [ ] IRI-resolution-07 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#IRI-resolution-07)
- [ ] IRI-resolution-08 (http://w3c.github.io/rdf-tests/turtle/manifest.ttl#IRI-resolution-08)

### LINKED ISSUE #352: Error when parsing valid shacl/ttl file
Recently a specific file on my Solid Pod began throwing errors. The file in question is [here](https://jmartin.inrupt.net/public/shapes/book-shacl.ttl), but others were able to replicate the problem with the same file contents as evidenced [here](https://megoth.solid.community/public/testdir/book-shacl.ttl).

The error is strange because the source view is fine, and I can edit and save and it validates as valid ttl. But the data view is throwing errors. In the console I see: `setCredentials: https://jmartin.inrupt.net/public/shapes/book-shacl.ttl starts as: undefined` and a larger error is displayed on the screen:

```
Error: Unhandled node type: undefined
  undefined
    Error: Unhandled node type: undefined  undefined
        at m (https://megoth.solid.community/mashlib.min.js:27:330581)  undefined
        at m (https://megoth.solid.community/mashlib.min.js:27:330386)  undefined
        at h (https://megoth.solid.community/mashlib.min.js:27:329136)  undefined
        at m (https://megoth.solid.community/mashlib.min.js:27:330021)  undefined
        at h (https://megoth.solid.community/mashlib.min.js:27:329136)  undefined
        at Object.statementsAsTables (https://megoth.solid.community/mashlib.min.js:27:331015)  undefined
        at https://megoth.solid.community/mashlib.min.js:27:332063  undefined
        at Object.render (https://megoth.solid.community/mashlib.min.js:27:332090)  undefined
        at _.propertyTable (https://megoth.solid.community/mashlib.min.js:21:35517)  undefined
        at u (https://megoth.solid.community/mashlib.min.js:21:42368)
```

### LINKED ISSUE #405: Merging graphs with blank nodes

### LINKED ISSUE #494: Does not properly process 8-bit unicode escape sequences on string literals
This is an extension of this issue on N3.js which does process 8-bit unicode escape sequences properly: https://github.com/rdfjs/N3.js/issues/248. There is also a test suite to test this: https://github.com/w3c/rdf-tests/pull/65.

## Problem 1: Strange backslashes when writing to a Pod. 

Let's say I want to PATCH the following triple to my POD `<https://pod.example/chat1.ttl#message1> <http://rdfs.org/sioc/ns#content> "Here's my emoji: 😊"`.

Parsing this using n3 turns this into `<https://pod.example/chat1.ttl#message1> <http://rdfs.org/sioc/ns#content> "Here's my emoji: \\U0001f60a"`. Or at least that's what you would want to write to a Pod.

Running the following code
```typescript
function patchToPod(uri: string, dataset: DatasetCore) {
  const writer = new Writer({ format: "N-Triples" });
  for (const quad of dataset) {
    writer.addQuad(quad);
  }
  writer.end(async (error, parsedString: string) => {
    // Parsed String is `<https://pod.example/chat1.ttl#message1> <http://rdfs.org/sioc/ns#content> "Here's my emoji: \U0001f60a".`
    fetch(uri, {
      method: "PATCH",
      body: `INSERT { ${parsedString} }`,
      headers: { 'content-type': 'application/sparql-update' }
    })
  });
}
```

causes the following to be written to the Pod on NSS:
```
@prefix : <#>.
@prefix ch: <https://pod.example/chat1.ttl#>.
@prefix n: <http://rdfs.org/sioc/ns#>.

ch:message1 n:content "Here's my emoji: \uf60a".
```

Notice that instead of `\U0001F60A` it's `\uf60a`.  I'm not 100% sure why this is because I'm not super versed in escape codes, but @RubenVerborgh believes this is a problem with rdflib.js on NSS.

Adding an additional slash to the `parsed` string in the code above fixes it:

```typescript
fetch(uri, {
  method: "PATCH",
  body: `INSERT { ${parsedString.replace(`\\U`, `\\\\U`)} }`,
  headers: { 'content-type': 'application/sparql-update' }
})
```

However, @RubenVerborgh confirmed that adding these backslashes shouldn't be required.

## Problem 2: Solid Clients don't understand 8-bit unicode escape sequences

When a unicode escape sequence is correctly formatted in the Pod, clients using rdflib.js are unable to understand it. For example, here's what a message in the SolidOS chat looks like with unicode escape:

```
:fd0071f5-01f9-416d-aa01-24eb6666d618
    n:content "Emoji3 \\U0001f60a";
```
Causes:
image

```
:fd0071f5-01f9-416d-aa01-24eb6666d618
    n:content "Emoji3 \U0001f60a";
```
Causes:
image

But,
```
:fd0071f5-01f9-416d-aa01-24eb6666d618
    n:content "Emoji3 \ud83d\ude0a";
```
Causes:
image

So, even if I were to build functionality into my app to parse unicode, it wouldn't be interoperable with the apps that do not.

### LINKED ISSUE #626: Support `PREFIX` directives in turtle parsers.
Currently turtle parsers seems doesn't take `PREFIX` case-sensitive sparql-like directives, which are supported by turtle syntax from 2014. It would be highly valuble for their support to handle turtle from wild.

### LINKED ISSUE #647: Parsing error on prefix:N.N syntax
I noticed this turtle/N3 syntax fails to parse

`@prefix myprefix: <http://www.example.com/ontologies/test/> .
<http://example.com/data/> myprefix:version myprefix:1.18 .`

with error:  "Bad syntax: EOF found in middle of path syntax"
this syntax is used in many open source / public domain ontologies for describing ontology / definition versions

if the value of the statement is fully qualified 
`<http://example.com/data/> myprefix:version <http://www.example.com/ontologies/test/1.18> .`
it parses ok, it is the prefix / qname style syntax that throws the parsing error

### LINKED ISSUE #651: Use SPARQL Query prefix declaration instead of
It seems that `sparqlUpdateParser` in [`src/patch-parser.js`](
https://github.com/linkeddata/rdflib.js/blob/cfc0f8c437d7872d76dc384d686c28d5d6100e35/src/patch-parser.js) is partly relying on the N3 Parser instead of conforming to SPARQL Query, in particular the prefix declaration. `@prefix` line ending with a dot (`.`) is not valid syntax in SPARQL Query AFAICT:

The syntax of the prefix declaration in SPARQL Query is as follows ( https://www.w3.org/TR/sparql11-query/#rPrefixDecl ):

>[6]  	PrefixDecl	  ::=  	'PREFIX' [PNAME_NS](https://www.w3.org/TR/sparql11-query/#rPNAME_NS) [IRIREF](https://www.w3.org/TR/sparql11-query/#rIRIREF)

And SPARQL Update states ( https://www.w3.org/TR/sparql11-update/#languageForm ):

>PREFIX definitions and the [syntax for IRIs](https://www.w3.org/TR/sparql11-query/#QSynIRI) in update requests in general follow the same conventions as in the [SPARQL1.1 Query Language](https://www.w3.org/TR/sparql11-query/).

---

https://github.com/linkeddata/rdflib.js/blob/cfc0f8c437d7872d76dc384d686c28d5d6100e35/src/patch-parser.js#L76-L91

I think instead of `@prefix`, it should be `PREFIX` (and adjust the length):

`if (!found && str.slice(j, j + 6) === 'PREFIX') {`

and the prefix declaration line does not end with a dot (`.`) so perhaps the following is unnecessary:

`i = p.checkDot(str, i);`

That's close to getting it working but the following line:

`i = p.directive(str, j);`

depends on n3parser. See:

```
var _n3parser = _interopRequireDefault(require("./n3parser"));
var p = (0, _n3parser.default)(kb, kb, base, base, null, null, '', null);
```

And going further down the rabbit-hole we have this in token checker in n3parser:

https://github.com/linkeddata/rdflib.js/blob/cfc0f8c437d7872d76dc384d686c28d5d6100e35/src/n3parser.js#L450

which essentially checks to see if there is "@" before prefix. Since the valid syntax "PREFIX" does not, it all falls apart.

---

So I don't know if there is a quick fix here but as far as conformance goes `PrefixDecl` is defined in SPARQL Query. I do not know if rdflib.js has a SPARQL Query parser somewhere or can borrow from somewhere.

Perhaps [`src/sparql-to-query`](https://github.com/linkeddata/rdflib.js/blob/cfc0f8c437d7872d76dc384d686c28d5d6100e35/src/sparql-to-query.js) is of help? See https://github.com/linkeddata/rdflib.js/blob/main/src/sparql-to-query.js#L267-L285

I was going to PR something but think that what I wrote here should be reviewed. I don't have the bandwidth to dig deeper or find a clever solution because there is some legacy code and it can use more eyes.

---

This whole thing causes an issue in code depending on it like, e.g., NSS. So, AFAICT, NSS does not conform in that it rejects requests with valid payload containing "PREFIX", and also accepts invalid syntax "@prefix".

jeswr:
> **Agent-generated draft PR** for @jeswr to review — nothing merges without his sign-off. Part of the triaged-backlog programme (#823 / #824). The mandate, per @jeswr's steer: *"rip out the current n3parser code entirely … where possible remove as much custom code from RDFlib as you can and make use of the N3.js API instead."* The behavioural calls below need explicit sign-off; nothing here is final.

Replaces the hand-rolled 2005-era N3 parser with N3.js across the whole Turtle/N3 family, so `parse()`, `Fetcher.load()` and `sparqlUpdateParser` agree with each other for the first time — one parser, net **−1,022 lines** under `src/`.

## What is deleted / added

| File | Lines | Fate |
|---|---|---|
| `src/n3parser.js` | **−1,610** | deleted — no fallback kept |
| `src/lists.ts` (`convertFirstRestNil`, `substituteInDoc`, `substituteNillsInDoc`) | **−121** | deleted — replaced by N3.js `Store#extractLists` (it was not exported from the package index) |
| `src/patch-parser.js` | 94 → 192 | rewritten: no longer drives legacy-parser internals (`p.skipSpace`/`p.node`/`p.directive`); it rewrites the patch to N3 and reuses the same N3.js pipeline |
| `fetcher.ts` `N3Handler` inline parser | — | now delegates to `parse()` |
| `src/n3-adapter.ts` | **+398** (new) | the single N3.js↔rdflib adapter everything routes through |
| `src/utils/literalValue.ts` | **+156** (new) | value-space literal helpers + the opt-in `canonicalize` flag (the migration path — see mitigation M3 below) |

## How it uses the N3.js API (not custom code)

- **Parsing**: one `N3.Parser` call per document, with the format string enforcing the right grammar per content type — `text/turtle` (+`application/x-turtle`), `text/n3` (+`application/n3`), `application/n-triples`, `application/n-quads` (+`nquads`), `application/trig`. N-Triples/TriG support is new; the ad-hoc n-quads callback path is gone. Synchronous, using N3.js's array-returning mode with `{ onPrefix }` for prefix capture.
- **Lists**: `N3.Store#extractLists({ remove: true })` finds, validates and removes `rdf:first/rest/nil` chains; the adapter resolves the returned head→items map (recursively, for nested lists) into rdflib `Collection` terms, and turns lone `rdf:nil` into an empty `Collection` — matching the legacy parser's output, including for *explicitly reified* chains and `rdf:nil`-as-element (the existing `lists-test.js` expectations all pass unchanged). Folding is gated on the factory's `COLLECTIONS` support exactly as before, and skipped entirely when a document never mentions `rdf:nil` (no intermediate store on the hot path). Line-dump formats (N-Triples/N-Quads/TriG) are never folded — quad dumps round-trip verbatim.
- **Formulae (`text/n3`)**: N3.js emits each `{ … }` as quads under a fresh blank-node graph label; the adapter rebuilds them into rdflib `Formula` sub-stores (recursively for nesting, with the empty-formula `{}` marker special-cased). `=` → `owl:sameAs`, `=>` → `log:implies`, `<=` → reversed `log:implies`, `!`/`^` paths, and `?x` → `Variable` are all N3.js-native.
- **Quantifiers**: parsed with `explicitQuantifiers: true`; the reified declarations in `urn:n3:quantifiers` are read back into `newUniversal()` / `declareExistential()` on the right scope (store or formula), keeping the legacy representation (terms stay `NamedNode`s).
- **SPARQL-Update patches**: `sparqlUpdateParser` scans only the top-level keywords (`INSERT`/`DELETE`/`WHERE`, optional `DATA`, `@prefix`, and now SPARQL-style `PREFIX`), rewrites the document to `<#query> patch:where { … } .` N3, and hands it to the same adapter — clause `{ … }` bodies are parsed by N3.js, not by hand.

## Issues closed

- Closes #329 — Turtle 1.1 conformance, now on **every** entry point (`parse()`, `Fetcher.load()`, patches): SPARQL `PREFIX`/`BASE`, single-quoted long literals, numeric/`\U` IRI escapes, bareword numbers/booleans, dotted local names.
- Closes #214 — `'''…'''` long literals (old parser: `Bad syntax`).
- Closes #494 — astral `\U0001F60A` escapes decode to the correct code point.
- Closes #626 — SPARQL-style `PREFIX` directives.
- Closes #110, Closes #220 — `application/n-triples` parsing support (previously threw "Don't know how to parse").
- Closes #352 — booleans in collections are real `Literal`s with `termType` (v2.4.0 pushed raw JS `true`/`false` into `Collection.elements`, crashing consumers; regression test added).
- Closes #651 — patches accept SPARQL-style `PREFIX` without a trailing dot (test added). `@prefix` is still accepted for backwards compatibility — say the word if you'd rather hard-drop it.
- Closes #647 — dotted local names (`myprefix:1.18`) parse to the correct IRI (valid Turtle 1.1 `PN_LOCAL`; probe: 1 statement, object `.../test/1.18`).
- Closes #405 — blank nodes are fresh per parse: the same document parsed under two base URIs yields distinct bnodes instead of conflated ones (probe: 2 distinct labels).
- Closes #205 — the question is about a comment in `src/n3parser.js:118`; that file is deleted here.
- Addresses #567 (soft claim — the failure is not reproducible as a v2.4.0 regression, so it may have already worked; kept as coverage): N3 documents with formula subjects (the cha-58-direct.n3 repro shape) now load through the Fetcher into the store; regression test added at the fetcher level.
- **Not** claiming #275 (line/character positions on parse errors): N3.js error *messages* now always carry the line number (e.g. `Unexpected "???" on line 1`), but there are no structured `line`/`column` fields on the thrown error and no character offset — left open as a small follow-up on the adapter.

## Behavioural changes (need your sign-off)

1. **`text/turtle` is now strict Turtle.** The legacy code parsed full N3 under a turtle content type (same parser for both), and the Fetcher routed *all* fetched turtle through it. Mislabelled N3-in-turtle documents now fail with N3.js's descriptive errors. *Escape hatch if you want wire-compat instead: route turtle content types to `format: 'text/n3'` in the adapter (n3 mode is a turtle superset) — a one-line change; my recommendation is strict.*
2. **Literal lexical forms are preserved** (`12.0`, `3.141e0`, `true` stay as written; the legacy parser normalised to `12`, `3.141`, `1`). This is what Turtle 1.1 requires, but it changes term identity (`"12"^^xsd:decimal` ≠ `"12.0"^^xsd:decimal`) and serializer output. The Turtle serializer had a latent bug here — it emitted `false` for `"true"^^xsd:boolean` because it only knew the legacy `'1'` form — fixed in `serializer.js`. Mitigation M3 (below) is the migration path for consumers that match on the normalised forms; the wider read-lenient/write-strict question is under discussion in this thread.
3. **N-Quads default-graph statements** are attributed to the document graph `kb.sym(base)` (rdflib's provenance convention) instead of the old callback path's `DefaultGraph` term. Consistent with every other format; flagged as breaking for anyone matching on the default graph.
4. **`N3Parser` is no longer exported** (breaking API; a throwing stub with a descriptive pointer remains — M5 below). `sparqlUpdateParser` keeps its exact signature/return shape (`{query, insert?, delete?, where?}` with `Formula` values, plus the `<#query> patch:*` statements on the store).
5. **Empty-prefix leniency kept, reimplemented safely**: `:name` without `@prefix :` still resolves against `<base#>`. Instead of a document-wide regex + unconditional input mutation, the adapter parses strictly first and retries **once** only on N3.js's specific `Undefined prefix ":"` error — zero cost and zero false positives on conforming documents, and mid-document `@prefix :` declarations override exactly like the legacy sequential semantics. (TimBL's own `structures.n3` fixture depends on this leniency.)
6. **Dropped cwm-era leniencies — now loud, descriptive parse errors** (each with a test asserting the error): `@keywords`, bareword date/dateTime literals (`2026-07-01` → write `"2026-07-01"^^xsd:date`), `has …` / `is … of` (write the inverse triple; the modern N3 CG spec dropped them), string-literal subjects, undeclared non-empty prefixes (note: these already errored in v2.4.0 — only the error message/type changes).
7. **Upstream N3.js bug found & worked around** (a candidate for a fix in N3.js itself): in n3 mode, a document-labelled bnode `_:c` mentioned inside a `[ … ]` property list gets the *graph* label as its scope prefix (top level: label `.c`), splitting it from its other mentions. Repro: `new N3.Parser({format:'text/n3'}).parse('_:c <urn:p> 1. [ <urn:q> _:c ] <urn:r> 2.')`. The adapter passes an explicit `blankNodePrefix` and repairs `.`-prefixed labels; formula-scoped labels (correct per N3 semantics) are untouched.
8. **RDF-star**: N3.js parses quoted triples but rdflib's store has no term type for them — the adapter throws a descriptive error rather than crashing mid-store as `main` does today. (Real support would be a store-model project.)
9. **Lists inside N3 formulae are now folded into `Collection` terms** (surfaced by the independent review pass; v2.4.0 left raw `rdf:first`/`rdf:rest` triples inside formula sub-stores — verified by repro against the published 2.4.0). Consistent with the migration and arguably a fix, but it changes formula contents for consumers that match on `first`/`rest` inside clauses (e.g. SPARQL-update WHERE bodies).

## Breaking changes, semver and mitigations

A full breaking-change assessment ran against this branch (probe-by-probe vs built v2.4.0, plus consumer hunts across the SolidOS stack, @ldflex/rdflib, NSS, and npm/GitHub reverse-deps): see [the assessment comment](https://github.com/linkeddata/rdflib.js/pull/831#issuecomment-4883145352). Verdicts: **4 BREAKING** (strict `text/turtle` vs legacy pod files; literal lexical preservation flipping solid-ui boolean reads; dropped cwm leniencies vs live `text/n3` documents; re-parse bnode accumulation at force-reload sites), 10 potentially-breaking with no live consumer hit found, the rest safe or strict fixes.

**This must ship as v3.0.0** — solid-ui/mashlib/solid-logic/solid-panes (`^2.3.x`) and @ldflex/rdflib (`^2.2.0`) all auto-adopt any 2.x release.

The full mitigation set recommended by the assessment is on the branch:

| Mitigation | Commit |
|---|---|
| **M1** — `force: true` now implies `clearPreviousData: true` in `Fetcher.load` — de-breaks the 81 observed force-reload sites (bnode accumulation) | 0ccb5d7 |
| **M2** — fix for a regression this branch had introduced: `sparqlUpdateParser` under a fragment-bearing base silently dropped its insert/delete clauses; regression test added | 0ccb5d7 |
| **M3** (per @jeswr: both helper + opt-in flag) — value-space helpers `isTrue(term)`, `literalToBoolean(term)`, `literalToNumber(term)` (also exposed as `Literal.toBoolean` / `Literal.toNumber`), plus an opt-in `parse(str, kb, base, contentType, { canonicalize: true })` flag that restores the ≤2.x parse-time normalisation of boolean/numeric lexical forms (`true` → `"1"`, `12.0` → `"12"`, `3.141e0` → `"3.141"`) across the Turtle family — off by default, so lexical preservation stays the shipped behaviour. `isTrue(kb.any(s, p))` is the one-line fix for `^2.x` patterns like `term.value === '1'` | c9ec6a6 |
| **M5** — deprecated `N3Parser` export stub that throws a descriptive pointer to `parse()` | 0ccb5d7 |
| **M8** — docs: `Documentation/turtle-intro.html` updated for the two removed extensions (0ccb5d7); `reference/README.md` marks `dumpParser.js` / `ldpatchParser.js` / `fetcher-classes.js` as archived relics that drive `N3Parser` internals which no longer exist (2b10b5b) | 0ccb5d7 / 2b10b5b |
| **Bundle hygiene** — the adapter deep-imports `n3/lib/{N3Parser,N3Store,N3DataFactory}.js`, preserving #836's #449 deep-import discipline (no `N3StreamWriter`/`readable-stream` in browser bundles); a source-level hygiene test now fails any future root `'n3'` import under `src/`; verified zero stream/writer/`readable-stream` occurrences in the built `dist/rdflib.min.js` | 5a6bce9 |

## Test changes — all justified, none loosened

- `parse-turtle-family-test.js` (new): 30+ cases — the conformance suite plus formulae (`=>`/`<=`/`=`, nesting, empty `{}`), variables, quantifier scoping, nested/reified/empty collections, #352, the label-repair regression, and explicit "documented gap" error assertions for `is…of` / bareword dates / `@keywords`.
- `patch-parser-test.js`: +6 cases (variables, `@prefix`/`PREFIX`, `INSERT DATA`, `;` separators + braces-inside-strings, store side-effects, error reporting).
- `literal-value-test.js`, `parse-canonicalize-test.js`, `tests/types/value-space.ts` (new): the M3 helpers and `canonicalize` flag, value- and type-level.
- `n3parser-stub-test.js`, `n3-import-hygiene-test.js` (new): the M5 stub's error pointer; the no-root-`'n3'`-import source guard.
- `fetcher-test.js`: the N3 fixture used a **bareword date** — modernised to `"2012-03-12"^^xsd:date` (same graph; the old syntax is a documented gap, behavioural change 6). New #567 formula-subject load test.
- `lists-test.js`: expectations unchanged; only the redundant post-parse `convertFirstRestNil()` calls were removed (the function is deleted; `parse()` itself folds).
- `serialize-test.js`: the RDF/XML expectation now uses the preserved lexical forms (`12.0`/`3.141e0`) — the old expectation encoded the legacy normalisation; the JSON-LD expectation in the same file already used the preserved forms.
- `indexed-formula-test.js`: the store under test is now `DataFactory.graph()` (the previous `new IndexedFormula()` version passed vacuously because the legacy parser misfiled the status list into the default graph), **plus** a new companion test pinning the non-collection-store outcome explicitly (raw `first/rest` triples survive `removeMetadata` — the pre-existing #631 gap, now visible instead of hidden).
- **`tests/serialize` golden fixtures changed** — flagging every ref-file diff explicitly. Sources `t1.ttl`, `t2.ttl`, `t7.n3`, `structures.n3` used bareword dates / `is…of` and are rewritten to express the **same graphs** in standard syntax; the regenerated references are `t1-ref.xml`, `t2-ref.xml`, `t7-ref.nt`, `t11-ref.xml`, `t12-ref.ttl`, `t13-ref.ttl`, `t17-ref.xml`. Every delta was hand-checked to be one of exactly two classes: (a) preserved lexical forms (`10.0`, `1.45e5`, `true`), (b) parser-generated bnode label names (`_:dl1_coretta` vs `_:_g_L4C88`). The `t11` structures ref also exposed the serializer boolean bug fixed above (behavioural change 2).

## Gates

CI is green on the branch head (Node 22.x + 24.x). Full local runs, all exit 0: at the migration commit, `npm run build` · `test:unit` (**349 passing, 0 failing**) · `test:serialize` (all 17) · `test:types` · `build:types` · `build:esm` · `lint` (0 errors; 5 pre-existing warnings untouched); re-verified with the M1/M2/M5 mitigations at 0ccb5d7 (**355 unit passing**, serialize/types/lint clean). The bundle-hygiene commit (5a6bce9) additionally verified the built `dist/rdflib.min.js` contains no stream/writer code.


jeswr:
This tripped into using opus halfway through so I am asking fable to rewrite it.

bourgeoa:
If I understand well this is a major change that breaks parsing of existing turtle files with 0/1 as Boolean values in rdflib context.

jeswr:
> If I understand well this is a major change that breaks parsing of existing turtle files with 0/1 as Boolean values in rdflib context.

N3.js is compliant to the current RDF 1.1/1.2 test suites for turtle and N3 CG test suites for Notation3.

Looking at the changes to the test files it seems like there might indeed be some deviations in this parser from those suites. I'll take your guidance on whether that is considered a patch or breaking change - I see the merit in holding this PR for a next mver.

jeswr:
## Breaking-change assessment (v2.4.0 → this branch)

We diffed runtime behavior probe-by-probe against built v2.4.0 and this branch, then hunted consumers (SolidOS stack, @ldflex/rdflib, node-solid-server, npm/GitHub reverse-deps). Full report with repros, per-delta migration snippets, and mitigation costings is in the linked document. Verdicts:

| Delta | Verdict | Evidence |
|---|---|---|
| Strict `text/turtle` (N3 syntax rejected) | **BREAKING** | mashlib issue-tracker fixtures (TBL-era pod copies with `@keywords`, `is…of`) now `FetchError`; NSS 500s on PATCH of stored N3-flavoured `.ttl`; 221 `log:implies` `.ttl` files on GitHub |
| Literal lexical preservation | **BREAKING** | 4 solid-ui sites silently read stored `true` as false (`preferences.js:211`, `forms.js:1892`/`:192`, booleanField `holds()`); rdflib's own serializer emits `true`, so pods hold the breaking form |
| cwm leniencies dropped (`is…of`, barewords, `@keywords`) | **BREAKING** | @ldflex/rdflib recorded tests replay TimBL's card as text/n3 with `is doap:developer of` (line 235) → parse throws; the live w3.org card's N3 variant still uses it today |
| Re-parse duplicates bnode subgraphs | **BREAKING** | `fetcher.load(doc, {force:true})` never clears → stores grow every refresh; 81 call sites incl. solid-ui chat, pod-chat-client (per ACL check), gitter-solid daemons. Repro: 3→3 stmts old, 3→5 new |
| nquads graph attribution, `N3Parser` removal, lists-in-formulae→Collection, sync nquads callback, bnode label instability/merging, langtag lowercasing, patch `why`, fragment-base patch regression | potentially breaking | plausible patterns; adversarial search found no live consumer hit (details in report) |
| Everything else (empty-prefix leniency, RDF-star rejection, new content types, Turtle 1.1 fixes, #352 boolean collections, SPARQL PREFIX patches, perf) | safe | parity, additive, or strict fixes — #352 fix resolves the known mashlib data-pane crash, and the old `\U` escape silently produced empty-string literals, so several changes are corruption fixes |

**Semver: this needs to be v3.0.0.** solid-ui/mashlib/solid-logic/solid-panes (`^2.3.x`) and @ldflex/rdflib (`^2.2.0`) all auto-adopt a 2.x release on their next fresh install.

Cheap mitigations worth adopting in this PR:
1. Make `force: true` imply `clearPreviousData: true` in `Fetcher.load` — one line, de-breaks all 81 observed force-reload sites (the accumulation delta).
2. Fix a regression this branch introduces: `sparqlUpdateParser` with a fragment-bearing base URI silently returns a patch without its `insert`/`delete` clauses (`<#query>` resolves inconsistently) — one-line fix; NSS/solid-rest only survive because they strip fragments first.
3. Ship a boolean value-space helper (and/or an opt-in `canonicalize` parse flag) so `^2.x` consumers have a one-line fix for the literal-identity change.
4. Keep a deprecated `N3Parser` export that throws a pointer to `parse()`.
5. Docs/body: `Documentation/turtle-intro.html` still documents two now-throwing extensions (naked dates, `is…of`); body nits — undeclared prefixes already errored in 2.4.0 (message-only change), n-quads RDF-star previously corrupted the store silently rather than crashing, "Closes #567" isn't reproducible as a 2.4.0 regression, and the dropped-leniency list should add the bareword *datetime* form (which also worked under text/turtle).

cc @jeswr — this assessment was agent-assisted (Claude Code). Every verdict was produced by running identical probes against both built checkouts; consumer hits are cited to file:line or reproduced against the consumers' own fixtures.

bourgeoa:
> > If I understand well this is a major change that breaks parsing of existing turtle files with 0/1 as Boolean values in rdflib context.
> 
> N3.js is compliant to the current RDF 1.1/1.2 test suites for turtle and N3 CG test suites for Notation3.
> 
> Looking at the changes to the test files it seems like there might indeed be some deviations in this parser from those suites. I'll take your guidance on whether that is considered a patch or breaking change - I see the merit in holding this PR for a next mver.

@jeswr  I'm not sure we are on the same line.
My main issue is `user data related`.
The new N3.js parsing cannot break actual data parsing
This means that Boolean 0/1 should still be parseable
true/false should also be parsed correctly
And then may be serialise could always render true/false. With may be apps updates.

We cannot decide that years of data are not parseable rdflib turtle anymore.

Boolean is may be not the only issue. But this is an old one. And ran in a blocker to solve it

jeswr:
Ah I am with you now @bourgeoa - I can think of two things that we can do here:
 - 1. Target that the next non-breaking release of rdflib is "read lenient" and "write strict"; so that apps are "cleaning up" files to be as compliant to current RDF 1.1 / 1.2 / N3 CG standards as possible (noting that there may be some cases where this is not possible); whilst also being able to work with the existing data.
 - 2. Write a data migration app if we ever did introduce breaking functionality; so that app authors can link their users to the data migration app when the app encounters a file it can't parse. The migration app would make its way through files in the Pod and clean them up to use RDF 1.1 / 1.2 / N3 CG standards complaint syntax.

bourgeoa:
This is is a very long open issue. See https://github.com/nodeSolidServer/node-solid-server/issues/1468#issuecomment-1451552027.

It needs a more open discussion.

Not sure that serialization to false/true will not introduce more bugs.
Among all issues reported it is the only one that is really touchy.

jeswr:
> Not sure that serialization to false/true will not introduce more bugs.

Ah .... I see. I'll look at this more closely in a bit to see if there is a way of going about this that doesn't break anything or annoy anoyone.

Issue #458, opened 2020-11-21
Disable sameAs automatic node merging by default


timbl:
https://github.com/linkeddata/rdflib.js/blob/master/src/store.ts#L198

sets the default mode for a Store to implement 'smushing' of nodes together in the graph.  This is not something standard functionality, and may confuse future developers.

These features were used in the mashlib when it was used in the old says of 'mashups', to when combining data from many trusted open data sources, like the network of FOAF files.   In fact the FOAF ontology specifically identify people by having the same properties, such as same email address or hash or email address.

But nowadays in the solid world, the data is from different resources has very different trust and this technique is not so appropriate.  Anyway it is I think reasonable to expect devs in future who want to use the features to turn them on explicitly.

This is formally a breaking change to the API though I doubt many people even know about the features.  This will of improve RDF/JS compatibility.

timbl:
Fixed by https://github.com/linkeddata/rdflib.js/pull/492

PR #318, opened 2019-06-01
Typography corrections

codam:
Good job guys!

PR #832, opened 2026-07-05
chore: repo & docs hygiene quick-fix wave (QF-1/5)

### LINKED ISSUE #87: rdflib includes jsonparser.js but it doesn't appear to be used
In web.js, function parse(str, kb, base, contentType, callback) handles contentType application/ld+json by using JSON.parse to parse the JSON document, use jsonld.toRDF to convert the jsonDocument o application/nquads, then re-parses the nquads with the n3Parser - causing the text to be parsed twice in two different formats. 

File jsonparser.js defines the $rdf.jsonParser which would seem to parse the JSON or JSON-LD directly into the kb without going through the extra format conversion. However, I can't find anyplace this parser is used. What is this parser intended to be used for, and why isn't used in the parse function?

### LINKED ISSUE #88: Bad function in example/book.js
I'm trying to work from https://github.com/linkeddata/rdflib.js/blob/master/example/people/book.js

But there is a function
kb.load(uri) which doesn't work.

I get an error that kb.load isn't a function

Has this function been deprecated or renamed?

### LINKED ISSUE #221: Document when http://www.w3.org/2007/ont/link#uri	statements are added
It seems that when there is an `owl:sameAs` statements rdflib.js adds a statement with the predicate `http://www.w3.org/2007/ont/link#uri` pointing to a literal with the same value as the IRI to with the `owl:sameAs` statement points. It would be nice to have this behavior and its rationale documented.

### LINKED ISSUE #301: SPARQL Query Documentation
I can't find information on how to perform SPARQL-Queries on a triplestore using rdflib.js except in #160. Although I kind of understand it now, I would have been happier if I would have discovered it earlier along along with some explanation of what's actually happening.

Wouldn't it make sense to add this to the documentation in linkeddata/rdflib.js/Documentation/webapp-intro.html?

### LINKED ISSUE #319: Readme: Broken Link - RDFJS Representation Task Force spec
This link goes to a 404:
https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md

### LINKED ISSUE #328: Literal datatype definition unclear / undocumented
According to the [documentation](https://github.com/solid/solid-tutorial-rdflib.js#rdf-term-types) typed literals should be created with `literal(value,undefined, datatype)`. This does not work in `rdflib 0.20.3`. You need to use `new Literal(value, undefined, dataype)` or explicitly set the datatype:

```JS

const testtype = new $rdf.NamedNode('http://www.w3.org/2001/XMLSchema#gYear')
let test = new $rdf.Literal('2019', undefined, testtype)

console.log();

console.log(`Datatype:\n${JSON.stringify(testtype,null,4)}\n\n`);
console.log(`Literal created with class call:\n${JSON.stringify(test,null,4)}\n\n`);

test = $rdf.literal('2019', undefined, testtype)

console.log(`literal created with function call:\n${JSON.stringify(test,null,4)}`)

test.datatype=testtype

console.log(`Literal after explicitly setting datatype:\n${JSON.stringify(test,null,4)}`)



const FOAF = $rdf.Namespace("http://xmlns.com/foaf/0.1/")
const iri = new URL('http://example.org/name/name')

let sts = []
let graph = new $rdf.graph()

sts.push(new $rdf.Statement($rdf.Namespace(iri.href)('#notme'),FOAF('name'), 'Meier'))
sts.push(new $rdf.Statement($rdf.Namespace(iri.href)('#notme'),FOAF('birthday'), test))


graph.addAll(sts)

console.log($rdf.serialize(undefined, graph, 'http://example.org/name/', 'text/turtle'))
```

### Output:
```
Datatype:
{
    "termType": "NamedNode",
    "value": "http://www.w3.org/2001/XMLSchema#gYear"
}


Literal created with class call:
{
    "termType": "Literal",
    "value": "2019",
    "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#gYear"
    }
}


literal created with function call:
{
    "termType": "Literal",
    "value": "2019"
}
Literal after explicitly setting datatype:
{
    "termType": "Literal",
    "value": "2019",
    "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#gYear"
    }
}
@prefix : <#>.
@prefix n: <name#>.
@prefix n0: <http://xmlns.com/foaf/0.1/>.
@prefix XML: <http://www.w3.org/2001/XMLSchema#>.

n:notme n0:birthday "2019"^^XML:gYear; n0:name "Meier".
```

### related:
- https://forum.solidproject.org/t/specify-data-types-with-rdflib-js/1156
- https://github.com/linkeddata/rdflib.js/issues/150
- https://github.com/solid/solid-tutorial-rdflib.js#rdf-term-types

### LINKED ISSUE #654: Allow library building from source when including in `package.json`
The current `package.json` does not have a `prepare` step, which means that it is hard for development projects to include a specific branch or commit to their dependencies (as no libraries will be built automatically).

I propose to include:

```
"prepare": "npm run build && npm run build:types && npm run build:browser && npm run build:esm",
```

This allows projects to include rdflib as follows in their `package.json` dependencies:

```
"rdflib": "github:linkeddata/rdflib.js#COMMITHASH",
```

As per <https://docs.npmjs.com/cli/v10/configuring-npm/package-json#git-urls-as-dependencies>

jeswr:
Consolidated quick-fix **wave 1/5 (repo & docs hygiene)** from the rdflib.js backlog triage (see #824); consolidated into one PR to keep PR volume low.

## Covered issues

| Issue | Fix |
| --- | --- |
| #654 | Add a `prepare` script so installs from a git URL build `lib/` (npm's documented mechanism for git dependencies); previously the package's `main` entry did not exist. |
| #87 | Remove `src/jsonparser.js`: unused by any code path (JSON-LD goes through `src/jsonldparser.js`), undocumented, untested; its only reference was the vestigial `$rdf.jsonParser` re-export in `src/index.ts`, which is removed with it. |
| #88 | `example/people/book.js` called the non-existent `kb.load()`; switched to `fetcher.load()` and moved the friends processing inside the fetch callback so it runs against a populated store. |
| #319 | README's RDFJS "task force spec" link had been a 404 for years; now points at the RDF/JS data model spec at rdf.js.org. |
| #328 | Documented the `literal()` factory signature in the README — `literal(value, languageOrDatatype)` — including the pitfall that the 3-argument `literal(value, undefined, datatype)` form from older tutorials silently ignores the datatype (verified against current build). |
| #301 | Added a "Querying the store with SPARQL" section to `Documentation/webapp-intro.html`: `SPARQLToQuery` + `store.query()/querySync()` usage and an explicit list of the supported SPARQL subset (SELECT-only, PREFIX, BGPs, OPTIONAL, minimal FILTER; no UNION/LIMIT/etc.). |
| #221 | Documented in the README when `<canonical> <http://www.w3.org/2007/ont/link#uri> <obsoleted>` statements are added: only when smushing features are enabled (off by default since #458), as alias book-keeping when the store equates two nodes. Behavior verified empirically against the current build. |

Closes #654
Closes #87
Closes #88
Closes #319
Closes #328
Closes #301
Closes #221

## Re-applied stale contributor PRs (close manually)

- Re-applies **#281** by @darrengarvey, with credit (`Co-authored-by` trailer) — updated to link directly to https://rdf.js.org/data-model-spec/ per the #319 thread, rather than the rdf.js.org homepage.
- Re-applies **#317** by @codam, with credit — turtle-intro typo fixes; the 2019 patch still applied cleanly.
- Re-applies **#318** by @codam, with credit — webapp-intro typo/code-sample fixes. Hunks already fixed on main in the meantime (the smart-quote code sample) were skipped, as was one wording change ("turtle text" → "a turtle text") that no longer reads as an improvement.

## Deviations / notes for review

- **#87 is technically a breaking change**: `$rdf.jsonParser` was an exported (though undocumented and unused) symbol. If you'd rather deprecate first, drop commit `chore: remove unused jsonparser (#87)` from this PR.
- **#654**: `prepare` runs `npm run build` only (transpiled JS). It does **not** run `build:types`, so a git-URL install still lacks `lib/index.d.ts`; extending it to `npm run build && npm run build:types` would fix that at the cost of slower installs (`prepare` also runs on local `npm ci`, including CI). Kept minimal per the issue; happy to extend.
- **#221**: the store adds the `link#uri` statement (in `IndexedFormula.replaceWith`), not the fetcher; and its object is nowadays the obsoleted **NamedNode**, not a literal as in the 2017 report. Documented current behavior.

## Expected conflicts with sibling waves

- `package.json` (one added line) will trivially conflict with wave QF-5's package.json edits.
- No overlap with in-flight #830/#831 (serializer / Turtle parser untouched).

## Local gates (all green before push)

- `npm run lint`: 0 errors (5 pre-existing `no-console` warnings in untouched files)
- `npm test`: unit 308 passing / 0 failing; serialize suite all pass; `test:types` (tsc) clean

_Generated by an agent (Claude Fable 5) on @jeswr's behalf — draft for his review._

> **Review timing:** This draft was prepared with Claude; I (@jeswr) will personally review it before it progresses. I'm currently batching a lot of work in flight, so expect active review Wednesday–Friday (8–10 July).


jeswr:
Documentation change only, typo fixing and examples. LGTM and I see nothing contriversial here.

bourgeoa:
Nothing controversial. Agreed.

Issue #593, opened 2022-12-15
Jsonld serialise collections to @list


bourgeoa:
Jsonld serialise collections as `rdf:first ...rdf:rest` @list should be a better solution if jsonld parse generates collections

TallTed:
@bourgeoa — `@list` should be wrapped in backticks, as ``` `@list` ```, in your initial comment (and the issue title), so that that GitHub user is not pinged for every update to this issue...

bourgeoa:
Thanks. This is not the first time.

PR #838, opened 2026-07-05
feat: rebuild the JSON-LD parse path on jsonld.toRDF(), removing the bespoke walker and its SSRF surface

### LINKED ISSUE #391: Add standard JSONLD parser tests
## JSON-LD tests for rdflib

https://w3c.github.io/json-ld-api/tests/

Points to manifest

 https://w3c.github.io/json-ld-api/tests/manifest.jsonld

jeswr:
> **Agent-generated draft PR** for @jeswr to review, from the backlog-triage programme (#823 / #824). The flagged behaviour deltas below (especially 5 and 6) need his sign-off; nothing is final.

## Summary

This PR deletes the bespoke JSON-LD parse walker (a hand-rolled traversal of `jsonld.flatten()` output) and rebuilds `src/jsonldparser.js` on **`jsonld.toRDF()`** — the W3C JSON-LD 1.1 *Deserialize JSON-LD to RDF* algorithm from the **already-installed** `jsonld@^9` dependency. No new runtime dependencies; no bundle-size change (the dynamic `await import('jsonld')` + `jsonld.default || jsonld` browser-ESM dance is preserved verbatim).

The module keeps its public shape: `export default async function jsonldParser(str, kb, base)` — with one **additive** optional `options` argument (`{ documentLoader, expandContext }`) that is also threaded through `parse()` (new optional 6th parameter) and the Fetcher's JSON-LD handler (`options.documentLoader` on load options).

### Design

- `jsonld.toRDF(JSON.parse(str), { base, documentLoader, expandContext? })` produces abstract RDF/JS-style quads; a small adapter maps them onto rdflib terms via `kb.rdfFactory` and adds them with `kb.addStatement()`.
- **Doc-graph placement matches rdflib conventions**: default-graph quads get `why` = `kb.rdfFactory.namedNode(base)` (exactly as before); quads inside `@graph` get that named graph as `why`.
- **Blank nodes are fresh per parse**: a per-parse label→`blankNode()` map replaces the old pass-through of the document's `_:` labels, so two documents parsed into one store can never conflate blank nodes again.
- **`@list` → `Collection` folding is preserved**: when `kb.rdfFactory.supports["COLLECTIONS"]` is true, well-formed `rdf:first`/`rdf:rest` chains emitted by `toRDF` are folded back into rdflib `Collection` terms (nested and empty lists included), mirroring the Turtle path's idiom in `src/lists.ts` — but scoped to *this parse's quads only* and non-throwing: non-well-formed chains stay as raw triples. Downstream reliance on `Collection.elements` keeps working.
- Downstream contract untouched: `parse()` callback semantics, fetcher metadata-graph statement shapes, and `Collection.elements` are all preserved.

## SSRF surface removal

Until now, parsing attacker-controlled JSON-LD through rdflib invoked jsonld.js's **platform-default document loader** to dereference remote `@context` URLs — arbitrary outbound HTTP(S) requests from the process doing the parsing, bypassing rdflib's Fetcher entirely: a server-side request forgery surface for any service that parses untrusted JSON-LD through rdflib (NSS-style servers do), with no injection hook and no offline mode.

This PR **removes that SSRF surface: `parse()` no longer issues network requests for remote `@context` URLs unless a `documentLoader` is explicitly injected.** The new default loader refuses with an actionable error that names the context URL and points at the opt-in:

```
Refused to fetch remote @context <https://…/context.jsonld>: rdflib.js disables remote
JSON-LD context fetching by default, … To opt in, pass an explicit `documentLoader` …
```

Opt-in paths (all additive): `jsonldParser(str, kb, base, { documentLoader })`, `parse(str, kb, base, contentType, callback, { documentLoader })`, and `fetcher.load(uri, { documentLoader })`. The refusal surfaces through `parse()`'s error callback and through `Fetcher`'s `failFetch` path, both covered by tests.

## Behavior deltas

| # | Delta | Before (flatten-walker) | After (`jsonld.toRDF`) | Why acceptable |
|---|-------|-------------------------|------------------------|----------------|
| 1 | Named-graph documents (`@graph`) | **Crash**: `NamedNode IRI "@graph" must be absolute` — document unparseable | Parses; `@graph` contents get the named graph as `why`, top-level quads stay in the doc graph | Fixes a verified defect (regression-tested) |
| 2 | `@json` literals | **Crash** (`namedNode("@json")`) | Literal typed `rdf:JSON` | Fixes a verified defect |
| 3 | Native JSON booleans | `true` → `"1"^^xsd:boolean` | `"true"^^xsd:boolean` (canonical) | Spec-correct; same direction as #831's move to spec-mandated lexical forms |
| 4 | Native JSON non-integer numbers | `1.5` → `"1.5"^^xsd:decimal` (wrong datatype) | `"1.5E0"^^xsd:double` (spec datatype + canonical lexical) | Spec-correct; same direction as #831. Native integers are unchanged (`42` → `"42"^^xsd:integer`) |
| 5 | String values explicitly typed `xsd:double` | Lexical form preserved (`"3.141e0"`) | Canonicalized (`"3.141E0"`) — mandated by the W3C *Object to RDF Conversion* algorithm | **FLAGGED**: lexical change for non-canonical doubles in JSON-LD sources; two round-trip expectations in `tests/unit/serialize-test.js` updated accordingly (the `tests/serialize` golden refs are untouched). Other datatypes (`xsd:decimal`, `xsd:integer`, `xsd:boolean` as strings) keep their lexical forms |
| 6 | Remote `@context` URLs | Silently dereferenced over the network (SSRF) | **Refused by default** with a clear error; opt-in via injected `documentLoader` | **FLAGGED (breaking-ish)**: documents relying on remote contexts now error by default instead of silently fetching. This is the security point of the PR |
| 7 | Blank node labels | Document's `_:b0`-style labels copied into the store → cross-document conflation (data corruption) | Fresh store-allocated labels per parse | Fixes a verified defect; only raw label *strings* change, structure is identical (tests updated to match by structure) |
| 8 | Statement insertion order | `flatten()` output order | `toRDF` node-map-sorted order | Insertion order was never a documented contract; order-dependent tests rewritten to pattern matching |
| 9 | Empty `@list` / lone `rdf:nil` objects | `{"@list": []}` → empty `Collection`; `{"@id": "rdf:nil"}` → NamedNode | Any lone `rdf:nil` object (not under `rdf:rest`, COLLECTIONS mode) → empty `Collection` | `toRDF` emits `rdf:nil` for both, so they can no longer be distinguished; matches the Turtle path's `substituteNillsInDoc` behavior |
| 10 | Relative IRIs that cannot be resolved | Throw from the `NamedNode` constructor | Dropped silently (spec behavior) | Edge case; spec-mandated |
| 11 | Module-internal named exports (`jsonldObjectToTerm`) | Exported from `src/jsonldparser.js` | Removed with the walker | Never re-exported from the package index; not part of the public API |
| 12 | Error shapes for invalid JSON-LD | Walker/`NamedNode` constructor errors | jsonld.js `JsonLdError`s (e.g. `Invalid JSON-LD syntax; "@id" value must a string.`) | Invalid *JSON* still surfaces the familiar `JSON.parse` `SyntaxError` (existing tests unchanged); JSON-LD-level errors are now precise and spec-sourced |

## Ecosystem backward-compatibility (SolidOS / NSS / solid-panes / mashlib)

- **Only `application/ld+json` parsing changes.** Turtle/N3/RDFa/XML paths and the serializer are untouched; every `tests/serialize` golden ref passes unchanged.
- **The one delta consumers are likely to hit in the wild is 6**: a JSON-LD document whose `@context` is a remote URL now fails to parse — with an error naming the URL and the opt-in — instead of silently fetching it. Documents with inline contexts are unaffected. The opt-in (`documentLoader`) is available at every level: `jsonldParser`, `parse()`, and `fetcher.load()`.
- Deltas 5/7/8 change literal lexical forms (non-canonical `xsd:double` from JSON-LD sources only), blank-node label *strings*, and insertion order. None of these were documented contracts, but code matching on exact labels or statement order would notice.
- The API surface is purely additive (the optional `options` parameter); the only removed export is the module-internal `jsonldObjectToTerm` (delta 11), which was never re-exported from the package index.
- Semver: at least `release-minor` for the additive API; delta 6's default-behaviour change is the argument for treating it as major — maintainer's call on the label.

## W3C conformance

rdflib runs **zero** JSON-LD conformance tests today (#391). With this change, toRdf behavior is **inherited from jsonld.js's upstream W3C JSON-LD 1.1 test-suite CI** (digitalbazaar runs the official toRdf manifests) instead of a hand-rolled walker that was verifiably non-conformant on named graphs, `@json`, native datatypes, and blank-node handling. Locally, this PR adds the four integration cases @ericprud proposed in #391 (embedded `@context`, API-supplied context via `expandContext`, API-supplied base, remote context with relative IRI resolution) rather than mechanically importing the full suite.

## Tests

New file `tests/unit/jsonld-parser-test.js` (33 new tests overall) covering: doc-graph placement (string and NamedNode base), named graphs (regression for the crash, incl. through `parse()`'s callback), `@json` → `rdf:JSON`, native datatypes (boolean/integer/double/string/lang-tagged), relative-IRI resolution against base and against a base **with a fragment**, per-parse blank-node freshness (cross-document non-conflation + within-parse identity), `@list` → `Collection` (elements + datatypes), nested lists → nested Collections, empty list → empty Collection, raw first/rest chains under `CanonicalDataFactory`, non-well-formed chains staying raw (non-throwing), remote-context refusal (error message, `contextUrl`, via promise and via `parse()` error callback, empty store), injected `documentLoader` (incl. ericprud case 4 and threading through `parse()`), ericprud cases 1–3, invalid-JSON and invalid-JSON-LD error shapes, and callback/promise API compatibility.

`tests/unit/fetcher-jsonld-test.js`: kept green as-is, extended with two fetcher-level tests — remote-context document fails the load with the refusal message (no outbound request), injected `documentLoader` via `fetcher.load(uri, { documentLoader })` succeeds.

Updated with tabled justification: `tests/unit/parse-test.js` (deltas 7/8: index-based assertions → pattern-based; blank-node labels matched by structure) and `tests/unit/serialize-test.js` (delta 5: the two *source jsonld* round-trip blocks now expect canonical `xsd:double` lexicals; the *source ttl* golden blocks are untouched).

**Full suite**: `test:unit` 341 passing (308 on `main` + 33 new) · `test:serialize` green with zero golden-ref diffs, incl. t16's JSON-LD output (serializer untouched) · `test:types` green · `lint` 0 errors (5 pre-existing warnings in untouched files).

## LOC

- `src/`: **+327 / −107** across 4 files. The 144-line walker is fully deleted; its replacement is ~130 lines of `toRDF` adapter + ~120 lines of documented Collection-folding (the rest of the src delta is the additive `options` typing in `types.ts`/`parse.ts`/`fetcher.ts`).
- Overall: +932 / −170 (the bulk is the new 500-line test file).

## Issue closures

Closes #391.

Completes the JSON-LD half of #349 and the JSON-LD instance of #405 (both umbrella issues remain open for their other formats).

## Follow-up: JSON-LD serializer (#593)

Deliberately **not** bundled here. The natural counterpart of this PR is to replace the Turtle→`@frogcat/ttl2jsonld` serializer detour with `jsonld.fromRDF()` + `jsonld.compact()`, which would emit `@list` for Collections (the #593 ask) and drop the ttl2jsonld dependency — but it forces golden-output design decisions (the `tests/serialize` t16 byte-golden and the compaction-context/prefix-selection policy) that deserve their own focused review rather than riding a parser PR. Plan: `fromRDF` over the store's doc-graph quads (unfolding Collections to first/rest for the algorithm), then `compact` against a context assembled from the store's prefix map; golden updates reviewed as their own delta set.

## Merge-order note

No conflict with #831 (Turtle/N3 only). #836 touches neighbouring lines in `parse.ts`'s import block and moves `utils.ts` → `utils/index.ts`; this PR's rewritten `jsonldparser.js` no longer imports `./utils` at all, so the two merge cleanly in either order.

### DIFF
--- a/src/fetcher.ts
+++ b/src/fetcher.ts
@@ -41,7 +41,7 @@ import crossFetch, { Headers } from 'cross-fetch'
 import type { Document as XmldomDocument, Element as XmldomElement, Node as XmldomNode } from '@xmldom/xmldom'
 
 import {
-  ContentType, TurtleContentType, RDFXMLContentType, XHTMLContentType
+  ContentType, TurtleContentType, RDFXMLContentType, XHTMLContentType, JsonLdDocumentLoader
 } from './types'
 import { termValue } from './utils/termValue'
 import {
@@ -178,6 +178,12 @@ export interface AutoInitOptions extends RequestInit{
   noRDFa?: boolean
   handlers?: Handler[]
   timeout?: number
+  /**
+   * JSON-LD only: loader used to resolve remote `@context` URLs. When
+   * omitted, remote context fetching is refused (SSRF protection) and
+   * documents that rely on it fail to parse.
+   */
+  documentLoader?: JsonLdDocumentLoader
   method?: HTTPMethods
   retriedWithNoCredentials?: boolean
   requestedURI?: string
@@ -511,7 +517,9 @@ class JsonLdHandler extends Handler {
   ): Promise<ExtendedResponse | FetchError> {
     const kb = fetcher.store
     try {
-      await jsonldParser(responseText, kb, options.original.value)
+      // Remote @context fetching is refused unless the caller explicitly
+      // injected options.documentLoader (deliberate SSRF protection)
+      await jsonldParser(responseText, kb, options.original.value, { documentLoader: options.documentLoader })
       fetcher.store.add(options.original, ns.rdf('type'), ns.link('RDFDocument'), fetcher.appNode)
       return fetcher.doneFetch(options, response)
     } catch (err) {
--- a/src/jsonldparser.js
+++ b/src/jsonldparser.js
@@ -101,8 +101,8 @@ function isNil (term) {
  * parse path does (see src/lists.ts) so `Collection.elements` keeps working
  * for downstream consumers.
  *
- * Operates only on the quads produced by the current parse — never on
- * pre-existing store content — and, unlike lists.ts's convertFirstRestNil,
+ * Operates only on the quads produced by the current parse (never on
+ * pre-existing store content) and, unlike lists.ts's convertFirstRestNil,
  * leaves non-well-formed chains as raw triples instead of throwing.
  *
  * @param kb - The store whose rdfFactory builds the Collections.
--- a/src/jsonldparser.js
+++ b/src/jsonldparser.js
@@ -1,145 +1,317 @@
-import {arrayToStatements} from './utils'
+const RDF_NS = 'http://www.w3.org/1999/02/22-rdf-syntax-ns#'
+const RDF_FIRST = RDF_NS + 'first'
+const RDF_REST = RDF_NS + 'rest'
+const RDF_NIL = RDF_NS + 'nil'
+const XSD_STRING = 'http://www.w3.org/2001/XMLSchema#string'
 
 /**
- * Parses json-ld formatted JS objects to a rdf Term.
- * @param kb - The DataFactory to use.
- * @param obj - The json-ld object to process.
- * @return {Literal|NamedNode|BlankNode|Collection}
+ * Marker property used to recognize our own remote-context refusal error when
+ * jsonld.js wraps it inside a JsonLdError.
  */
-export function jsonldObjectToTerm(kb, obj) {
-  if (typeof obj === 'string') {
-    return kb.rdfFactory.literal(obj)
-  }
+const REFUSED_REMOTE_CONTEXT = 'rdflibRemoteContextRefused'
 
-  if (Object.prototype.hasOwnProperty.call(obj, '@list')) {
-    if (kb.rdfFactory.supports["COLLECTIONS"] === true) {
-      return listToCollection(kb, obj['@list'])
+/**
+ * Builds the error thrown when a document requests a remote `@context` and no
+ * documentLoader has been injected by the caller.
+ */
+function remoteContextRefusedError (url) {
+  const err = new Error(
+    'Refused to fetch remote @context <' + url + '>: rdflib.js disables remote ' +
+    'JSON-LD context fetching by default, because dereferencing contexts from ' +
+    'untrusted documents lets attackers trigger arbitrary outbound HTTP(S) ' +
+    'requests (SSRF). To opt in, pass an explicit `documentLoader` in the ' +
+    'options argument of jsonldParser/parse (e.g. jsonld.js\'s ' +
+    'documentLoaders.node() or a loader backed by your own fetch policy).'
+  )
+  err[REFUSED_REMOTE_CONTEXT] = true
+  err.contextUrl = url
+  return err
+}
+
+/**
+ * The default documentLoader: refuses every remote context fetch.
+ */
+async function refusingDocumentLoader (url) {
+  throw remoteContextRefusedError(url)
+}
+
+/**
+ * jsonld.js wraps documentLoader failures in a JsonLdError whose generic
+ * message hides the cause. If our refusal error is anywhere in the cause
+ * chain, surface it directly so callers get the actionable message.
+ */
+function unwrapRefusalError (err) {
+  const seen = new Set()
+  let e = err
+  while (e && typeof e === 'object' && !seen.has(e)) {
+    seen.add(e)
+    if (e[REFUSED_REMOTE_CONTEXT]) {
+      return e
     }
-
-    return listToStatements(kb, obj)
+    e = e.cause || (e.details && e.details.cause)
   }
-
-  if (Object.prototype.hasOwnProperty.call(obj, '@id')) {
-    return nodeType(kb, obj)
-  }
-
-  if (Object.prototype.hasOwnProperty.call(obj, '@language')) {
-    return kb.rdfFactory.literal(obj['@value'], obj['@language'])
-  }
-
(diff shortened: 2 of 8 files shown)


bourgeoa:
@jeswr I think this opens the following question for rdflib.
Is it a library open to full multigraph jsonld or limited to one that can be rendered as turtle.
The why property is used in rdflib for example for metadata.

As an example Activity pub is using only one graph. It is more an enhanced JSON

To my understanding @timbl was not very fan of JSON -LD

I think this need to be more openly discussed.

jeswr:
I'm not exactly a fan of JSON-LD either. My hope for this PR was that it would reduce the LOC around JSON-LD to make things easier to maintain. It has not.

The main need I see for JSON-LD in the ecosystem at present is for [verifiable credentials](https://www.w3.org/TR/vc-data-model-2.0/); which rely on the graph for attaching proofs, and often do signing over the JSON serialisation. Whilst I draw issue with both of those design decisions - there is a demand for putting verifiable credentials in Pods, and this is the standard that many issuers use.

As far as this PR, I will look at the code more closely in the coming days to see if it is possible to get the LOC down without introducing breaking changes to RDFLib.

Thanks for flagging that RDFlib currently rejects named graphs in JSON-LD. I was not aware of this.

